PR #8, opened 2024-12-04
Update to go 1.23

jmhbnz:
Let's try and keep the operator aligned with our other etcd-io projects.

-  Use go version from go.mod in actions workflows.
- Update to go 1.23.
- Align devcontainer with `etcd-io/etcd`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,9 +1,11 @@
 {
   "name": "Kubebuilder DevContainer",
-  "image": "golang:1.22",
+  "image": "mcr.microsoft.com/devcontainers/go:1.23-bookworm",
   "features": {
     "ghcr.io/devcontainers/features/docker-in-docker:2": {},
-    "ghcr.io/devcontainers/features/git:1": {}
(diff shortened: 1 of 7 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,9 +1,11 @@
 {
   "name": "Kubebuilder DevContainer",
-  "image": "golang:1.22",
+  "image": "mcr.microsoft.com/devcontainers/go:1.23-bookworm",
```
kvaps, line 3: why?

jmhbnz: Thanks for the question, as mentioned in the commit message this aligns the devcontainer used across `etcd-io/etcd` and `etcd-io/etcd-operator`.  I don't have particularly strong opinions on what the devcontainer actually is so longs it is debian based and has sensible tooling included. Where it makes sense we should align it across the projects.

ahrtr (approved):
LGTM

Thanks @jmhbnz

ivanvc:
Link to etcd-io/etcd#19000.

PR #17, opened 2025-01-04
Initial implementation for the Reconcile Logic

gdasson:
This PR is an intial implementation of the design mentioned [here](https://docs.google.com/document/d/1r-woKinZtLw9Diw2DdznZl5ASD_oVelAvkFJaqk11lk/edit?tab=t.0#heading=h.hkrgg422ub1d)

The code is now in working state and able to create an etcd cluster. However, the code will continue to be refined and made more production ready as we review and incorporate feedback.


internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+
+	err := c.Get(ctx, types.NamespacedName{Name: configMapNameForEtcdCluster(ec), Namespace: ec.Namespace}, &corev1.ConfigMap{})
+	if err != nil && errors.IsNotFound(err) {
+		createErr := c.Create(ctx, cm)
+		return createErr
+	}
+
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("cannot find ConfigMap for EtcdCluster %s: %w", ec.Name, err)
+	}
```
ahrtr, line 327:
```suggestion
	if err != nil {
		if errors.IsNotFound(err) {
			createErr := c.Create(ctx, cm)
			return createErr	
		}
		return fmt.Errorf("cannot find ConfigMap for EtcdCluster %s: %w", ec.Name, err)
	}
```

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func CreateHeadlessServiceIfDoesntExist(ctx context.Context, logger logr.Logger, c client.Client, ec *ecv1alpha1.EtcdCluster, scheme *runtime.Scheme) error {
```
ahrtr, line 228:
```suggestion
func CreateHeadlessServiceIfNotExist(ctx context.Context, logger logr.Logger, c client.Client, ec *ecv1alpha1.EtcdCluster, scheme *runtime.Scheme) error {
```

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+	if *sts.Spec.Replicas > int32(etcdCluster.Spec.Size) {
+		return SSBiggerThanDesired, nil
+	}
+
+	if *sts.Spec.Replicas == int32(etcdCluster.Spec.Size) {
+		return SSEqualToDesired, nil
+	}
+
+	return "", nil
+}
```
ahrtr, line 363: It's a little over engineering to me. I think we just need to provide a function to get the statefulset.

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	if !ok {
+		return fmt.Errorf("StatefulSet %s/%s is not ready", ec.Namespace, ec.Name)
+	}
+
+	//err = SetControllerReference(ctx, ec, *c, scheme)
+	//if err != nil {
+	//	return err
+	//}
```
ahrtr, line 171: Please remove the commented source code

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+
+	err = CheckStatefulSetControlledByEtcdOperator(ctx, *c, ec)
+	if err != nil {
+		return err
+	}
+	//os.Exit(0)
```
ahrtr, line 177: Please remove the commented source code

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+		UID:                ec.GetUID(),
+		BlockOwnerDeletion: ptr.To(true),
+		Controller:         ptr.To(true),
+	}
+
+	var owners []metav1.OwnerReference
+	owners = append(owners, ref)
+	return owners, nil
+}
+func CreateOrPatchSS(ctx context.Context, logger logr.Logger, ec *ecv1alpha1.EtcdCluster, c *client.Client, replicas int32, scheme *runtime.Scheme) error {
```
ahrtr, line 48:
```suggestion
func CreateOrPatchStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha1.EtcdCluster, c *client.Client, replicas int32, scheme *runtime.Scheme) error {
```

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
+package utils
```
ahrtr, line 1: Suggest to move the `utils.go` into the existing package `internal/controller`. Usually a dedicate package should be agnostic to the business logic, but I see that it's almost tightly coupled with the etcd operator's reconciliation logic.

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+			Namespace: ec.Namespace,
+		},
+	}
+
+	labels := map[string]string{
+		"app":        ec.Name,
+		"controller": ec.Name,
+	}
+
+	if err := validateOwner(ec, sts); err != nil {
```
ahrtr, line 61: You already set the StatefulSet's namespace = ec.Namespace above, I don't see any necessarity to check their namespaces match again. Also the name `validateOwner` isn't accurate.

So suggest to remove the function `validateOwner` to keep it as simple as possible for now.

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil && errors.IsNotFound(err) {
+		createErr := c.Create(ctx, cm)
+		return createErr
+	}
+
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("cannot find ConfigMap for EtcdCluster %s: %w", ec.Name, err)
+	}
+
+	if err := validateOwner(ec, cm); err != nil {
```
ahrtr, line 329: ditto, suggest to remove it

internal/utils/utils.go
```diff
@@ -0,0 +1,363 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	SSSmallerThanDesired = "SSSmallerThanDesired"
+	SSBiggerThanDesired  = "SSBiggerThanDesired"
+	SSEqualToDesired     = "SSEqualToDesired"
```
ahrtr, line 27: Suggest to remove them for now. Also usually a better way to define an enum for such case.

ahrtr:
@gdasson Thanks for the PR. 

Could you please resolve the comments and ensure the PR is in the best shape you think before marking it as ready for review and also remove "draft" from the title?

internal/etcdutils/etcdutils.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer func() {
+		err := c.Close()
+		if err != nil {
+			return
+		}
+		cancel()
```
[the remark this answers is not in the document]
gdasson: I am not sure if I understand the review comment. Are you suggesting I should not cancel here?

internal/etcdutils/etcdutils.go
```diff
@@ -0,0 +1,230 @@
...
+	"errors"
+	"fmt"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	"go.etcd.io/etcd/client/pkg/v3/logutil"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+	"sort"
+	"strings"
+	"sync"
+	"time"
```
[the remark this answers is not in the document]
gdasson: I am sorry, I am confused about this. Are you suggesting to sort this in a specific order so that it passes the linter check?

internal/etcdutils/etcdutils.go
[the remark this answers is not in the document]
ahrtr: When `err != nil` is true, your current implement skips calling `cancel()`. It isn't correct, `cancel()` should always be called.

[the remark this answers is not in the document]
gdasson: Ah, ok. Got it. Will fix.

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,196 @@ type EtcdClusterReconciler struct {
...
+					if err != nil {
+						// The member is not promoted yet, so we error out
+						return ctrl.Result{}, err
+					}
+				} else {
+					// Learner is not yet ready. We can't add another learner until this one is promoted
+					// So let's requeue
+					logger.Info("The learner member isn't ready to be promoted yet", "learnerID", learner)
+					return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+				}
```
ahrtr, line 203: Please process the learner node promotion separately. Let's do not get it tightly coupled with cluster scale in & out, so as to make it clearer and more readable.

Refer to https://github.com/ahrtr/etcd-operator/blob/f3b16f5e0b8e2751a1ddc35e0c6a8ccaf4afe568/controller.go#L321-L350

ahrtr:
@jmhbnz @ivanvc Do you know why workflow checks are not triggered in this PR? cc @hakman 

Also it'd be great if we could add more e2e test case as mentioned in https://github.com/etcd-io/etcd-operator/issues/19. Of course, in followup PRs.

internal/controller/etcdcluster_controller.go
```diff
@@ -18,24 +18,36 @@ package controller
...
+	Recorder record.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=operator.etcd.io,resources=etcdclusters,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=operator.etcd.io,resources=etcdclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=operator.etcd.io,resources=etcdclusters/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch;get;list;update
```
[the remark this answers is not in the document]
gdasson: I added these markers manually and then ran `make manifests` to automatically generate the required role/clusterrole objects.

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+
+	err := r.Get(ctx, req.NamespacedName, etcdCluster)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("EtcdCluster resource not found. Ignoring since object may have been deleted")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
```
ahrtr, line 75: let's add a check something like below: 
```
if etcdCluster.Size == 0 {
   print a warning
   return
}
```

Afterwards, simplify the following logic since we can assume that etcdCluster.Size is guaranteed to be > 0.

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+	// If the Statefulsets is not controlled by this EtcdCluster resource, we should log
+	// a warning to the event recorder and return error msg.
+	err = checkStatefulSetControlledByEtcdOperator(etcdCluster, sts)
+	if err != nil {
+		logger.Error(err, "StatefulSet is not controlled by this EtcdCluster resource")
+		return ctrl.Result{}, err
+	}
+
+	// If statefulset size is 0. try to instantiate the cluster with 1 member
+	if sts.Spec.Replicas != nil && *sts.Spec.Replicas == 0 && etcdCluster.Spec.Size > 0 {
```
ahrtr, line 115:
```suggestion
	if sts.Spec.Replicas != nil && *sts.Spec.Replicas == 0 {
```

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+		if _, err := etcdutils.AddMember(eps, []string{peerURL}, true); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		logger.Info("Learner member added successfully", "peerURLs", peerURL)
+
+		sts, err = createOrPatchStatefulSet(ctx, logger, etcdCluster, r.Client, int32(targetReplica), r.Scheme)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
ahrtr, line 202: Minor enhancement: We only need to call `createOrPatchStatefulSet` once, currently you call it twice in two if branches.

The rough psedo code is something like below,

```
if replica == expectedSize {
    return
}

if replica < expectedSize {
    call AddMember
} else{
    call memberRemove
}

call createOrPatchStatefulSet with the targetReplica
```

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+			return ctrl.Result{}, fmt.Errorf("couldn't find leader, memberCnt: %d", memberCnt)
+		}
+
+		learner, learnerStatus = etcdutils.FindLearnerStatus(healthInfos, logger)
+
+		if learner > 0 {
+			// There is at least one learner. Let's try to promote it or wait
+			// Find the learner status
+			logger.Info("Learner found", "learnedID", learner)
+			logger.Info("Learner status", "learnerStatus", learnerStatus)
```
ahrtr, line 163: Combine them into one log entry

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+
+	/* if memberCnt != len(healthInfos) {
+		// Only proceed when all members are healthy
+		return ctrl.Result{}, fmt.Errorf("memberCnt (%d) isn't equal to healthy member count (%d)", memberCnt, len(healthInfos))
+	} */
 
-	// TODO(user): your logic here
+	var learnerStatus *clientv3.StatusResponse
+	var learner uint64
+	var leaderStatus *clientv3.StatusResponse
```
ahrtr, line 147:
```suggestion
	var (
		learner uint64
		learnerStatus *clientv3.StatusResponse
		leaderStatus *clientv3.StatusResponse
	)
```

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+
+	memberCnt := 0
+	if memberListResp != nil {
+		memberCnt = len(memberListResp.Members)
+	}
+
+	/* if memberCnt != len(healthInfos) {
+		// Only proceed when all members are healthy
+		return ctrl.Result{}, fmt.Errorf("memberCnt (%d) isn't equal to healthy member count (%d)", memberCnt, len(healthInfos))
+	} */
```
ahrtr, line 143:
```suggestion
	 if memberCnt != len(healthInfos) {
		// Only proceed when all members are healthy
		return ctrl.Result{}, fmt.Errorf("memberCnt (%d) isn't equal to healthy member count (%d)", memberCnt, len(healthInfos))
	} 
```

internal/controller/utils.go
```diff
@@ -0,0 +1,391 @@
...
+		BlockOwnerDeletion: ptr.To(true),
+		Controller:         ptr.To(true),
+	}
+
+	var owners []metav1.OwnerReference
+	owners = append(owners, ref)
+	return owners, nil
+}
+
+func createOrPatchStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha1.EtcdCluster, c client.Client, replicas int32, scheme *runtime.Scheme) (*appsv1.StatefulSet, error) {
```
ahrtr, line 47: Let's follow SRP (Single Responsibility principle): Each function should only do one thing.

Let's add a function or method `reconcileStatefulSet` something like below,

```
func reconcileStatefulSet (...) {
     // prepare configmap (statefulset state)
     applyEtcdClusterState(...)

    createOrPatchStatefulSet(...)

     waitForStatefulSetReady(...)

     return getStatefulSet(...)
}
```

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +59,193 @@ type EtcdClusterReconciler struct {
...
+				// So let's requeue
+				logger.Info("The learner member isn't ready to be promoted yet", "learnerID", learner)
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+		}
+	}
+
+	//  Check if the size of the stateful set is less than expected size
+	//  Or if there is a pending learner to be promoted
+	if *sts.Spec.Replicas < int32(etcdCluster.Spec.Size) && memberCnt > 0 {
```
ahrtr, line 185:
```suggestion
	if *sts.Spec.Replicas < int32(etcdCluster.Spec.Size)  {
```

We should guarantee:
- `sts.Spec.Replica == memberCnt (returned from etcd cluster)` is always `true`. refer to https://github.com/ahrtr/etcd-operator/blob/f3b16f5e0b8e2751a1ddc35e0c6a8ccaf4afe568/controller.go#L302-L315
- `sts.Spec.Replica >= 1 ` is always true before performing the scale in & out

ahrtr:
@hakman @justinsb @jmhbnz @ArkaSaha30 Please take a look at this PR. It will be the base for the following PRs. Thanks.

cc @ivanvc 

Also I am most concerned about the e2e test as mentioned in https://github.com/etcd-io/etcd-operator/pull/17#issuecomment-2572899831

jmhbnz:
> @jmhbnz @ivanvc Do you know why workflow checks are not triggered in this PR? cc @hakman

Probably the GitHub actions bug we see from time to time, we may need to close and re-open this pr for it trigger. I will migrate the existing workflows to prow longer term.

/ok-to-test
/retest

ahrtr:
@gdasson please rebase this PR, thx

internal/controller/etcdcluster_controller.go
```diff
@@ -47,17 +63,189 @@ type EtcdClusterReconciler struct {
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("EtcdCluster resource not found. Ignoring since object may have been deleted")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if etcdCluster.Spec.Size == 0 {
+		logger.Info("EtcdCluster size is 0..Skipping next steps")
```
ahrtr, line 81: nit, a warning message might a little better. It is unusual to create a EtcdCluster with 0 member.

internal/controller/etcdcluster_controller.go
gdasson: @ahrtr : I am a little confused on this one. In the reference code you have given the inverse condition i.e replica != memberCnt but in the comment you are suggesting for it to be true? I am also thinking the pros/cons for the validation of sts.Spec.Replica == memberCnt before statefulset scale in/out. What additional assurance would this check provide that the reconcile logic is not currently guaranteeing?

ahrtr: Normally `sts.Spec.Replica == memberCnt (returned from etcd cluster)` should be always `true` after etcd-operator finishes each round of [Reconcile](https://github.com/etcd-io/etcd-operator/blob/b4cd582fd4657de1a3e102fc2dd1b42cd5c14a71/internal/controller/etcdcluster_controller.go#L49).

If it isn't true (not equal), then it means there are something wrong happened in previous Reconcile round, possible reasons:

1. etcd-operator crashes right **after** we add a new member (call MemberAdd)  and **before** reconcile the statefulSet (update the replica). In this case, `sts.Spec.Replica < memberCnt` is true.
2. etcd-operator crashes right **after** we delete a member (call MemberDelete)  and **before** reconcile the statefulSet (update the replica). In this case, `sts.Spec.Replica > memberCnt` is true.


We should always fix any problems coming from previous reconcile rounds before we do something new in current reconcile round. So once we get to current round's scale in & out steps, we can assume that `sts.Spec.Replica == memberCnt (returned from etcd cluster)` is true.

You can add a TODO item for now and do it in a followup PR, just similar to what I did in https://github.com/ahrtr/etcd-operator/blob/f3b16f5e0b8e2751a1ddc35e0c6a8ccaf4afe568/controller.go#L302-L315.

ahrtr:
@gdasson please rebase this PR and let's merge this PR. Also squash the commits (I can also do it for you if you have any difficulties). I will create more following tasks.

gdasson:
@ahrtr : I'll work on it today. I had a busy couple of days, sorry about not responding earlier.

internal/controller/etcdcluster_controller.go
gdasson: I am using https://github.com/go-logr/logr as the logging library for now and although it does offer verbosity levels, but as a simple library, I think it doesn't have a warning concept per se. 

We can move to a more mature logging library such as zap logger in a follow-on PR and then address this feedback comment.

gdasson:
@ahrtr : Done. Please review. Thanks.

internal/controller/etcdcluster_controller.go
gdasson: Makes sense. For now, I have added a placeholder for it, let's revisit as an additional task to cover this.

ahrtr (approved):
LGTM

Thanks for the nice work!

Will plan/raise more tasks next week.

PR #36, opened 2025-01-14
Add tool/mod into dependabot.yml and group bumping go.etcd.io/* dependencies

ahrtr:
cc @hakman @jmhbnz @soma00333


soma00333:
This change is really great
Thank you so much for your help

PR #39, opened 2025-01-15
Align Dockerfile builder image

jmhbnz:
Ensure golang patch version is specified.
Base of Debian Bookworm in line with other etcd projects.

Addresses: https://github.com/etcd-io/etcd-operator/pull/37#discussion_r1914896599

cc @ahrtr

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.23@sha256:7ea4c9dcb2b97ff8ee80a67db3d44f98c8ffa0d191399197007d8459c1453041 AS builder
+FROM golang:1.23.4-bookworm@sha256:6f085f2a025fcd189fefd7dc51c98ad34fa40c749f73b5522aa53b27278e4ec1 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 


jmhbnz:
Test failure addressed by https://github.com/kubernetes/test-infra/pull/34146

ivanvc:
Follow up on test failure at PR: kubernetes/test-infra#34147

hakman:
LGTM

hakman:
@jmhbnz I am curious, why do we set toolchain version in the `go.mod` file?

ahrtr:
> @jmhbnz I am curious, why do we set toolchain version in the `go.mod` file?

It should be a best practice. Pinning to a specific/consistent go version ensures we have the same go version in test and production environment, also ensure we always build the binary/image with a safe (usually the latest patch version) go version.

hakman:
> > @jmhbnz I am curious, why do we set toolchain version in the `go.mod` file?
> 
> It should be a best practice. Pinning to a specific/consistent go version ensures we have the same go version in test and production environment, also ensure we always build the binary/image with a safe (usually the latest patch version) go version.

I was a little confused on what Go version GHA uses from the `go.mod` file, either `toolchain go1.23.4` or `go 1.23.0`.
Maybe we need to set the Go version to the toolchain version.
```
2025-01-15T09:52:28.3490983Z Adding to the cache ...
2025-01-15T09:52:32.5799032Z Successfully cached go to /opt/hostedtoolcache/go/1.23.0/x64
2025-01-15T09:52:32.5800450Z Added go to the path
2025-01-15T09:52:32.5803586Z Successfully set up Go version 1.23.0
2025-01-15T09:52:36.6251957Z go: downloading go1.23.4 (linux/amd64)
2025-01-15T09:52:36.6408632Z [command]/opt/hostedtoolcache/go/1.23.0/x64/bin/go env GOMODCACHE
2025-01-15T09:52:36.6447123Z [command]/opt/hostedtoolcache/go/1.23.0/x64/bin/go env GOCACHE
```

ahrtr:
> I was a little confused on what Go version GHA uses from the `go.mod` file, either `toolchain go1.23.4` or `go 1.23.0`.
> Maybe we need to set the Go version to the toolchain version.

Good catch, we should follow the toolchain, just raised https://github.com/etcd-io/etcd-operator/pull/41

ahrtr:
Also I am thinking probably setting a tag 1.23.4-bookworm should be enough. There is no need to add a digest (SHA 256 value). I assume that the tag and digest is 1:1 mapping.

hakman:
> Also I am thinking probably setting a tag 1.23.4-bookworm should be enough. There is no need to add a digest (SHA 256 value). I assume that the tag and digest is 1:1 mapping.

It is not enough to use the tag, they regularly rebuild with newer bookworm base image.

ahrtr (approved):
LGTM

Thanks @jmhbnz

PR #57, opened 2025-01-27
Add reconciliation workflow diagram

ahrtr:
Note that .uxf file is the original file, and you can open it using [umlet](https://www.umlet.com/)

cc @gdasson @hakman @jmhbnz @justinsb @jberkus


jmhbnz (approved):
/lgtm
/approve

Thanks @ahrtr

PR #59, opened 2025-01-28
Add the initial scaffold of certificate management

ahrtr:
See the [Spec & Design doc](https://docs.google.com/document/d/1k8os77Hqxe36nUaZ31YqBGBuXcgdozVPppJYOfwUDFA/edit?tab=t.0)

cc @hakman @gdasson @ArkaSaha30 

also cc @jberkus @jmhbnz @justinsb

### DIFF
--- a/pkg/certificate/interface.go
+++ b/pkg/certificate/interface.go
@@ -1,0 +1,112 @@
+package certificate
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"time"
+)
+
+type ProviderType string
+
+const (
+	Auto        ProviderType = "auto"
+	CertManager ProviderType = "cert-manager"
+	// add more ...
+)
+
+func NewProvider(pt ProviderType) (Provider, error) {
+	switch pt {
+	case Auto:
+		return nil, nil // change me later
+	case CertManager:
+		return nil, nil // change me later
+	}
+
+	return nil, fmt.Errorf("unknown provider type: %s", pt)
+}
+
+// AltNames contains the domain names and IP addresses that will be added
+// to the x509 certificate SubAltNames fields. The values will be passed
+// directly to the x509.Certificate object.
+type AltNames struct {
+	DNSNames []string
+	IPs      []net.IP
+}
+
+// Config contains the basic fields required for creating a certificate
+type Config struct {
+	CommonName       string
+	Organization     []string
+	AltNames         AltNames
+	ValidityDuration time.Duration
+	CABundle         []byte
+
+	// ExtraConfig contains provider specific configurations.
+	ExtraConfig map[string]any
+}
+
+type Provider interface {
+	// EnsureCertificateSecret ensures the specified certificate is
+	// available as a Secret in Kubernetes. If the Secret does not
+	// exist, it will be created.
+	//
+	// Parameters:
+	// - ctx: Context for cancellation and deadlines.
+	// - secretName: Name of the Secret to ensure.
+	// - namespace: Namespace where the Secret should reside.
+	// - cfg: Configuration for the certificate.
+	//
+	// Returns:
+	// - nil if the operation succeeds, or an error otherwise.
+	EnsureCertificateSecret(ctx context.Context, secretName string, namespace string, cfg *Config) error
+
+	// ValidateCertificateSecret validates the certificate stored
+	// in the specified Secret. This checks if the certificate is
+	// valid (e.g., not expired, matches configuration).
+	//
+	// Parameters:
+	// - ctx: Context for cancellation and deadlines.
+	// - secretName: Name of the Secret to validate.
+	// - namespace: Namespace where the Secret resides.
+	// - cfg: Configuration to validate against.
+	//
+	// Returns:
+	// - true if the Secret is valid, false otherwise, along with
+	//   an error if validation fails.
+	ValidateCertificateSecret(ctx context.Context, secretName string, namespace string, cfg *Config) (bool, error)
+
+	// DeleteCertificateSecret explicitly deletes the Secret containing
+	// the certificate. This should only be used if the certificate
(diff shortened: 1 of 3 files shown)


ahrtr:
also cc @neolit123

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+	// is no longer needed.
+	//
+	// Parameters:
+	// - ctx: Context for cancellation and deadlines.
+	// - secretName: Name of the Secret to delete.
+	// - namespace: Namespace where the Secret resides.
+	//
+	// Returns:
+	// - nil if the operation succeeds, or an error otherwise.
+	DeleteCertificateSecret(ctx context.Context, secretName string, namespace string) error
```
neolit123, line 100: potentially move it below ValidateCertificateSecret

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+func NewProvider(pt ProviderType) Provider {
+	switch pt {
+	case Auto:
+		return nil // change me later
+	case CertManager:
+		return nil // change me later
+	}
+	panic(fmt.Sprintf("unknown provider type: %s", pt))
```
neolit123, line 24: better to return error instead of panic.

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type Config struct {
+	CommonName       string
+	Organization     []string
+	DNSNames         []string
+	ValidityDuration time.Duration
+	CA               []byte
```
neolit123, line 32: `CABundle` maybe, as commented on the doc.
and it can still be a bundle of 1 CA.

then again, most of the flags of k8s components use just `ca` that can actually be a bundle.

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+	//
+	// Parameters:
+	// - ctx: Context for cancellation and deadlines.
+	// - secretName: Name of the Secret to ensure.
+	// - namespace: Namespace where the Secret should reside.
+	// - cfg: Configuration for the certificate.
+	//
+	// Returns:
+	// - nil if the operation succeeds, or an error otherwise.
+	EnsureCertificateSecret(ctx context.Context, secretName string, namespace string, cfg Config) error
```
neolit123, line 51: cfg can be a pointer in all interface functions.

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// GetCertificateConfig returns the certificate configuration from the provider.
+	//
+	// Parameters:
+	// - secretName: Name of the Secret containing the certificate.
+	// - namespace: Namespace where the Secret resides.
+	//
+	// Returns:
+	// - Config if the Secret exists and is valid, or an error otherwise.
+	GetCertificateConfig(ctx context.Context, secretName string, namespace string) (Config, error)
```
neolit123, line 87: can be *Config

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+	case CertManager:
+		return nil // change me later
+	}
+	panic(fmt.Sprintf("unknown provider type: %s", pt))
+}
+
+type Config struct {
+	CommonName       string
+	Organization     []string
+	DNSNames         []string
```
neolit123, line 30: should be extraSANs or similar as a SAN can also be an IP

pkg/certificate/interface.go
```diff
@@ -0,0 +1,101 @@
...
+	ValidityDuration time.Duration
+	CA               []byte
+
+	// ExtraConfig contains provider specific configurations.
+	ExtraConfig map[string]string
```
neolit123, line 35: could be `map[string]interface{}` for more flexibility and leaving it to the provider to cast.

pkg/certificate/interface.go
ahrtr: updated & thx

pkg/certificate/interface.go
ahrtr: updated & thx

pkg/certificate/interface.go
ahrtr: Renamed to CABundle & thx

pkg/certificate/interface.go
ahrtr: updated & thx

pkg/certificate/interface.go
ahrtr: updated & thx

pkg/certificate/interface.go
ahrtr: updated & thx

pkg/certificate/interface.go
ahrtr: Currently we only use DNSName, as POD IP is uncertain and changeable.

We can extend it in the future when needed.

ahrtr:
Resolved all comments from @neolit123 (Thanks).

@hakman @neolit123 @gdasson @ArkaSaha30 @jmhbnz @justinsb @jberkus  PTAL, if there is no further comment, let me merge this PR.

gdasson:
@ahrtr : LGTM!

pkg/certificate/interface.go
ahrtr: Updated anyway now to keep it as more generic as possible. Added `AltNames` similar to what k8s.io/client-go does

https://github.com/kubernetes/client-go/blob/f2030849e1d6d961fa150b9be9560da33e5d00df/util/cert/cert.go#L51-L57

ahrtr:
cc @jmhbnz @hakman

jmhbnz (approved):
LGTM - Thanks @ahrtr

PR #40, opened 2025-01-15
Chore: Manage tool versions at tools/mod

soma00333:
## Overview
`golangci-lint` is managed at `tools/mod` by https://github.com/etcd-io/etcd-operator/pull/24
This pull request includes updates to manage other tools at `tools/mod`

## Check
I checked make command locally.
```
$ make lint
Downloading github.com/golangci/golangci-lint/cmd/golangci-lint@v1.63.4
/workspaces/etcd-operator/bin/golangci-lint run
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar. 
```
```
$ make kustomize
Downloading sigs.k8s.io/kustomize/kustomize/v5@v5.5.0
go: downloading github.com/spf13/cobra v1.8.0
go: downloading golang.org/x/text v0.16.0
go: downloading google.golang.org/protobuf v1.33.0
```
```
 $ make controller-gen
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.4
go: downloading golang.org/x/tools v0.26.0
go: downloading golang.org/x/sys v0.26.0
go: downloading golang.org/x/net v0.30.0
go: downloading golang.org/x/text v0.19.0
```
```
$ make envtest
Downloading sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20250106171007-7436275c4311
go: downloading sigs.k8s.io/controller-runtime v0.0.0-20250106171007-7436275c4311
go: downloading github.com/spf13/afero v1.6.0
go: downloading go.uber.org/multierr v1.10.0
```
```
$ make crd-ref-docs
Downloading github.com/elastic/crd-ref-docs@v0.1.0
go: downloading golang.org/x/tools v0.19.0
go: downloading sigs.k8s.io/controller-tools v0.14.0
go: downloading k8s.io/apimachinery v0.29.3
go: downloading github.com/google/uuid v1.3.0
go: downloading golang.org/x/crypto v0.21.0
go: downloading github.com/fatih/color v1.16.0
go: downloading golang.org/x/sys v0.18.0
go: downloading golang.org/x/mod v0.16.0
go: downloading k8s.io/klog/v2 v2.110.1
go: downloading k8s.io/apiextensions-apiserver v0.29.0
go: downloading github.com/gobuffalo/flect v1.0.2
go: downloading k8s.io/utils v0.0.0-20230726121419-3b25d923346b
[...]

PR #62, opened 2025-02-04
Convert existing e2e tests from Gingko to e2e framework

abdurrehman107:
The above adds the default e2e tests written using the e2e framework. These were earlier written in Gingko. The PR creates a Kind Cluster and installs Prometheus, Cert Manager and Controller manager.

The e2e_test.go file is left empty so I can include tests in it. Most of the work done in Gingko earlier has been replicated in the e2e_suite_test.go.

There are a couple things such as displaying logs at the end of each test (AfterAll() in Gingko) which I cannot add in e2e framework until we begin writing the tests. I have a few more tests ready, and we can begin on them once this PR is merged.


test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,175 @@ limitations under the License.
...
+				return ctx, p.Err()
+			}
+			if p := utils.RunCommand(
+				fmt.Sprintf("go install sigs.k8s.io/controller-tools/cmd/controller-gen@%s", ctrlGenVer),
+			); p.Err() != nil {
+				log.Printf("Failed to install controller-gen binary: %s: %s", p.Err(), p.Out())
+				return ctx, p.Err()
+			}
+			return ctx, nil
+		},
```
ahrtr, line 109: There are several targets in Makefile, i.e. `kustomize`, `controller-gen`, which are used to install the tools. Why do you install them again?

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,175 @@ limitations under the License.
...
+			}
+
+			// gen manifest files
+			log.Println("Generate manifests...")
+			if p := utils.RunCommand(
+				`controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases`,
+			); p.Err() != nil {
+				log.Printf("Failed to generate manifests: %s: %s", p.Err(), p.Out())
+				return ctx, p.Err()
+			}
```
ahrtr, line 129: The Makefile already has several targets: `install`, `manifests`, `kustomize`. Are you going cleanup all such targets in Makefile later?  It isn't clear what's your intention.

ahrtr:
The high level structure looks ok, but there is some overlaps between the e2e test suite and the Makefile as pointed out the above comments. The most straightforward way is to reuse the existing Makefile targets. Any refactoring or improvement should be in separate PRs. 

@hakman @gdasson @ArkaSaha30 PTAL

The [etcdcluster_controller_test.go](https://github.com/etcd-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller_test.go) needs to be updated as well.  It can be addressed in a separate PR.

test/e2e/e2e_suite_test.go
abdurrehman107: I did not look into the Makefile @ahrtr. I'll address this using the makefile and push an updated version

abdurrehman107:
> The [etcdcluster_controller_test.go](https://github.com/etcd-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller_test.go) needs to be updated as well. It can be addressed in a separate PR

Right. I'll take this up in the PR right after the `e2e-tests.go` PR.

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,185 @@ limitations under the License.
...
-	_, _ = fmt.Fprintf(GinkgoWriter, "Starting etcd-operator integration test suite\n")
-	RunSpecs(t, "e2e suite")
+func TestMain(m *testing.M) {
+	testEnv = env.New()
+	kindClusterName := "etcd-cluster"
+	kindCluster := kind.NewCluster(kindClusterName)
+
+	log.Println("Creating KinD cluster...")
+	testEnv.Setup(
+		envfuncs.CreateCluster(kindCluster, kindClusterName),
```
ahrtr, line 58: The workflow check already created a Kind cluster,  we need to either remove it from workflow check or from here. cc. @ivanvc 

https://github.com/kubernetes/test-infra/blob/f414bfaa8700111b528079226ac3e26fbbe61011/config/jobs/etcd/etcd-operator-presubmits.yaml#L53

ahrtr:
Suggest to follow the same order to setup & teardown the env. All e2e tests will be executed in between.

## Setup:
- Create a Kind cluster
  - either in workflow check or in e2e test case as mentioned in https://github.com/etcd-io/etcd-operator/pull/62#discussion_r1944699074
- prepare the resources
  - generate files
  - generate manifests
  - build operator images
- install dependencies (optional for now)
  - install prometheus operator
  - install certificate maanger
- set up environment
  - create namespace
  - install CRD
  - deploy etcd operator

## Teardown
- cleanup environment
  - undeploy etcd-operator
  - uninstall CRD
  - remove namespace
- remove dependencies (optional)
  - uninstall prometheus operator
  - uninstall certificate manager


Also please always keep in mind that one PR is supposed to do only one thing. Please move unrelated change (i.e. etcd version bumping) into separate PR.

test/e2e/e2e_suite_test.go
ahrtr: If there is no strong objection, let's keep it in the workflow check.

abdurrehman107: Hmm. Quick question @ahrtr, if we remove the step where we create a Kind cluster how would the tests function locally? Would there be a cluster where it'd load the docker image to?

abdurrehman107: No, I just mean to ask on how would you run it locally. I normally run `go test` in the e2e directory. Removing `envfuncs.CreateCluster(kindCluster, kindClusterName)`stop that from working would it not?

ahrtr: Will it work if you manually execute `kind create cluster`?

It looks like keeping `envfuncs.CreateCluster(kindCluster, kindClusterName)` removes the manual step. Have you verified this PR locally?

ahrtr: Do you know why the kubebuilder generates the initial e2e test this way? Why it doesn't create the Kind cluster in e2e test code itself?

abdurrehman107: > Have you verified this PR locally?

Yes I did

ahrtr: OK, then let's keep `envfuncs.CreateCluster(kindCluster, kindClusterName`, then we need to remove `kind create cluster ` in https://github.com/kubernetes/test-infra/blob/f414bfaa8700111b528079226ac3e26fbbe61011/config/jobs/etcd/etcd-operator-presubmits.yaml#L53. cc. @ivanvc

ivanvc: > OK, then let's keep `envfuncs.CreateCluster(kindCluster, kindClusterName`, then we need to remove `kind create cluster ` in https://github.com/kubernetes/test-infra/blob/f414bfaa8700111b528079226ac3e26fbbe61011/config/jobs/etcd/etcd-operator-presubmits.yaml#L53. cc. @ivanvc

I can do it after we merge this PR. The e2e test still runs with the manual creation, and other PR would fail without it, so I think it can be done after merging it.

I'll keep an eye on this PR.

abdurrehman107:
> Please move unrelated change (i.e. etcd version bumping) into separate PR.

@ahrtr do you want me to remove the `go.mod` file in my PR, that'll stop from bumping the etcd version but it'll also remove all the new libraries. Or do you want me to only remove the specific lines referring to

abdurrehman107:
@ahrtr, implemented your suggestions. Can you please review again?

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,218 @@ limitations under the License.
...
+			// wait for CertManager to be ready
+			log.Println("Waiting for cert-manager deployment to be available...")
+			if err := wait.For(
+				conditions.New(client.Resources()).DeploymentAvailable("cert-manager-webhook", "cert-manager"),
+				wait.WithTimeout(5*time.Minute),
+				wait.WithInterval(10*time.Second),
+			); err != nil {
+				log.Printf("Timedout while waiting for cert-manager deployment: %s", err)
+				return ctx, err
+			}
```
ahrtr, line 141: Is there any reason not to reuse 
https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/test/utils/utils.go#L64


and 

https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/test/utils/utils.go#L117

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,218 @@ limitations under the License.
...
+			return ctx, nil
+		},
+
+		// set up environment
+		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
```
ahrtr, line 146: It seems that you do not install the CRD? `make install`

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,218 @@ limitations under the License.
...
+				log.Printf("Unable to set working directory: %s", err)
+				return ctx, err
+			}
+
+			return ctx, nil
+		},
+	)
+
+	// Use the Environment.Finish method to define clean up steps
+	testEnv.Finish(
```
ahrtr, line 186: pls organize these into several groups, in the exactly the reverse order as the `Setup`

test/e2e/e2e_suite_test.go
abdurrehman107: ah sorry my bad. I'll fix this.

test/e2e/e2e_suite_test.go
abdurrehman107: Done

test/e2e/e2e_suite_test.go
abdurrehman107: I did not look into these. I've added these and removed my workflow to address this.

abdurrehman107:
Addressed feedback @ahrtr, can you please review again?

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,216 @@ limitations under the License.
...
+	testEnv = env.New()
+	kindClusterName := "etcd-cluster"
+	kindCluster := kind.NewCluster(kindClusterName)
+
+	log.Println("Creating KinD cluster...")
+	origWd, _ := os.Getwd()
+	testEnv.Setup(
+		// create KinD cluster and namespace
+		envfuncs.CreateCluster(kindCluster, kindClusterName),
+		envfuncs.CreateNamespace(namespace),
```
ahrtr, line 50: Pls move it into "setup up environment" as the first step.

And remove the namespace in "cleanup environment" as the last step.

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,216 @@ limitations under the License.
...
+
+		// prepare the resources
+		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+			// change dir for Make file
+			if err := os.Chdir("../../"); err != nil {
+				log.Printf("Unable to set working directory: %s", err)
+				return ctx, err
+			}
```
ahrtr, line 58: We expect users to run the e2e test from the root directory. `go test ./test/e2e/`
```suggestion
```

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,216 @@ limitations under the License.
...
+			return ctx, nil
+		},
+
+		// set up environment
+		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+			// change working directory for Make file
+			if err := os.Chdir("../../"); err != nil {
+				log.Printf("Unable to set working directory: %s", err)
+				return ctx, err
+			}
```
ahrtr, line 142: We expect users to run the e2e test from the root directory. go test ./test/e2e/
```suggestion
```

config/manager/manager.yaml
```diff
@@ -63,7 +63,7 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
-        image: controller:latest
+        image: etcd-operator-controller:v0.1
```
ahrtr, line 66: You change "latest" to "v0.1.0", it means that we will have to maintain it each time when we bump the version. Suggest not to change it, and let's always add tag "latest" for the image as well.

test/e2e/e2e_suite_test.go
abdurrehman107: @ahrtr `make undeploy` deletes the namespace earlier. So deleting it manually errors out. I think we should also remove the step to create the namespace, since `make deploy` creates it eitherway.

ahrtr: The existing e2e test create the namespace and deploy the controller in `BeforeAll`, and undeploy the controller and remove the namespace in `AfterAll`.  Can you sort out why the existing e2e test doesn't run into any issue?

https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/test/e2e/e2e_test.go#L50-L85

abdurrehman107: Just manually ran the Gingko tests by commenting out the part where they delete the namespace. It still worked. From my understanding it is meant to be a fault tolerant way to ensure that the namespace does not remain not-deleted at the end of the test. 

However, I don't think this is needed in our case. Because we dont have to manually pre-create a Kind cluster like with Gingko (we must have a kind cluster already there to run the e2e test in Gingko). In our case we create a destroy the entire Kind cluster, all within the e2e test.

config/manager/manager.yaml
abdurrehman107: @ahrtr we are building the image locally. If I put the tag latest, it for some reason defaults to trying to pull the image from a registry (the default one being DockerHub). I used the tag `current` and it works. Sounds good?

test/e2e/e2e_suite_test.go
abdurrehman107: @ahrtr, I did manually test using `go test ./test/e2e` from the root directory, the tests fail if I do not move out of the directory. The `go test ./test/e2e/` makes them start from the directory test/e2e.

go.mod
```diff
@@ -9,17 +9,18 @@ require (
...
+	go.etcd.io/etcd/client/v3 v3.5.18
+	go.etcd.io/etcd/server/v3 v3.5.18
 	go.uber.org/zap v1.27.0
 	k8s.io/api v0.32.1
 	k8s.io/apimachinery v0.32.1
 	k8s.io/client-go v0.32.1
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	sigs.k8s.io/controller-runtime v0.20.1
+	sigs.k8s.io/e2e-framework v0.5.0
```
ArkaSaha30, line 23: Can we maintain a separate go.mod for test/e2e ...similar to etcd:https://github.com/etcd-io/etcd/tree/main/tests ?
Also, a newer [v0.6.0](https://github.com/kubernetes-sigs/e2e-framework/releases/tag/v0.6.0) is available, can we use that?

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,248 @@ limitations under the License.
...
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/support/kind"
+	"sigs.k8s.io/e2e-framework/support/utils"
 
-	"go.etcd.io/etcd-operator/test/utils"
+	kubebuilder_utils "go.etcd.io/etcd-operator/test/utils"
```
ArkaSaha30, line 34: Does it make more sense to name it as `e2e_utils` or `test_utils`?

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,248 @@ limitations under the License.
...
-// CertManager and Prometheus.
-func TestE2E(t *testing.T) {
-	RegisterFailHandler(Fail)
-	_, _ = fmt.Fprintf(GinkgoWriter, "Starting etcd-operator integration test suite\n")
-	RunSpecs(t, "e2e suite")
-}
+func TestMain(m *testing.M) {
+	testEnv = env.New()
+	kindClusterName := "etcd-cluster"
+	kindCluster := kind.NewCluster(kindClusterName)
```
ArkaSaha30, line 45: Should we keep the option for which version of Kubernetes the kind cluster should spin up with? Default can use the latest available.
Ref: https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L28

test/e2e/e2e_suite_test.go
ahrtr: > Just manually ran the Gingko tests by commenting out the part where they delete the namespace. It still worked.

It works doesn't means it's correct. You create it, then you need to remove it.

It seems that the namespace `etcd-operator-system` automatically applied?  The existing e2e test still manually create the namespace on setup and remove it on teardown as mentioned above. Is it necessary or not? @ArkaSaha30



https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/config/default/kustomization.yaml#L2

Generally, I suggest to following existing e2e pattern. Any improvement can be done later in separate PR.

ahrtr:
I suggest to add a very simple e2e test to do some simple sanity test to ensure the environment is healthy.

test/e2e/e2e_suite_test.go
abdurrehman107: I think `test_utils` would make sense. I'll change it up

test/e2e/e2e_suite_test.go
abdurrehman107: I don't think spinning a cluster with the latest k8s version should be a problem?

config/manager/manager.yaml
ivanvc: I don't think changing this would be a good idea unless we want to change the Kustomize stack that comes from the scaffolding. But, if we want to replace it, I believe you'll need to change the instances from the `Makefile`:

https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/Makefile#L132
https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/Makefile#L151

ivanvc:
@abdurrehman107, could you please rebase/sync your branch with the remote `main`, so it's easier to review without the `go.mod` changes? Thanks :)

test/e2e/e2e_suite_test.go
abdurrehman107: The existing e2e test works because it uses the RunCommand() function mentioned in the utils file. I don't use that specific function because it uses `Gingko`. I've used the RunCommand() function given by the `e2e-framework`.

go.mod
abdurrehman107: Updated the version. Do we want to use a separate `go.mod`? 
cc @ahrtr

config/manager/manager.yaml
abdurrehman107: So do we just wanna leave the image name to be `controller`? 
cc @ivanvc

test/e2e/e2e_suite_test.go
abdurrehman107: I suggest we also update the `test/utils/utils.go` file, because it refers to Gingko. I can take it up in another PR.

test/e2e/e2e_suite_test.go
abdurrehman107: > The existing e2e test still manually create the namespace on setup and remove it on teardown as mentioned above. Is it necessary or not?
Generally, I suggest to following existing e2e pattern. Any improvement can be done later in separate PR.

I think we can add it as of now (it does throw an error that the namespace does not exist), and in a later PR we can remove it in case we feel it is necessary.

abdurrehman107:
> @abdurrehman107, could you please rebase/sync your branch with the remote main, so it's easier to review without the go.mod changes? Thanks :)

The PR keeps failing without the `go.mod` @ivanvc. We add a new framework, the `etcd-framework` which adds to the `go.mod` file. I think it is necessary for the PR.

test/e2e/e2e_suite_test.go
ahrtr: > The existing e2e test works because it uses the Run() function mentioned in the utils file. I don't use that specific function because it uses `Gingko`. I've used the RunCommand() function given by the `e2e-framework`.

This isn't convincing. Shouldn't `RunCommand()` and `command.Run()` have the same or similar effect? I don't think the root cause is clear. Resolving a problem without knowing the root cause is usually dangerous, nor a recommended way.

abdurrehman107: I understand what you're saying Benjamin, and technically they should, but they are using different methods to solve the problem. The `Run()` function takes the input of type `exec.Cmd` while the `RunCommand()` function takes in a string and then uses the `gexe` library to process the command. I'd be happy to hop on a call and go through this together. 

cc @ahrtr @hakman @ArkaSaha30

ahrtr: I just spend about 5 minutes to read the source code of the existing e2e test. 

Is there any difficulty to follow the same pattern ( reuse the `GetProjectDir`)?

https://github.com/etcd-io/etcd-operator/blob/3879c6e4d469c8dd9cf6131bd29f71caea1f8760/test/utils/utils.go#L195-L202

abdurrehman107: Nope, no issues. I just thought implementing the `e2e-framework`'s approach was simpler. However, I get your idea to just replicate the Gingko tests for now, and we can improve in the future. 

No worries, I'll update it.

config/manager/manager.yaml
ivanvc: Is there a particular reason not to? That file is usually just a template. When running `make` (the target that calls line 151), it replaces it with the actual location of the image (i.e., `gcr.io/etcd-development/etcd-operator:v1.2.3`).

But we can leave it as you did and later address this when we do the first release.

test/e2e/e2e_suite_test.go
ivanvc: I think being able to set the version makes sense in cases where the controller doesn't work with the latest version yet. That way, we could set it to the latest supported version.

test/e2e/e2e_suite_test.go
ivanvc: An alternative would be to replace the `make` calls with `make -f../../Makefile`. But that would require the `Makefile` to be agnostic of where it's being executed, and although doable, it will make it dirtier.

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +16,271 @@ limitations under the License.
...
+			return ctx, nil
+		},
+
+		// remove the installed dependencies
+		func(ctx context.Context, c *envconf.Config) (context.Context, error) {
+			// change working directory for Make file
+			if err := os.Chdir("../../"); err != nil {
+				log.Printf("Unable to set working directory: %s", err)
+				return ctx, err
+			}
```
ivanvc, line 245: I think you'll be removing the change in directories, so this may not be relevant. But, it is not required here because there are no calls to `make`.

test/e2e/e2e_suite_test.go
ivanvc: An alternative would be to replace the `make` calls with `make -f../../Makefile`. But that would require the `Makefile` to be agnostic of where it's being executed, and although doable, it will make it dirtier.

test/e2e/e2e_suite_test.go
abdurrehman107: aighty i'll add this.

test/e2e/e2e_suite_test.go
abdurrehman107: Removed this entirely.

config/manager/manager.yaml
abdurrehman107: No worries. I've made it right in this PR.

test/e2e/e2e_suite_test.go
abdurrehman107: Fixed this @ahrtr can you please review again?

abdurrehman107:
Rebased the PR, @ahrtr @ivanvc can you please review?

ahrtr:
Please also squash the commits. thx

test/e2e/e2e_test.go
```diff
@@ -17,291 +17,35 @@ limitations under the License.
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
 
-	"go.etcd.io/etcd-operator/test/utils"
+	apiextensionsV1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
-// namespace where the project is deployed in
-const namespace = "etcd-operator-system"
+// Sample Feature-based test with e2e-framework
+func TestBasicFeature(t *testing.T) {
```
ahrtr, line 30:
```suggestion
func TestClusterHealthy(t *testing.T) {
```

Can you please also ensure the etcd-operator POD is running?

ahrtr (commented):
It's much better now. Thanks for all the effort!

test/e2e/e2e_test.go
abdurrehman107: Addressed in my latest commit.

test/e2e/e2e_suite_test.go
```diff
@@ -17,104 +17,209 @@ limitations under the License.
...
-// CertManager and Prometheus.
-func TestE2E(t *testing.T) {
-	RegisterFailHandler(Fail)
-	_, _ = fmt.Fprintf(GinkgoWriter, "Starting etcd-operator integration test suite\n")
-	RunSpecs(t, "e2e suite")
+func TestMain(m *testing.M) {
+	testEnv = env.New()
+	kindClusterName := "etcd-cluster"
+	kindCluster := kind.NewCluster(kindClusterName)
+	clusterVersion := kind.WithImage("kindest/node:v1.32.0")
```
ahrtr, line 48: It's not good to hard code the version in source code.

I think we can use tool/mod to automatically manage the version, refer to https://github.com/etcd-io/etcd-operator/pull/40

Also the e2e test also need some binaries, (i.e. kubectl, kind, etc, any other dependencies?) are installed beforehand, we can also manage their version using tool/mod.

This can be resolved in a separate PR.

ahrtr:
@ArkaSaha30 @gdasson @hakman do you have any further comment on this PR?

test/e2e/e2e_suite_test.go
abdurrehman107: Sounds good. I can look into this in my next PR. We can raise an issue and address this. 

Looking forward to it.

Makefile
```diff
@@ -70,15 +70,7 @@ test: manifests generate fmt vet envtest ## Run tests.
...
-	@command -v kind >/dev/null 2>&1 || { \
-		echo "Kind is not installed. Please install Kind manually."; \
-		exit 1; \
-	}
-	@kind get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
-	}
-	go test ./test/e2e/ -v -ginkgo.v
+	go test ./test/e2e/
```
ahrtr, line 73:
```suggestion
	go test ./test/e2e/ -v
```

ahrtr:
also cc @ivanvc @jmhbnz

test/e2e/e2e_suite_test.go
ivanvc: I suggest setting this in a configurable variable or constant rather than hardcode the "1.32.0". But with all the rounds already in this pull request, I recommend doing it in a follow-up pull request :)

ivanvc (approved):
LGTM (with a follow-up comment). Thanks, @abdurrehman107.

I suggest that someone with more tenure in the project gives a review, too, as I'm not too familiar with the code base yet :v:

ahrtr:
The e2e test log looks OK!
```
go test ./test/e2e/ -v
2025/02/13 23:51:00 Creating KinD cluster...
2025/02/13 23:51:46 Installing bin tools...
  running: make kustomize
  running: make controller-gen
2025/02/13 23:51:59 Generate manifests...
  running: make manifests
2025/02/13 23:52:00 Building docker image...
  running: make docker-build IMG=etcd-operator:v0.1
2025/02/13 23:53:33 Loading docker image into kind cluster...
2025/02/13 23:53:36 Installing prometheus operator...
  running: kubectl create -f https://github.com/prometheus-operator/prometheus-operator/releases/download/v0.77.1/bundle.yaml
2025/02/13 23:53:39 Installing cert-manager...
  running: kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.16.0/cert-manager.yaml
  running: kubectl wait deployment.apps/cert-manager-webhook --for condition=Available --namespace cert-manager --timeout 5m
2025/02/13 23:53:52 Install crd...
  running: make install
2025/02/13 23:53:53 Deploying components...
2025/02/13 23:53:53 Deploying controller-manager resources...
  running: make deploy IMG=etcd-operator:v0.1
2025/02/13 23:53:55 Waiting for controller-manager deployment to be available...
=== RUN   TestClusterHealthy
=== RUN   TestClusterHealthy/etcd-operator-controller
=== RUN   TestClusterHealthy/etcd-operator-controller/ensure_the_etcd-operator_pod_is_running
--- PASS: TestClusterHealthy (10.01s)
    --- PASS: TestClusterHealthy/etcd-operator-controller (10.01s)
        --- PASS: TestClusterHealthy/etcd-operator-controller/ensure_the_etcd-operator_pod_is_running (10.01s)
PASS
2025/02/13 23:54:25 Finishing tests, cleaning cluster ...
2025/02/13 23:54:25 Undeploy etcd controller...
  running: make undeploy ignore-not-found=true
2025/02/13 23:54:31 Uninstalling crd...
  running: make uninstall ignore-not-found=true
2025/02/13 23:54:32 Destroying namespace...
2025/02/13 23:54:32 failed to delete namespace: delete namespace func: namespaces "etcd-operator-system" not found
2025/02/13 23:54:32 Removing dependencies...
  running: kubectl delete -f https://github.com/prometheus-operator/prometheus-operator/releases/download/v0.77.1/bundle.yaml
  running: kubectl delete -f https://github.com/jetstack/cert-manager/releases/download/v1.16.0/cert-manager.yaml
2025/02/13 23:54:49 Destroying cluster...
ok  	go.etcd.io/etcd-operator/test/e2e	229.363s
wrapper.sh] [TEST] Test Command exit code: 0
wrapper.sh] [CLEANUP] Waiting 30 seconds for pods stopped with terminationGracePeriod:30
wrapper.sh] [CLEANUP] Cleaning up after Docker in Docker ...
cecd983[119](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd-operator/62/pull-etcd-operator-test-e2e/1890186039805349888#1:build-log.txt%3A119)77
wrapper.sh] [CLEANUP] Waiting for docker to stop for 30 seconds
Stopping Docker: dockerProgram process in pidfile '/var/run/docker-ssd.pid', 1 process(es), refused to die.
wrapper.sh] [CLEANUP] Done cleaning up after Docker in Docker.
================================================================================
wrapper.sh] Exiting 0
```

One followup: https://github.com/etcd-io/etcd-operator/pull/62#discussion_r1954914492

Thanks for the effort again!

ahrtr (approved):

PR #66, opened 2025-02-12
Move certificate interface into dedicated package to avoid cycle dependency

ahrtr:
Each provider implementation (i.e. auto, cert-manager) needs to depend on the interface. The certificate framework depends on the interface, but it shouldn't depend on the each implementation based on the [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). 

Both the framework and each implementation shouldn't depend on each other, instead both of them should only depend on the interface(abstract). So move the interface into a dedicated package.

cc @ArkaSaha30 @hakman @jmhbnz @justinsb @neolit123

### DIFF
--- a/pkg/certificate/certificate.go
+++ b/pkg/certificate/certificate.go
@@ -1,0 +1,26 @@
+package certificate
+
+import (
+	"fmt"
+
+	certInterface "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+)
+
+type ProviderType string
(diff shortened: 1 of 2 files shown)


ahrtr:
cc @ivanvc as well

ArkaSaha30:
Thank you!! I was facing cyclic import issue initially, with this PR it is fixed 👍

PR #72, opened 2025-02-15
Remove unnecessary make command to install tools in e2e tests

ivanvc:
The make `kustomize` and `controller-gen` targets are prerequisites of the targets later called in the e2e tests (`controller-gen` from `manifests`, and `kustomize` from `install`), so there's no need to trigger them manually.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -64,21 +64,9 @@ func TestMain(m *testing.M) {
 
 		// prepare the resources
 		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
-			log.Println("Installing bin tools...")
-			cmd := exec.Command("make", "kustomize")
-			if _, err := test_utils.Run(cmd); err != nil {
-				log.Printf("Failed to install kustomize binary: %s", err)
-				return ctx, err
-			}
-			cmd = exec.Command("make", "controller-gen")
-			if _, err := test_utils.Run(cmd); err != nil {
-				log.Printf("Failed to install controller-gen binary: %s", err)
-				return ctx, err
-			}
-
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -64,21 +64,9 @@ func TestMain(m *testing.M) {
...
-			cmd = exec.Command("make", "controller-gen")
-			if _, err := test_utils.Run(cmd); err != nil {
-				log.Printf("Failed to install controller-gen binary: %s", err)
-				return ctx, err
-			}
-
 			// gen manifest files
 			log.Println("Generate manifests...")
-			cmd = exec.Command("make", "manifests")
+			cmd := exec.Command("make", "manifests")
```
ahrtr, line 69: It seems that even `make manifests" isn't needed.  Feel free to raise a new PR to remove & test it. thx 

https://github.com/etcd-io/etcd-operator/blob/4442c81fdab5735987e1f7e041bf404ac398c89a/Makefile#L134

ahrtr (approved):

test/e2e/e2e_suite_test.go
ivanvc: I see. `manifests` is a prereq from `test-e2e`. So, we could also call `docker-build` by setting it as a prereq for `test-e2e`. I'll open a pull request with this change, too.

ivanvc: Ah, nevermind. It is a pre-requisite for `install`, as you mentioned. It doesn't make sense to have it as prereq for `test-e2e`, because if the tests are run with `go test`, it won't have the manifests. It makes more sense to remove it from `test-e2e`.

PR #71, opened 2025-02-15
Bump go to 1.24.0

ivanvc:
Upgrades Go to 1.24.0.

Part of etcd-io/etcd#19417.

PR #76, opened 2025-02-18
Add ‘kind’ to pinned tool dependencies and ensure it’s on PATH for E2e tests

abdurrehman107:
What This PR Does:

- Pins the `sigs.k8s.io/kind` dependency in our `tools/mod` to ensure a consistent, versioned installation of the kind CLI.
- Adds a Makefile target (kind) to install the pinned version locally in ./bin.
- Updates the test-e2e Makefile target to depend on kind and prepend ./bin to PATH, ensuring the e2e tests always use the correct version of kind.

Why We Need It:

Previously, the e2e tests either assumed kind was already installed or auto-installed it to `$GOPATH/bin or $HOME/go/bin`. This caused inconsistencies, especially if those paths weren’t on the user’s or CI’s PATH.
By pinning kind in `tools/mod` and installing it to `./bin`, we ensure that everyone (developers, CI, etc.) uses the same version.


tools/mod/go.mod
```diff
@@ -12,9 +12,12 @@ require (
 	sigs.k8s.io/kustomize/kustomize/v5 v5.6.0
 )
 
+require sigs.k8s.io/kind v0.27.0
```
hakman, line 15: Should we just move it to the `required` block above?

tools/mod/tools.go
```diff
@@ -26,4 +26,5 @@ import (
 	_ "sigs.k8s.io/controller-runtime/tools/setup-envtest"
 	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"
 	_ "sigs.k8s.io/kustomize/kustomize/v5"
+	_ "sigs.k8s.io/kind"
```
ahrtr, line 29: pls sort the import items in alphabetical order.

We might want to import the similar verification as etcd. cc @ivanvc

tools/mod/go.mod
ahrtr: Use `go get` command instead of manually updating the go.mod file

abdurrehman107:
addressed comments, can you guys review again? 
@hakman @ahrtr

ahrtr (approved):

tools/mod/tools.go
ivanvc: The `goimports` linter is enabled in this project. However, we're not linting `tools/mod`.

With [Go 1.24 and the `tools` command](https://tip.golang.org/doc/go1.24#tools), it's not required still to do the `tools/mod` workaround to manage these dependency versions. I think it would be better to roll out 1.24 tools rather than start linting `tools/mod`.

I already have a PR to bump to 1.24 here: https://github.com/etcd-io/etcd-operator/pull/71. It has the tool functionality, but I'll break down that pull request in two.

Issue #77, opened 2025-02-18
Remove ginkgo from test/utils/utiles.go


ahrtr:
Try to replace ginkgo with golang standard test uitilities or [e2e-framework](sigs.k8s.io/e2e-framework)

ahrtr:
Resolved in https://github.com/etcd-io/etcd-operator/pull/79

PR #78, opened 2025-02-19
Remove ginkgo dependency in test/utils

nayihz:
Remove ginkgo dependency in test/utils.
Fixes: #77 .


ahrtr:
#77 has already been resolved in #79 , thx

Issue #19, opened 2025-01-06
Add e2e test cases of managing etcd cluster


ahrtr:
We need to add e2e test cases to verify the etcd-operator's functionalities of managing etcd clusters:
- create & delete 1 member etcd cluster
- create & delete 3 member etcd cluster
- scale in & out etcd cluster
- sanity test against the etcd cluster
  - read & write k/v
  - member list
  - health check

Kubebuilder automatically generates e2e test using [ginkgo](github.com/onsi/ginkgo), but personally I am not a big fan of ginkgo. It creates some syntax sugars on top of golang standard test, and organize the test in a hierarchical structure. When the test file gets bigger and bigger, it will become more and more unreadable.  

So it would be great if we can add the e2e test case just using the golang standard test utilities.

abdurrehman107:
I can take this up guys.

ahrtr:
The https://github.com/etcd-io/etcd-operator/pull/17 has been merged. This task isn't blocked anymore.

abdurrehman107:
Sounds good. I'll get started on this today.

ahrtr:
Note, essentially, the e2e test automates the manual human test process, ensuring comprehensive coverage and repeatability. For each e2e test cases, we should be able to manually run/reproduce it as well.

abdurrehman107:
Hey there @ahrtr, I've made a rough scaffold of how the tests are going to look like. It is still in my working branch, I'd really appreciate if you can take a look and give feedback. 

The current pushed code is able to: 

1. build a Kind cluster, 
2. deploy the controller to the cluster, 
3. deploy the etcd cluster crd and resource to the cluster, 
4. scale the etcd cluster from 1 to 3

I am getting started with the sanity testing you mentioned. I'd really appreciate feedback.

ahrtr:
@abdurrehman107 Please raise a PR to get feedback, thx

ahrtr:
Also please let's do it step by step.
- Step 1: let's just rewrite the existing e2e test;
- Step 2: add more tests

abdurrehman107:
Aighty I'll raise a PR. @ahrtr can you elaborate on what you mean by rewriting the existing e2e test? Do you mean that I should rewrite the tests in Ginkgo (by default) to e2e framework?

ahrtr:
> Do you mean that I should rewrite the tests in Ginkgo (by default) to e2e framework?

[...]

PR #84, opened 2025-02-28
Remove usage of ginkgo entirely

abdurrehman107:
This PR refers to the [idea](https://github.com/etcd-io/etcd-operator/issues/19#issuecomment-2670923301) to remove `Ginkgo` entirely from the `etcd-operator` project. 

We removed Ginkgo instances from: 
- etcd controller tests
- golangci.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
+	ENVTEST_K8S_VERSION=$(ENVTEST_K8S_VERSION) \
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # TODO(user): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
--- a/internal/controller/etcdcluster_controller_test.go
+++ b/internal/controller/etcdcluster_controller_test.go
@@ -17,10 +17,8 @@ limitations under the License.
 package controller
 
 import (
-	"context"
+	"testing"
 
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -30,55 +28,64 @@ import (
 	operatorv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
-var _ = Describe("EtcdCluster Controller", func() {
-	Context("When reconciling a resource", func() {
-		const resourceName = "test-resource"
+func TestControllerReconcile(t *testing.T) {
+	ctx := t.Context()
 
-		ctx := context.Background()
+	const resourceName = "test-etcd-cluster"
+	typeNamespacedName := types.NamespacedName{
+		Name:      resourceName,
+		Namespace: "default",
+	}
 
-		typeNamespacedName := types.NamespacedName{
-			Name:      resourceName,
-			Namespace: "default", // TODO(user):Modify as needed
+	etcdcluster := &operatorv1alpha1.EtcdCluster{}
+
+	t.Log("Checking for EtcdCluster resource")
+	err := k8sClient.Get(ctx, typeNamespacedName, etcdcluster)
+	if err != nil && errors.IsNotFound(err) {
+		t.Log("Creating EtcdCluster resource")
+		resource := &operatorv1alpha1.EtcdCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      resourceName,
+				Namespace: "default",
+			},
 		}
-		etcdcluster := &operatorv1alpha1.EtcdCluster{}
+		if createErr := k8sClient.Create(ctx, resource); createErr != nil {
+			t.Fatalf("Failed to create EtcdCluster resource: %v", createErr)
+		}
+	} else if err != nil {
+		t.Fatalf("Failed to get EtcdCluster resource: %v", err)
+	}
 
-		BeforeEach(func() {
-			By("creating the custom resource for the Kind EtcdCluster")
-			err := k8sClient.Get(ctx, typeNamespacedName, etcdcluster)
-			if err != nil && errors.IsNotFound(err) {
-				resource := &operatorv1alpha1.EtcdCluster{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      resourceName,
-						Namespace: "default",
-					},
-					// TODO(user): Specify other spec details if needed.
-				}
-				Expect(k8sClient.Create(ctx, resource)).To(Succeed())
+	// Defer a cleanup function to remove the resource after the test finishes.
+	defer func() {
+		t.Log("Cleaning up the EtcdCluster resource")
+		resource := &operatorv1alpha1.EtcdCluster{}
+		if getErr := k8sClient.Get(ctx, typeNamespacedName, resource); getErr != nil {
+			t.Errorf("Failed to get EtcdCluster resource before deletion: %v", getErr)
+		} else {
+			if deleteErr := k8sClient.Delete(ctx, resource); deleteErr != nil {
+				t.Errorf("Failed to delete EtcdCluster resource: %v", deleteErr)
 			}
-		})
+		}
+	}()
 
-		AfterEach(func() {
-			// TODO(user): Cleanup logic after each test, like removing the resource instance.
-			resource := &operatorv1alpha1.EtcdCluster{}
-			err := k8sClient.Get(ctx, typeNamespacedName, resource)
-			Expect(err).NotTo(HaveOccurred())
+	reconciler := &EtcdClusterReconciler{
+		Client: k8sClient,
+		Scheme: k8sClient.Scheme(),
+	}
 
-			By("Cleanup the specific resource instance EtcdCluster")
-			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
-		})
-		It("should successfully reconcile the resource", func() {
-			By("Reconciling the created resource")
-			controllerReconciler := &EtcdClusterReconciler{
-				Client: k8sClient,
-				Scheme: k8sClient.Scheme(),
-			}
-
-			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
-				NamespacedName: typeNamespacedName,
-			})
-			Expect(err).NotTo(HaveOccurred())
-			// TODO(user): Add more specific assertions depending on your controller's reconciliation logic.
-			// Example: If you expect a certain status condition after reconciliation, verify it here.
-		})
+	_, err = reconciler.Reconcile(ctx, reconcile.Request{
+		NamespacedName: typeNamespacedName,
 	})
-})
+	if err != nil {
+		t.Fatalf("Reconciliation failed: %v", err)
+	}
+
+	// TODO: Add more specific checks (e.g., verifying status conditions or created resources).
+	// For example:
+	// updated := &operatorv1alpha1.EtcdCluster{}
+	// if err := k8sClient.Get(ctx, typeNamespacedName, updated); err != nil {
+	//     t.Fatalf("Failed to retrieve updated resource: %v", err)
+	// }
+	// // Validate updated fields or status
+}
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -17,80 +17,78 @@ limitations under the License.
 package controller
 
 import (
-	"context"
 	"fmt"
+	"log"
+	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
 
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
-
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	operatorv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 	// +kubebuilder:scaffold:imports
 )
 
-// These tests use Ginkgo (BDD-style Go testing framework). Refer to
-// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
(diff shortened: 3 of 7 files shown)


internal/controller/suite_test.go
```diff
@@ -37,60 +35,76 @@ import (
...
+
+// placeholder for your actual test logic.
+func TestControllers(t *testing.T) {
+	t.Run("ExampleTest", func(t *testing.T) {
+		// Example: check that the client is not nil
+		if k8sClient == nil {
+			t.Fatalf("k8sClient is nil, expected a valid client")
+		}
+	})
+}
```
ahrtr, line 110:
```suggestion
```

internal/controller/suite_test.go
```diff
@@ -37,60 +35,76 @@ import (
...
-
-var _ = BeforeSuite(func() {
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
```
ahrtr, line 44: pls consider removing or minimizing the global variables

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +29,69 @@ import (
...
+			}
+		}()
 
-			By("Cleanup the specific resource instance EtcdCluster")
-			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
-		})
-		It("should successfully reconcile the resource", func() {
-			By("Reconciling the created resource")
-			controllerReconciler := &EtcdClusterReconciler{
+		t.Run("should successfully reconcile the resource", func(t *testing.T) {
```
ahrtr, line 75: try not to embed the sub test.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +29,69 @@ import (
 	operatorv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
-var _ = Describe("EtcdCluster Controller", func() {
-	Context("When reconciling a resource", func() {
-		const resourceName = "test-resource"
+func TestEtcdClusterController(t *testing.T) {
+	ctx := context.Background()
```
ivanvc, line 33: I believe we can replace this context with regular `T` contexts: 
```suggestion
	ctx := t.Context()
```

internal/controller/suite_test.go
```diff
@@ -37,60 +35,76 @@ import (
...
-	By("tearing down the test environment")
+	// Teardown after tests
 	cancel()
-	err := testEnv.Stop()
-	Expect(err).NotTo(HaveOccurred())
-})
+	if err := testEnv.Stop(); err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to stop test environment: %v\n", err)
+		os.Exit(1)
+	}
```
ahrtr, line 96: pls do this in a defer function

internal/controller/suite_test.go
```diff
@@ -37,60 +35,76 @@ import (
...
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
 
-	ctx, cancel = context.WithCancel(context.TODO())
+func TestMain(m *testing.M) {
+	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+	ctx, cancel = context.WithCancel(context.Background())
```
ArkaSaha30, line 48: I believe the same can be applied here: https://github.com/etcd-io/etcd-operator/pull/84/files#r1976743587

internal/controller/suite_test.go
```diff
@@ -37,60 +35,76 @@ import (
...
 	}
 
+	// Start the envtest environment.
 	var err error
-	// cfg is defined in this file globally.
 	cfg, err = testEnv.Start()
-	Expect(err).NotTo(HaveOccurred())
-	Expect(cfg).NotTo(BeNil())
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to start test environment: %v\n", err)
```
ArkaSaha30, line 64: Just trying to understand...why we are using `Fprintf` and not `Sprintf`?

ArkaSaha30: Nvm, got my answer 👍

internal/controller/suite_test.go
abdurrehman107: @ArkaSaha30 I don't think we can remove this, because `testing.M` does not provide a Context() function. We can only do that with the test functions using `testing.T`. Any suggestions here @ivanvc

abdurrehman107:
@ahrtr @ArkaSaha30 @ivanvc: Can you folks please review again?

internal/controller/suite_test.go
```diff
@@ -37,60 +34,72 @@ import (
...
 
-})
+	// Teardown after tests
+	defer func() {
+		cancel()
+		if err := testEnv.Stop(); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to stop test environment: %v\n", err)
+			os.Exit(1)
+		}
+	}()
```
ahrtr, line 101: pls put the defer func right close to `testEnv.Start`

internal/controller/suite_test.go
```diff
@@ -37,60 +34,72 @@ import (
...
-
-func TestControllers(t *testing.T) {
-	RegisterFailHandler(Fail)
-
-	RunSpecs(t, "Controller Suite")
-}
+var (
+	ctx       context.Context
+	cancel    context.CancelFunc
+	k8sClient client.Client
```
ahrtr, line 40: can you try to remove the global variable `ctx` and `cancel`?

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,69 @@ import (
...
+				t.Fatalf("Failed to create EtcdCluster resource: %v", createErr)
+			}
+		} else if err != nil {
+			t.Fatalf("Failed to get EtcdCluster resource: %v", err)
+		}
 
-		AfterEach(func() {
-			// TODO(user): Cleanup logic after each test, like removing the resource instance.
+		// Defer a cleanup function to remove the resource after the test finishes.
+		defer func() {
```
ahrtr, line 60: The deferred function get executed when the sub test complete. It means that the resource has already been removed before your next sub test "reconcile".

Pls just remove the sub test. ( I mean just remove `t.Run`, and keep the test code in the parent test function directly)

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,67 @@ import (
...
-			if err != nil && errors.IsNotFound(err) {
-				resource := &operatorv1alpha1.EtcdCluster{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      resourceName,
-						Namespace: "default",
-					},
-					// TODO(user): Specify other spec details if needed.
-				}
-				Expect(k8sClient.Create(ctx, resource)).To(Succeed())
+		t.Log("Creating or retrieving the EtcdCluster resource")
```
ArkaSaha30, line 43: Small nit, at this step can we have some log message like "Checking for EtcdCluster resource"

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,67 @@ import (
...
+		err := k8sClient.Get(ctx, typeNamespacedName, etcdcluster)
+		if err != nil && errors.IsNotFound(err) {
```
ArkaSaha30, line 45: And log here: "Creating EtcdCluster resource"

ArkaSaha30 (commented):
Small nit, rest looks good.
Thank you!

internal/controller/suite_test.go
```diff
@@ -37,60 +34,70 @@ import (
...
-}
+var (
+	k8sClient client.Client
+)
 
-var _ = BeforeSuite(func() {
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+func TestMain(m *testing.M) {
+	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+	_, cancel := context.WithCancel(context.Background())
```
ahrtr, line 43: what's the use of this line of code?

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,67 @@ import (
...
+	typeNamespacedName := types.NamespacedName{
+		Name:      resourceName,
+		Namespace: "etcd-operator-system",
+	}
 
-		typeNamespacedName := types.NamespacedName{
-			Name:      resourceName,
-			Namespace: "default", // TODO(user):Modify as needed
-		}
+	t.Run("When reconciling a resource", func(t *testing.T) {
```
ahrtr, line 40: There is no need to create a sub test.

internal/controller/suite_test.go
abdurrehman107: I thought we'd need context for a future test maybe. Like that was my idea. But since we are not storing the context in a variable, I don't think there is a point to have the line. I'll get rid of it.

abdurrehman107: we don't use the context, hence the `context` variable is empty, however we do use the `cancel` function later in the defer() function block.

ahrtr: It's weird that you don't use context, but use cancel.  They are supposed to be used together.

ivanvc: Maybe the `cancel()` has no use at all, as it doesn't have a context to cancel. Then, removing it should be okay(?).

abdurrehman107: Hmm. Possibly yes. I added the context with the idea that it'll be needed in the future. I'll remove it for now, and if we need it in the future we can add it again.

ahrtr:
Pls do not merge main into your dev branch, rebase your dev branch instead.

abdurrehman107:
@ahrtr @ivanvc @ArkaSaha30 pushed new changes team, can we please review again?

Makefile
```diff
@@ -61,7 +61,8 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	ENVTEST_K8S_VERSION=$(ENVTEST_K8S_VERSION) \
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```
ahrtr, line 65: I intentionally added the `-v` in https://github.com/etcd-io/etcd-operator/pull/88, not sure why you remove it now?
```suggestion
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,66 @@ import (
 	operatorv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
-var _ = Describe("EtcdCluster Controller", func() {
-	Context("When reconciling a resource", func() {
-		const resourceName = "test-resource"
+func TestEtcdClusterController(t *testing.T) {
```
ahrtr, line 31:
```suggestion
func TestControllerReconcile(t *testing.T) {
```

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,66 @@ import (
...
+	}
+
+	// Defer a cleanup function to remove the resource after the test finishes.
+	defer func() {
+		t.Log("Cleaning up the EtcdCluster resource")
+		resource := &operatorv1alpha1.EtcdCluster{}
+		if getErr := k8sClient.Get(ctx, typeNamespacedName, resource); getErr != nil {
+			if !errors.IsNotFound(getErr) {
+				t.Errorf("Failed to get EtcdCluster resource before deletion: %v", getErr)
 			}
```
ahrtr, line 66: The resource/object should always be present, otherwise there must be something wrong.
```suggestion
			t.Errorf("Failed to get EtcdCluster resource before deletion: %v", getErr)
```

internal/controller/etcdcluster_controller_test.go
```diff
@@ -30,55 +28,66 @@ import (
...
+		t.Fatalf("Reconciliation failed: %v", err)
+	}
+
+	// TODO: Add more specific checks (e.g., verifying status conditions or created resources).
+	// For example:
+	// updated := &operatorv1alpha1.EtcdCluster{}
+	// if err := k8sClient.Get(ctx, typeNamespacedName, updated); err != nil {
+	//     t.Fatalf("Failed to retrieve updated resource: %v", err)
+	// }
+	// // Validate updated fields or status
```
ahrtr, line 92: If we need to test anything else, let's add new test case.
```suggestion
```

internal/controller/suite_test.go
```diff
@@ -37,60 +33,68 @@ import (
...
 
+	// Start the envtest environment.
 	var err error
-	// cfg is defined in this file globally.
-	cfg, err = testEnv.Start()
-	Expect(err).NotTo(HaveOccurred())
-	Expect(cfg).NotTo(BeNil())
+	cfg, err := testEnv.Start()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to start test environment: %v\n", err)
```
ahrtr, line 63: Can you use the golang std package `log` to print the message? The benefit is that the message contains timestamp.

Please try to replace all fmt.Print/fmt.FPrint with log in this PR, thx

Makefile
abdurrehman107: sorry that must've gotten removed by mistake. I'll add it again.

internal/controller/etcdcluster_controller_test.go
abdurrehman107: Do you have a sample test case in mind? If not, we can remove this comment and proceed with merging. We can add specific test cases in separate PRs if needed.

internal/controller/suite_test.go
```diff
@@ -17,80 +17,78 @@ limitations under the License.
...
-	RunSpecs(t, "Controller Suite")
-}
+var (
+	k8sClient client.Client
+)
 
-var _ = BeforeSuite(func() {
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+func TestMain(m *testing.M) {
+	logger := log.Default()
```
ahrtr, line 40: This seems unnecessary, you can use log.Printxx directly. BUt not a big deal

ahrtr (approved):
LGTM

THANKS !

Issue #86, opened 2025-03-02
Support customizing etcd options


SDBrett:
I'd like to implement this enhancement from the v0.2.0 roadmap.

My current thinking of how the user would provide the options is to add `.spec.etcdOptions` to the CR where the user can then specify the options needed.

```yaml
spec:
  etcdOptions:
    maxWals: 5
    election-timeout: 1000
    max-snapshots: 2
```

There are a couple implementation specifics that I do need clarification on:
- Should certain options not be configurable use this method? For example
  - Member name presents a challenge due to the dynamic name of pods and potential to scale the number of members
  - Directory paths like `data-dir` as these are opinionated via the controller with management of the STS
- If there are options which are not desirable to be configured in this manner, what action should be taken?
  - Log warnings and ignore
  - Throw error and halt

Does ETCD need to be restarted when a configuration update is made or does it handle that automatically?

Thanks

jmhbnz:
Thanks for volunteering to work on this for the `0.2.0` release @SDBrett 

/assign @SDBrett

PR #93, opened 2025-03-05
Add Etcd Options

SDBrett:
Link Issue #86 

Add the option for users to provide ETCD configuration options to the ETCD deployment.

Added the options to the existing test suite and tests continued to pass

Tested on local instance of K8S and the options configured are passed as EnvVars to the pods running

Also added file in the docs directory to help with future documentation efforts


jmhbnz (commented):
Hey @SDBrett - Thanks for working on this!  To ensure your contribution can be accepted can you please sign your commit so the Developer Certificate of Origin check passes, i.e:

```bash
git reset --soft HEAD~1
git commit --signoff --message "<message>"
git push --force
```

internal/controller/utils.go
```diff
@@ -149,6 +149,17 @@ func createOrPatchStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1a
 		},
 	}
 
+	if len(ec.Spec.EtcdOptions) > 0 {
+		for i := 0; i < len(podSpec.Containers); i++ {
+			if podSpec.Containers[0].Name == "etcd" {
```
frederiko, line 154: I wonder if the container name must become a constant. Yes, there are other parts of the code that could benefit from that.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -30,6 +30,8 @@ type EtcdClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// etcd configuration options are passed as environmental vars to the etcd container, refer to etcd documentation for configuration options application for the version of etcd being used.
+	EtcdOptions []EtcdOption `json:"etcdOptions,omitempty"`
```
frederiko, line 34: Aren't these options "optional"? Wonder it `// +optional` Kubebuilder marker makes sense here

internal/controller/utils.go
frederiko, line 154: Do you expect that `etcd` container will always be the first one, index 0?

SDBrett: It makes sense to have it as a constant, I wasn't sure if doing that would fit into the scope of this PR.

As a side note, I'll be creating an issue soon to look at refactoring how the container is defined for improved testability and cleaner code for things like this feature.

SDBrett: Good pick up.

This was meant to handle a scenario where the etcd container isn't at index 0, instead I botched it. Will push an update shortly

api/v1alpha1/etcdcluster_types.go
SDBrett: I thought that using `omitempty` would make it optional, but I'll test with the `// +optional` marker

SDBrett: I did some testing here, using `omitempty` does make the field optional. With the current configuration the CRD has EtcdOptions as an optional parameter. 

Using `omitempty` to make fields optional does appear to be the convention for this project so I've left this as is

internal/controller/utils.go
SDBrett: Updated loop to fix issue

internal/controller/utils.go
```diff
@@ -149,6 +149,17 @@ func createOrPatchStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1a
 		},
 	}
 
+	if len(ec.Spec.EtcdOptions) > 0 {
+		for i := range podSpec.Containers {
+			if podSpec.Containers[i].Name == "etcd" {
+				for _, v := range ec.Spec.EtcdOptions {
+					podSpec.Containers[i].Env = append(podSpec.Containers[i].Env, corev1.EnvVar{Name: v.Name, Value: v.Value})
```
ahrtr, line 156: I think we have two options:
- append more args directly (see line 103)
- use env variable as this PR does

Either way, it'd be better to let users to configure the item as the flag (i.e. "`data-dir`"),  and then,
- either append it to the container's Args,
- or convert it into environment (adding prefix "ETCD_"). refer to https://etcd.io/docs/v3.6/op-guide/configuration/

Which one do you prefer? Personally, the first option seems more direct and simpler, so better?

ahrtr: If you go for option 2, then you need to convert the flag to env variable, 

refer to https://github.com/etcd-io/etcd/blob/ffb6ff9acf049b6b864955bea46ac6e33c54d9c9/pkg/flags/flag.go#L69-L71

The prefix is `ETCD`.

SDBrett: If users provide them as flags then I would just go with appending the args array, don’t see the point of converting to env vars.

I’ll make the update when able.

SDBrett: @ahrtr I've made changes to use args instead of env vars and added some unit tests for the additional functions.

config/crd/bases/operator.etcd.io_etcdclusters.yaml
```diff
@@ -39,6 +39,13 @@ spec:
           spec:
             description: EtcdClusterSpec defines the desired state of EtcdCluster.
             properties:
+              etcdOptions:
+                description: etcd configuration options are passed as environmental
+                  vars to the etcd container, refer to etcd documentation for configuration
+                  options application for the version of etcd being used.
+                items:
+                  type: string
+                type: array
```
jmhbnz, line 48: Nit: Minor wording tweak to align with direction

```suggestion
              etcdOptions:
                description: etcd configuration options are passed as command line
                 arguments to the etcd container, refer to etcd documentation for 
                 configuration options applicable for the version of etcd being used.
                items:
                  type: string
                type: array
```

docs/configuring-etcd-options.md
```diff
@@ -0,0 +1,5 @@
+# Configuration Etcd Options
+
+Additional configuration options can be set using the `.spec.etcdOptions` field. The options provided are passed as command line arguments to the etcd container.
+
+Information about the different configuration options is available from the etcd documentation page here: https://etcd.io/docs/v3.5/op-guide/configuration/. Please note that this link is for version 3.5 of etcd and you may need to adjust it for the version of etcd being run.
```
jmhbnz, line 5: Suggest we use permalink

```suggestion
Information about the different configuration options is available from the etcd documentation page here: https://etcd.io/docs/latest/op-guide/configuration. Please note that this link is for version 3.5 of etcd and you may need to adjust it for the version of etcd being run.
```

internal/controller/utils_test.go
```diff
@@ -65,6 +66,10 @@ func TestReconcileStatefulSet(t *testing.T) {
 		Spec: ecv1alpha1.EtcdClusterSpec{
 			Size:    3,
 			Version: "3.5.17",
+			// EtcdOptions: []string{
+			// 	"--max-wals=7",
+			// 	"--discovery-failbox=proxy",
+			// },
```
SDBrett, line 72: Left commented out for the moment while validating that TestReconcileStatefulSet is running properly

config/crd/bases/operator.etcd.io_etcdclusters.yaml
SDBrett: Thanks for the pickup, made the updates

docs/configuring-etcd-options.md
SDBrett: Updated link and added a paragraph about overrides

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
 	return getStatefulSet(ctx, c, ec.Name, ec.Namespace)
 }
 
+func setDefaultArgs(name string) []string {
```
ahrtr, line 81:
```suggestion
func defaultArgs(name string) []string {
```

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
...
+		defaultArgs = slices.Delete(defaultArgs, index, index+1)
+	}
+	return defaultArgs
+}
+
+func getArgName(s string) string {
+	idx := strings.Index(s, "=")
+
+	if idx != -1 {
+		return s[:idx]
```
ahrtr, line 114: Please try to remove the leading and trailing space for the returned string using `strings.TrimSpace`

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
...
+		"--listen-client-urls=http://0.0.0.0:2379", // TODO: only listen on 127.0.0.1 and host IP
+		fmt.Sprintf("--initial-advertise-peer-urls=http://$(POD_NAME).%s.$(POD_NAMESPACE).svc.cluster.local:2380", name),
+		fmt.Sprintf("--advertise-client-urls=http://$(POD_NAME).%s.$(POD_NAMESPACE).svc.cluster.local:2379", name),
+	}
+}
+
+func RemoveStringFromSlice(defaultArgs []string, s string) []string {
+	var compareStr string
+	var index int
+	exists := false
```
ahrtr, line 94:
```suggestion
```

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
...
+
+	for i := range defaultArgs {
+		compareStr = getArgName(defaultArgs[i])
+		if compareStr == s {
+			index = i
+			exists = true
+			break
+		}
```
ahrtr, line 102:
```suggestion
		defaultArg := getArgName(defaultArgs[i])
		if compareStr == customArg {
			defaultArgs = slices.Delete(defaultArgs, index, index+1)
			break
		}
```

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
...
+	}
+	if exists {
+		defaultArgs = slices.Delete(defaultArgs, index, index+1)
+	}
```
ahrtr, line 106:
```suggestion
```

internal/controller/utils.go
```diff
@@ -77,6 +78,66 @@ func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha
...
+	return []string{
+		"--name=$(POD_NAME)",
+		"--listen-peer-urls=http://0.0.0.0:2380",   // TODO: only listen on 127.0.0.1 and host IP
+		"--listen-client-urls=http://0.0.0.0:2379", // TODO: only listen on 127.0.0.1 and host IP
+		fmt.Sprintf("--initial-advertise-peer-urls=http://$(POD_NAME).%s.$(POD_NAMESPACE).svc.cluster.local:2380", name),
+		fmt.Sprintf("--advertise-client-urls=http://$(POD_NAME).%s.$(POD_NAMESPACE).svc.cluster.local:2379", name),
+	}
+}
+
+func RemoveStringFromSlice(defaultArgs []string, s string) []string {
```
ahrtr, line 91:
```suggestion
func RemoveStringFromSlice(defaultArgs []string, customArg string) []string {
```

internal/controller/utils_test.go
```diff
@@ -503,3 +508,146 @@ func TestApplyEtcdClusterState(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			result := createArgs(tt.clusterName, tt.etcdOptions)
+			assert.Equal(t, tt.expectedResult, result)
+		})
+	}
+
+}
+func TestPodSpecArgs(t *testing.T) {
```
ahrtr, line 605: The above unit test case `TestCreatingArgs`  is already good enough, I don't think this case has much value. I'd love to see an e2e test.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -30,6 +30,8 @@ type EtcdClusterSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// etcd configuration options are passed as command line arguments to the etcd container, refer to etcd documentation for configuration options applicable for the version of etcd being used.
+	EtcdOptions []string `json:"etcdOptions,omitempty"`
```
ahrtr, line 34: minor comment: could you please move this new field at the end of the struct?

internal/controller/utils.go
SDBrett: I've been  thinking about this and think it better to make the function more generic. Changing line 91 to
`func RemoveStringFromSlice(s []string, str string) []string` would make the inputs better for use cases outside of custom args.

api/v1alpha1/etcdcluster_types.go
SDBrett: Just to confirm, you would like the struct to look like the below?

```
type EtcdClusterSpec struct {
	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster

	Size int `json:"size"`
	// Version is the expected version of the etcd container image.
	Version string `json:"version"`
	// StorageSpec is the name of the StorageSpec to use for the etcd cluster. If not provided, then each POD just uses the temporary storage inside the container.
	StorageSpec *StorageSpec `json:"storageSpec,omitempty"`
	// TLS is the TLS certificate configuration to use for the etcd cluster and etcd operator.
	TLS *TLSCertificate `json:"tls,omitempty"`
	// etcd configuration options are passed as command line arguments to the etcd container, refer to etcd documentation for configuration options applicable for the version of etcd being used.
	EtcdOptions []string `json:"etcdOptions,omitempty"`
}
```

If that is the case could I propose reordering all the fields so they are in alphabetical order?

internal/controller/utils.go
SDBrett: Yeah, that's a lot cleaner lol thanks

internal/controller/utils.go
SDBrett: Does this suggestion conflict with this one? https://github.com/etcd-io/etcd-operator/pull/93#discussion_r1997050394

internal/controller/utils_test.go
SDBrett: It seems my e2e test issues are due to a limitation with Kind when using podman.

api/v1alpha1/etcdcluster_types.go
ahrtr: > Just to confirm, you would like the struct to look like the below?

YES



> If that is the case could I propose reordering all the fields so they are in alphabetical order?

Suggest not to reorder them in alphabetical order. We intentionally put them like this way.

ahrtr:
Overall looks good to me. Thanks for the contribution. Please squash the commits, also raise a ticket to track the effort of adding an e2e test.

cc @hakman @ivanvc @jmhbnz

jmhbnz (approved):
The approach looks good to me, thanks for the contribution @SDBrett 

Edit: As mentioned please squash your commits so we can accept the contribution, thanks!

SDBrett:
@ahrtr @jmhbnz I've squashed commits and created an issue for e2e tests

ahrtr (approved):
Thanks for your first and awesome contribution.

PR #100, opened 2025-03-14
Add release script

ivanvc:
The release script does the following:

1. Generates the distribution YAML file and places it in the default location `dist/install.yaml`. Then, it commits the file to the repository.
2. Creates a tag and pushes it to the repository.
3. Builds the Docker image and pushes to the registry.

These steps are idempotent, so if the script fails, it can be resumed and skip what already succeeded.

I tested this in my fork and pushed to my registry.

1. It generates the distribution YAML, this is the commit: https://github.com/ivanvc/etcd-operator/commit/7d3d4b8d1e6831bcdfbd9d612bbffc8c8f0a7c40
2. It pushed the tag: https://github.com/ivanvc/etcd-operator/releases/tag/v0.1.0 (pointing to that commit)
3. This is the image it pushed: https://hub.docker.com/repository/docker/ivan/etcd-operator/tags/v0.1.0/sha256:b9d444f74381443d79cbdf7e449db14c8be8b46ff19b96378d74132dbeaf1b47

Question: Should we create a release based on the tag (like in the etcd repository, with the installation instructions)? I think it makes sense.

/cc @ahrtr

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+
+set -e
+
+REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd-operator.git}"
+REGISTRY="${REGISTRY:-gcr.io/etcd-development/etcd-operator}"
+CONTAINER_TOOL="${CONTAINER_TOOL:-docker}"
+
+function main {
+  if [ -z "$1" ]; then
+    cat <<EOF
+Usage: $0 [release_version]
+
+  Arguments:
+    release_version can be specified with our without the "v" prefix (i.e., v0.1.0 or 0.1.0)
+EOF
+    exit
+  fi
+
+  local version="v${1#v}"
+  local image="${REGISTRY}:${version}"
+
+  build_and_push_image "${image}"
+  build_and_commit_dist_files "${version}" "${image}"
+
+  if ! tag_repository "$version"; then
+    echo "Failed to tag repository"
+    exit 1
+  fi
+}
+
+function build_and_push_image {
+  if "${CONTAINER_TOOL}" manifest inspect "${image}" &>/dev/null; then
+    echo "Skipping pushing container image. Image ${image} already exists."
+    return
+  fi
+  echo "Building and pushing container image ${image}"
+  make docker-buildx IMG="${image}"
+}
+
+function build_and_commit_dist_files {
+  local version="$1"
+  local image="$2"
+
+  echo "Building distribution YAML and committing to repository"
+  make build-installer IMG="${image}" VERSION_SUFFIX="-${version}"
+  git checkout -- config/manager/kustomization.yaml
+  git add "dist/install-${version}.yaml"
+  if git diff --cached --exit-code; then
+    echo "Skipping commiting distribution files. No changes to commit."
+    return
+  fi
+  git commit --signoff --message "Release ${version} distribution files"
+  git push -f "${REPOSITORY}"
+}
+
+function get_gpg_key {
+  local git_email
+  local key_id
+
+  git_email=$(git config --get user.email)
+  key_id=$(gpg --list-keys --with-colons "${git_email}" | \
+    awk -F: '/^pub:/ { print $5 }')
+  if [[ -z "${key_id}" ]]; then
+    echo "Failed to load gpg key. Is gpg set up correctly for etcd releases?"
+    return 2
+  fi
+  echo "${key_id}"
+}
+
+function tag_repository {
+  local version="$1"
+
+  if [ "$(git tag --list | grep -c "${version}")" -gt 0 ]; then
+    echo "Skipping tag step. git tag ${RELEASE_VERSION} already exists."
+  else
+    echo "Tagging release..."
+    local key_id
+    key_id=$(get_gpg_key) || return 2
+    git tag --local-user "${key_id}" --sign "${version}" --message "${version}"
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,80 @@
...
+  local image="${REGISTRY}:${version}"
+
+  build_and_commit_dist_files "${image}"
+
+  if ! tag_repository "$version"; then
+    echo "Failed to tag repository"
+    exit 1
+  fi
+
+  build_and_push_image "${image}"
```
ahrtr, line 25: Minor comment: I think we should build & push images firstly, and generate the `dist/install.yaml` later because it refers to (or depends on) the image.

scripts/release.sh
```diff
@@ -0,0 +1,80 @@
...
+  make docker-buildx IMG="${image}"
+}
+
+function build_and_commit_dist_files {
+  local image="$1"
+
+  echo "Building distribution YAML and committing to repository"
+  make build-installer IMG="${image}"
+  git checkout -- config/manager/kustomization.yaml
+  git add dist/install.yaml
```
ahrtr, line 43: Previously dist isn't managed by git. If we want to manage it in git, can we add the version into the file name, i.e. install_v0.1.0.yaml?

scripts/release.sh
```diff
@@ -0,0 +1,80 @@
...
+
+set -e
+
+REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd-operator.git}"
+REGISTRY="${REGISTRY:-gcr.io/etcd-development/etcd-operator}"
+CONTAINER_TOOL="${CONTAINER_TOOL:-docker}"
+
+function main {
+  if [ -z "$1" ]; then
+    echo "Usage: $0 <release version>"
```
ahrtr, line 11: can we add examples to clarify that both `release.sh v0.1.0` and `release.sh 0.1.0` work?

scripts/release.sh
ivanvc: That sounds good. I moved the order many times and was never convinced. Your point makes sense. I'll update it.

ivanvc: Done.

scripts/release.sh
ivanvc: Done.

scripts/release.sh
ivanvc: Done. I named it with a hyphen for consistency with branch names, and other places where we add the version to the binary (i.e., install-v0.1.0.yaml).

ivanvc:
I ran the script with the changes doing a fictional 0.2.0 version release at my fork:

* dist YAML file: https://github.com/ivanvc/etcd-operator/tree/release-0.2/dist
* tag: https://github.com/ivanvc/etcd-operator/releases/tag/v0.2.0
* docker image: https://hub.docker.com/layers/ivan/etcd-operator/v0.2.0/images/sha256-71e3dd82d20cd26f9840967ba4a27b1042b7c07b457cf35341f6a64bcd850066

ahrtr:
> Question: Should we create a release based on the tag (like in the etcd repository, with the installation instructions)?

I'm not sure that I got the question. Do you mean to create a release branch something like `release-v0.1` for each minor release? If yes, I don't think we need to do it for now, at least before we release v1.0.

ahrtr (approved):
LGTM

Thanks

cc @jmhbnz @hakman

ahrtr:
cc @ArkaSaha30

ahrtr:
Also probably we should also add `--dry-run` similar to etcd, but of course can be discussed & addressed separately.

scripts/release.sh
```diff
@@ -0,0 +1,85 @@
...
+
+function tag_repository {
+  local version="$1"
+
+  if [ "$(git tag --list | grep -c "${version}")" -gt 0 ]; then
+    echo "Skipping tag step. git tag ${RELEASE_VERSION} already exists."
+  else
+    echo "Tagging release..."
+    local key_id
+    key_id=$(get_gpg_key) || return 2
```
jmhbnz, line 79: Not suggesting we address it in this pull request, but in future it would be great if this could support signing via ssh public key.  Signing commits via ssh public key is quite common, ref: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key

scripts/release.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+
+set -e
+
+REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd-operator.git}"
+REGISTRY="${REGISTRY:-gcr.io/etcd-development/etcd-operator}"
+CONTAINER_TOOL="${CONTAINER_TOOL:-docker}"
```
jmhbnz, line 7: Out of curiosity have you tested with `podman`?

jmhbnz (approved):
Overall this is looking great, thanks for the work on it @ivanvc

scripts/release.sh
hakman: Would [ko-build/ko](https://github.com/ko-build/ko) also be a good option?

ahrtr: we can continue to enhance the release script in future if needed.

ivanvc: Podman doesn't work because of the built-in cross architecture build from the scaffolded `Makefile`. I agree; we can iterate and enhance it.

PR #101, opened 2025-03-17
Change `ProviderCfg` into a non-pointer field

ahrtr:
As long as [EtcdClusterSpec.TLS](https://github.com/etcd-io/etcd-operator/blob/2c9366ae2f23f02b10ceb24072a5b3ffe5c6f0e2/api/v1alpha1/etcdcluster_types.go#L40) isn't nil, then one of the certificate providers (either `AutoCfg` or `CertManagerCfg` for now) must be provided.

Previously users need to check both `EtcdClusterSpec.TLS` and `EtcdClusterSpec.TLS.ProviderCfg`, now users only need to check whether `EtcdClusterSpec.TLS` is nil or not.

cc @ArkaSaha30 @ivanvc  @jmhbnz @hakman


ivanvc:
This looks good. I checked, though, and in #9, there's no such check. Should we have tested `TLSConfig` against `nil`, too?

ahrtr:
> I checked, though, and in #9, there's no such check. Should we have tested `TLSConfig` against `nil`, too?

The PR #9 already covered the `nil` check on TLS. Anyway, I think it can be discussed and addressed separately.

ivanvc:
> The PR #9 already covered the `nil` check on TLS. Anyway, I think it can be discussed and addressed separately.

Sounds good. Let's get this merged, then.

Issue #10, opened 2024-12-09
Certificate management & usage


ahrtr:
Refer to the [Spec & design](https://docs.google.com/document/d/1k8os77Hqxe36nUaZ31YqBGBuXcgdozVPppJYOfwUDFA/edit?tab=t.0).

## Tasks breakdown
- [ ] Mount secrets into etcd PODs
       From etcd POD perspective, it doesn't care about how the secrets/certificates are generated, it just mounts the secret and use the certificates directly.
    - [ ] Extend the CRD definition to include certificate, similar to https://github.com/etcd-io/etcd-operator/pull/9#discussion_r1873917550
    - [ ] Update reconciliation workflow to mount the secrets into etcd POD. Blocked by the reconciliation work.
    - [ ] Users are supposed to get the secretes/certificates prepared before creating etcd clusters. We need to clearly document this.
        - [ ] Users can manually create & manage the secrets & certificates themselves. 
        - [ ] etcd-operator can try to implement some utilities to manage secrets & certificates (see next section)
- [ ] etcd-operator supports certificate management
    - [ ] Define an interface, so that users can extend or integrate their own certificate service with the etcd-operator.
        - Note etcd-operator's priority is to simplify & automate the management of the etcd clusters instead of certificate management. 
    - [ ] Implement some simple certificate management utilities for testing purpose, 
        - [ ]  i.e. supporting auto self-signed certificate.
    - [ ] Investigate some best practice & popular certificate issuers to handle certificate management, and decide the next step.

References:
- https://docs.google.com/document/d/1k8os77Hqxe36nUaZ31YqBGBuXcgdozVPppJYOfwUDFA/edit?tab=t.0
- https://github.com/etcd-io/etcd-operator/pull/59

gdasson:
@ahrtr : Can you please assign me? I'll start working on it from next week. Thanks.

ahrtr:
Thanks @gdasson . Previously @ArkaSaha30 was working on this task. But since there isn't much progress so far,  we don't have much time to wait, so assigned this task to both of you.

PR #9, opened 2024-12-06
Issue certifcates for etcd-operator

ArkaSaha30:
This PR will add a basic capability to issue self-signed certificates for etcd-operator.

- Prerequisite: cert-manager needs to be installed before deploying the etcd-operator
`kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.16.2/cert-manager.yaml`

- Once etcd-operator is deployed it requests 3 certificates from an issuer in the `etcd-operator-system` namespace
   - peer certificate
   - client certificate
   - server certificate

```
deployment.apps/etcd-operator-controller-manager created
certificate.cert-manager.io/etcd-operator-etcd-client-certificate created
certificate.cert-manager.io/etcd-operator-etcd-peer-certificate created
certificate.cert-manager.io/etcd-operator-etcd-server-certificate created
issuer.cert-manager.io/etcd-operator-selfsigned created

etcd-operator on  issue-cert via 🐳 colima via 🐹 v1.22.7 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ kubectl get secrets -n etcd-operator-system
NAME              TYPE                DATA   AGE
etcd-client-tls   kubernetes.io/tls   3      33s
etcd-peer-tls     kubernetes.io/tls   3      33s
etcd-server-tls   kubernetes.io/tls   3      34s

etcd-operator on  issue-cert via 🐳 colima via 🐹 v1.22.7 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ kubectl -n etcd-operator-system get secret etcd-client-tls -o yaml | grep ca.crt
[...]

PR #103, opened 2025-03-17
Add `ProviderCertManagerConfig` struct fields

ArkaSaha30:
This commit will add the necessary fields to `ProviderCertManagerConfig` which are required as a user-input to create a Certificate via cert-manager provider.

Note:
Decreasing the code change volume in https://github.com/etcd-io/etcd-operator/pull/9 for simplicity and readability.

Part of https://github.com/etcd-io/etcd-operator/issues/10


api/v1alpha1/etcdcluster_types.go
```diff
@@ -54,6 +56,42 @@ type ProviderAutoConfig struct {
...
+	IssuerKind string `json:"issuerKind,omitempty"`
+
+	// accepts a string as the name of Issuer
+	// if nil, programmatically creates an issuer with name:selfsigned
+	// +optional
+	IssuerName string `json:"issuerName,omitempty"`
+
+	// organizations to be used on the Certificate.
+	// +optional
+	Organization []string `json:"organizations,omitempty"`
```
ahrtr, line 94: We already discussed this previously, two comments,

1. Please keep the same or similar order as the `Config`, put the common fields first, then provider specific fields later.
https://github.com/etcd-io/etcd-operator/blob/1adf166c0efc0a7f593dc61d228f502ecda87772/pkg/certificate/interfaces/interface.go#L18-L27

2. We also discussed about reuse the existing Config, and define a list of keys of the supported parameters for each provider. Each approach has pros and cons. I am OK to keep it as it's for now, and enhance it in future.
    - reuse Config
        - pros: less duplication definition
        - cons: the `api/v1alpha1` will depend on `pkg/certificate/interface` or vice versa
    - do not reuse
        - pros: no any dependency between `api/v1alpha1` and `pkg/certificate/interface`
        - cons: duplication definition because `ProviderCertManagerConfig` is very similar to the `Config`
        -  Note: I think even we do not reuse the Config, we can still extract the common fields into a common struct.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -54,6 +56,42 @@ type ProviderAutoConfig struct {
 }
 
 type ProviderCertManagerConfig struct {
+	// accepts a secret name with CABundle present
+	// this secret will be used to create an Issuer
+	// +optional
+	CABundleSecret string `json:"caBundleSecret,omitempty"`
```
ahrtr, line 62: Please follow the best practice to add comment, this is a generic comment for all fields

https://go.dev/wiki/CodeReviewComments#comment-sentences

Also based on our discussion, the CABundleSecret isn't for Issuer, it's for etcd instance in case the secret generated by the provider (cert-manager) doesn't have CA.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -54,6 +56,42 @@ type ProviderAutoConfig struct {
...
+
+	// accepts a string array, if nil programmatically populates the
+	// CommonName as DNSNames
+	// +optional
+	DNSNames []string `json:"dnsNames,omitempty"`
+
+	// accepts a string which can be parsed with time.ParseDuration()
+	// if nil, defaults to 90 days
+	// +optional
+	Duration string `json:"duration,omitempty"`
```
ahrtr, line 75:
```suggestion
	ValidityDuration time.Duration `json:"validityDuration,omitempty"`
```

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: In case of cert-manager, it requires an Issuer to create a certificate secret. If it's a self-signed Issuer or there is an existing Issuer in the cluster(updated by user in `IssuerKind` and `IssuerName` ) then no `CABundleSecret` is required. 
But if there is no Issuer, this `CABundleSecret` is then used to create an Issuer which will in turn create the certificates. Ref: https://cert-manager.io/v1.6-docs/configuration/ca/#deployment

So, I think we can keep both the `CABundleSecret` and `IssuerKind, IssuerName` as mutually exclusive, like if one is already filled in, the other is not required. If both are filled, `CABundleSecret` will take the higher priority. WDTY?

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: > do not reuse
pros: no any dependency between api/v1alpha1 and pkg/certificate/interface
cons: duplication definition because ProviderCertManagerConfig is very similar to the Config
Note: I think even we do not reuse the Config, we can still extract the common fields into a common struct.

I am more inclined to not reuse, but agree on having a common struct with the common fields, something like:
```
type CMExtraConfig struct {
    IssuerKind string `json:"issuerKind,omitempty"`
    IssuerName string `json:"issuerName,omitempty"`
}

type AutoExtraConfig struct {
}

type ExtraConfig struct {
    CMExtraConfig   CMExtraConfig
    AutoExtraConfig AutoExtraConfig
}

// Config contains the basic fields required for creating a certificate
type ProviderConfig struct {
    CommonName       string
    Organization     []string
    AltNames         AltNames
    ValidityDuration time.Duration
    CABundle         string

    // ExtraConfig contains provider specific configurations.
    ExtraConfig ExtraConfig
}
```
WDYT?

api/v1alpha1/etcdcluster_types.go
ahrtr: > But if there is no Issuer, this `CABundleSecret` is then used to create an Issuer

As we discussed previously, It isn't etcd-operator's responsibility to create an Issuer, nor should etcd-operator makes any changes to users' existing certificate system.

api/v1alpha1/etcdcluster_types.go
ahrtr: > having a common struct with the common fields, something like:

It will change the users's interface we already discussed in https://docs.google.com/document/d/1k8os77Hqxe36nUaZ31YqBGBuXcgdozVPppJYOfwUDFA/edit?tab=t.0

From users perspective, they configure the certificate something like below,

```
apiVersion: operator.etcd.io/v1alpha1
kind: EtcdCluster
metadata:
  labels:
    app.kubernetes.io/name: etcd-operator
    app.kubernetes.io/managed-by: kustomize
  name: etcdcluster-sample
spec:
  size: 3
  version: "v3.5.17"
  tls:
    provider: “cert-manager”
    providerConfig:
       autoCfg *AutoConfig
       certManagerCfg *CertManagerConfig
      …..
```

From our design/implementation perspective, I think we can get the common fields included in a separate struct, i.e. `CommonConfig` , and get it nested in both `CertManagerConfig` and `AutoConfig`, for example,

type CertManagerConfig struct {
    CommonConfig

    // any provider specific fields added here
}

ArkaSaha30: Updated the PR with the above changes. 
Probably the [interface.go](https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/interfaces/interface.go#L18-L27) also needs to be updated accordingly.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +53,58 @@ type ProviderConfig struct {
...
+}
+
 type ProviderAutoConfig struct {
+	// CommonConfig accepts all the common fields required to create a certificate
+	CommonConfig CommonConfig `json:"commonConfig"`
 }
 
 type ProviderCertManagerConfig struct {
+	// CommonConfig accepts all the common fields required to create a certificate
+	CommonConfig CommonConfig `json:"commonConfig"`
```
ahrtr, line 97: I meant embeded field, pls see https://github.com/etcd-io/etcd-operator/pull/103#discussion_r2071291036
```suggestion
	CommonConfig
```

api/v1alpha1/etcdcluster_types.go
ahrtr: can you deliver a separate PR to implement the solution 1 (reuse Config) mentioned above, also as we discussed in previous WG meeting and 1:1 meeting, so that we can compare the two solutions?

api/v1alpha1/etcdcluster_types.go
ahrtr: can you deliver a separate PR to implement the solution 1 (reuse Config) mentioned above, also as we discussed in previous WG meeting and 1:1 meeting, so that we can compare the two solutions?

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: My bad, missed the embedding part. Fixed it now.

ahrtr:
Overall looks good now, please squash the commits. 

The only concern is the `CommonConfig` and `AltNames` are almost identical to the existing  definitions below, it's exactly the reason why I raised https://github.com/etcd-io/etcd-operator/pull/103#discussion_r2073036370. But it's OK to any refactoring in followup PR(s).

https://github.com/etcd-io/etcd-operator/blob/1adf166c0efc0a7f593dc61d228f502ecda87772/pkg/certificate/interfaces/interface.go#L12-L27

ahrtr:
The other concern is that the certificate rotation (renew) process/design isn't super clear for now, which means there is som uncertainty (risk) on the rework. Given we are still on v0.1.0, it should be fine.

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: Got it 👍

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: > can you deliver a separate PR to implement the solution 1 (reuse Config) mentioned above, also as we discussed in previous WG meeting and 1:1 meeting, so that we can compare the two solutions?

Sure, as discussed in our 1:1, I will get the implementation of certificate management in by keeping the etcdclusterAPI and certManager/interface separate and follow up with the reuse Config approach PR.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +53,55 @@ type ProviderConfig struct {
...
+	// to the x509 certificate SubAltNames fields. The values will be passed
+	// directly to the x509.Certificate object.
+	AltNames AltNames `json:"altNames,omitempty"`
+
+	// ValidityDuration accepts time.Duration, if nil defaults to 90 days
+	// +optional
+	ValidityDuration time.Duration `json:"validityDuration,omitempty"`
+
+	// CABundleSecret accepts a secret name with CABundle present
+	// if nil, will consider IssuerKind, IssuerName for certificate creation
```
ahrtr, line 86: In golang, a string can never be a `nil`
```suggestion
	// if empty, will consider IssuerKind, IssuerName for certificate creation
```

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +53,58 @@ type ProviderConfig struct {
...
+	DNSNames []string `json:"dnsNames,omitempty"`
+
+	// IPs accepts a net.IP array of IP Addresses as alternative names
+	// +optional
+	IPs []net.IP `json:"ipAddresses,omitempty"`
+}
+
+type CommonConfig struct {
+	// CommonName accepts a string for a human-readable name of the certificate
+	CommonName string `json:"commonName"`
```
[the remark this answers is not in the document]
ArkaSaha30: Sure, updated to make it optional.

api/v1alpha1/etcdcluster_types.go
ahrtr: It seems that comment isn't valid based on https://github.com/etcd-io/etcd-operator/pull/103#discussion_r2070201524

ArkaSaha30: Yes, re-writing the comment based on the discussion.

pkg/certificate/interfaces/interface.go
```diff
@@ -20,7 +20,7 @@ type Config struct {
 	Organization     []string
 	AltNames         AltNames
 	ValidityDuration time.Duration
-	CABundle         []byte
+	CABundleSecret   string
```
ivanvc, line 23: Quick question, why did you change this? Is this because of https://github.com/etcd-io/etcd-operator/pull/103/files#diff-1d55d8dd7e081f3e3c7c5ab245f407c05a1b8b49903385124be229947e77a555R85-R89?

(I may be lacking some context)

ArkaSaha30: Yes, that is correct. CABundle will be stored in a secret which will be passed by the user. Also, this will close Issue https://github.com/etcd-io/etcd-operator/issues/104.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +53,56 @@ type ProviderConfig struct {
 	CertManagerCfg *ProviderCertManagerConfig `json:"certManagerCfg,omitempty"`
 }
 
+type AltNames struct {
+	// DNSNames accepts a string array, if empty programmatically populates the
+	// CommonName as DNSNames
```
ahrtr, line 58: - I think it makes more sense to by default populate DNSNames as string something like `*.example.default.svc.cluster.local`, see https://github.com/etcd-io/etcd-operator/blob/1adf166c0efc0a7f593dc61d228f502ecda87772/internal/controller/utils.go#L86
- nit: Please ensure the comments' style is consistent as the existing ones (no need to to add "accepts a string" as the variable type already self-explains it), this comment applies to all fields, https://github.com/etcd-io/etcd-operator/blob/1adf166c0efc0a7f593dc61d228f502ecda87772/api/v1alpha1/etcdcluster_types.go#L33-L42
- All the fields in `CommonConfig` and `AltNames` are very common in certificate utilities/packages, please refer to how others (i.e. golang std lib) add comments for them.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +53,56 @@ type ProviderConfig struct {
...
+	Organization []string `json:"organizations,omitempty"`
+
+	// AltNames contains the domain names and IP addresses that will be added
+	// to the x509 certificate SubAltNames fields. The values will be passed
+	// directly to the x509.Certificate object.
+	AltNames AltNames `json:"altNames,omitempty"`
+
+	// ValidityDuration accepts time.Duration, if empty defaults to 90 days
+	// +optional
+	ValidityDuration time.Duration `json:"validityDuration,omitempty"`
```
ahrtr, line 83: If you look at the generated yaml file `config/crd/bases/operator.etcd.io_etcdclusters.yaml`, the type is converted to `int64` (nanaseconds). It will be hard for users to configure it. We expect that users should be able configure human readable values something like "90d", "48h", "100d12h" etc

ArkaSaha30: In that case, it will be better to accept it as a string and parse it internally when setting the internal config.

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: Updated the PR 👍

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +52,58 @@ type ProviderConfig struct {
...
+	AltNames AltNames `json:"altNames,omitempty"`
+
+	// ValidityDuration is the expected duration until which the certificate will be valid,
+	// expects in human-readable duration: 100d12h, if empty defaults to 90d
+	// +optional
+	ValidityDuration string `json:"validityDuration,omitempty"`
+
+	// CABundleSecret is the expected secret name with CABundle present
+	// and validate if the Issuer provided has the same secretName
+	// if empty, directly use IssuerKind, IssuerName for certificate creation
```
ahrtr, line 88: As we discussed, the `CABundleSecret` has nothing to do with the users' certificate management system (i.e. cert-manager), so it has nothing to do with IssuerKind and IssuerName here. It's used by etcd instance to verify the TLS communication in case a CA isn't present in the secret created by cert-manager.

ArkaSaha30: Yes, updated the description.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -50,10 +52,56 @@ type ProviderConfig struct {
...
+	// to the x509 certificate SubAltNames fields. The values will be passed
+	// directly to the x509.Certificate object.
+	AltNames AltNames `json:"altNames,omitempty"`
+
+	// ValidityDuration is the expected duration until which the certificate will be valid,
+	// expects in human-readable duration: 100d12h, if empty defaults to 90d
+	// +optional
+	ValidityDuration string `json:"validityDuration,omitempty"`
+
+	// CABundleSecret is the expected secret name with CABundle present
```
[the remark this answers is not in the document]
ahrtr: Actually etcd supports set different CA for peer communication and client communication. We are using the same for both of them for now.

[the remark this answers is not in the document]
ArkaSaha30: Thank you 🙌

ahrtr (approved):
LGTM

cc @ivanvc

pkg/certificate/interfaces/interface.go
ivanvc: Thanks for the explanation :)

PR #109, opened 2025-03-22
Defaults to ReadWriteOnce when it's empty or not provided

ahrtr:
When `ec.Spec.StorageSpec.AccessModes` is empty, or not configured at all, we should defaults to `corev1.ReadWriteOnce`. The current implementation is just to set an empty string.

Still we should add test for the storage.

cc @ivanvc @gdasson


ivanvc (approved):
LGTM. Thanks, @ahrtr

PR #116, opened 2025-03-27
Add cluster size validation marker

frederiko:
This PR adds:
* minimum cluster size of 0 members
* generate CRD
* add defaults to config/samples  manifest

### DIFF
--- a/api/v1alpha1/etcdcluster_types.go
+++ b/api/v1alpha1/etcdcluster_types.go
@@ -33,6 +33,7 @@ type EtcdClusterSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
 	// Size is the expected size of the etcd cluster.
+	// +kubebuilder:validation:Minimum=0
 	Size int `json:"size"`
 	// Version is the expected version of the etcd container image.
 	Version string `json:"version"`
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: controller
+  newTag: latest
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: etcd-operator
+  newTag: v0.1
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -83,8 +83,6 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
-	// TODO: Implement finalizer logic here
-
 	logger.Info("Reconciling EtcdCluster", "spec", etcdCluster.Spec)
 
 	// Get the statefulsets which has the same name as the EtcdCluster resource
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"context"
+	"os"
 	"testing"
 	"time"
 
@@ -34,7 +35,10 @@ import (
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
+var etcdVersion = os.Getenv("ETCD_VERSION")
+
 // TestZeroMemberCluster tests if zero member Etcd Cluster does not create its StatefulSet
+// TODO: update this test once https://github.com/etcd-io/etcd-operator/issues/125 is addressed
 func TestZeroMemberCluster(t *testing.T) {
 	feature := features.New("zero-member-cluster")
 	etcdClusterName := "etcd-cluster-zero"
@@ -51,7 +55,7 @@ func TestZeroMemberCluster(t *testing.T) {
 		},
 		Spec: ecv1alpha1.EtcdClusterSpec{
 			Size:    size,
-			Version: "v3.5.18",
+			Version: etcdVersion,
 		},
 	}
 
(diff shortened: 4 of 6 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -78,13 +78,6 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
-	if etcdCluster.Spec.Size == 0 {
-		logger.Info("EtcdCluster size is 0..Skipping next steps")
-		return ctrl.Result{}, nil
-	}
```
ahrtr, line 84: Suggest to keep this no matter we add the restriction (minimum value of cluster size), the controller should be correct by itself. It's also harmless

api/v1alpha1/etcdcluster_types.go
```diff
@@ -31,6 +31,7 @@ type EtcdClusterSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
 	// Size is the expected size of the etcd cluster.
+	// +kubebuilder:validation:Minimum=1
```
ahrtr, line 34:
```suggestion
	// +kubebuilder:validation:Minimum=0
```

jberkus:
So, what happens right now if you apply a cluster of size 0 is that we have a cluster object with no pods.  Which seems desireable; I can think of a number of circumstances where you might (temporarily) want a cluster with no members.

So, +1 on @ahrtr's proposal

frederiko:
@jberkus @ahrtr Not a problem at all. I just happened to have a different view from user's perspective. If I create a cluster with 0 members and I have no information about what happened, but have the resource persisted, I wouldn't like to have to tap into controller log for extra information. I couldn't also see a use case where an user may need a cluster with members.  I  will make the adjustments.

ivanvc:
> I wouldn't like to have to tap into controller log for extra information.

In my opinion, we should start adding events to the EtcdCluster so the users can understand the current state of the CR.

frederiko:
> If that is not desirable, other mechanisms would entail to write an event or update the `status`. However, more logic would be required.

"**If that is not desirable, other mechanisms would entail to write an event or update the status. However, more logic would be required.**". I wouldn't oppose to that. :-)

frederiko:
> > "**If that is not desirable, other mechanisms would entail to write an event or update the status. However, more logic would be required.**".
> 
> We will need to update/make use of the status regardless. What's the behaviour of some existing controller when the replica is 0, i.e. StatefulSet?

tbh, I don't recall a controller that presents this behavior. I would have to check.

ahrtr:
@frederiko this PR is basing on a very old commit, please rebase this PR. thx

frederiko:
> > tbh, I don't recall a controller that presents this behavior. I would have to check.
> 
> I mean the Kubernetes built-in controller, i.e. StatefulSet. As mentioned above [#116 (comment)](https://github.com/etcd-io/etcd-operator/pull/116#issuecomment-2757328807), the minimum replica is 0. Not sure whether the status will have some message or warning to indicate the 0 replica. We will need to do some experiments.

Yes, sts supports that. I meant more in the context of non-k8s controllers.

internal/controller/etcdcluster_controller.go
```diff
@@ -83,8 +83,6 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
-	// TODO: Implement finalizer logic here
-
```
ahrtr, line 87: It seems unrelated change?

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
-- manager.yaml
\ No newline at end of file
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: controller
+  newTag: latest
```
ahrtr, line 8: It seems unrelated change?

ahrtr (commented):
Overall looks good with two minor comments, thx.

Also it would be better if we have an e2e test to verify the minimum size.

ahrtr:
@frederiko have you manually verified this PR and confirmed a negative value (for the size) will raise an error? If yes, I think it's OK to merge it and add e2e test in followup PR; but it's also OK if you want to add the test in this PR. Please let me know your thought.

frederiko:
> @frederiko have you manually verified this PR and confirmed a negative value (for the size) will raise an error? If yes, I think it's OK to merge it and add e2e test in followup PR; but it's also OK if you want to add the test in this PR. Please let me know your thought.

I have just added a basic test to ensure that, when an Etcdcluster is created with zero members, no statefulset is created. Let me know if you would like to something else.

test/e2e/e2e_test.go
```diff
@@ -28,8 +28,65 @@ import (
...
+			APIVersion: "operator.etcd.io/v1alpha1",
+			Kind:       "EtcdCluster",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdClusterName,
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    size,
+			Version: "v3.5.18",
```
ArkaSaha30, line 54: nit: Can we use the `ETCD_VERSION` variable here? 
Ref: https://github.com/etcd-io/etcd-operator/blob/main/Makefile#L77

test/e2e/e2e_test.go
```diff
@@ -28,8 +28,65 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
+// TestZeroMemberCluster tests if zero member Etcd Cluster does not create its StatefulSet
+func TestZeroMemberCluster(t *testing.T) {
+	feature := features.New("zero-member-cluster")
+	etcdClusterName := "etcd-cluster-zero"
+	size := 0
```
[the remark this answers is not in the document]
frederiko: @ahrtr yes, I am aware of the bug. What I tried to address in this PR was the 0 size behavior (aware of the bug). I am going to work on that bug later, on a separate PR, which should make this test to go away (remember my first was actually removing that block of code?). 

As for the -1 size, do we really need to test it, since the api server will not persist the etcdcluster resource altogether? I can certainly add if this is a concern.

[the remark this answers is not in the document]
frederiko: So, I have added a test for negative cluster size as well. I have kept the 0 size cluster one, and I will address the bug later, if that's fine with you. ltmk your thoughts.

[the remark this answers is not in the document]
ahrtr: OK, thx

test/e2e/e2e_test.go
```diff
@@ -28,8 +29,114 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 )
 
+var etcdVersion = os.Getenv("ETCD_VERSION")
+
+// TestZeroMemberCluster tests if zero member Etcd Cluster does not create its StatefulSet
```
ahrtr, line 40: Could you add a comment something like: we will update this test once https://github.com/etcd-io/etcd-operator/issues/125 is resolved.

ahrtr (approved):
LGTM with a minor comment

Thanks

jmhbnz (approved):
LGTM - Thanks for your work on this @frederiko

Issue #106, opened 2025-03-19
Add e2e tests for etcdOptions


SDBrett:
Linked to PR #93 

Add e2e tests for etcdOptions

I'm happy to work on this as I've finally got e2e tests working locally :)

jmhbnz:
/assign @SDBrett 

Thanks for volunteering to pick this up 👍🏻

SDBrett:
I've made some progress on this can created an e2e test that's functional on my local system but I am finding it difficult to understand how to write properly use the Kubernetes e2e frame work.

I'm looking for some help to better understand how to use the framework.

I've pushed my work to a branch on my repo https://github.com/SDBrett/etcd-operator/blob/e2e-test/test/e2e/flags_test.go

Currently the test is independent of the rest of existing e2e tests, this is just for my own benefit (like having a persistent cluster instead of kind) and not how I would intend for it to be added to the project.

PR #120, opened 2025-04-02
etcd options e2e testing

SDBrett:
Links to issue #106 

Added e2e tests for etcdOptions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
+# The version of ETCD to run e2e tests against
+E2E_ETCD_VERSION ?= $(shell go list -m -f {{.Version}} go.etcd.io/etcd/api/v3)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -72,7 +74,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 # - CERT_MANAGER_INSTALL_SKIP=true
 .PHONY: test-e2e
 test-e2e: generate fmt vet kind ## Run the e2e tests. Expected an isolated environment using Kind.
-	PATH="$(LOCALBIN):$(PATH)" go test ./test/e2e/ -v
+	ETCD_VERSION="$(E2E_ETCD_VERSION)" PATH="$(LOCALBIN):$(PATH)" go test ./test/e2e/ -v
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter
--- a/test/e2e/etcd_options_test.go
+++ b/test/e2e/etcd_options_test.go
@@ -1,0 +1,170 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"os"
+	"reflect"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+)
+
+func TestEtcdOptions(t *testing.T) {
+	feature := features.New("etcd-options")
+
+	const etcdClusterName = "example"
+	const namespace = "etcd-test"
+	var expectedArgs = []string{
+		"--name=$(POD_NAME)",
+		"--listen-peer-urls=http://0.0.0.0:2380",
+		"--listen-client-urls=http://0.0.0.0:2379",
+		"--initial-advertise-peer-urls=http://$(POD_NAME).example.$(POD_NAMESPACE).svc.cluster.local:2380",
+		"--advertise-client-urls=http://$(POD_NAME).example.$(POD_NAMESPACE).svc.cluster.local:2379",
+		"--experimental-peer-skip-client-san-verification",
+	}
+
+	etcdCluster := &ecv1alpha1.EtcdCluster{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "operator.etcd.io/v1alpha1",
+			Kind:       "EtcdCluster",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdClusterName,
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    1,
+			Version: os.Getenv("ETCD_VERSION"),
+			EtcdOptions: []string{
+				"--experimental-peer-skip-client-san-verification",
+			},
+		},
+	}
+
+	stsObj := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdClusterName,
+			Namespace: namespace,
+		},
+	}
+
+	feature.Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+
+		client := cfg.Client()
+
+		ctx, err := envfuncs.CreateNamespace(etcdCluster.Namespace)(ctx, cfg)
+		if err != nil {
+			t.Fatalf("failed to create namespace: %s", err)
+		}
+
+		// create etcd cluster
+		if err := client.Resources().Create(ctx, etcdCluster); err != nil {
+			t.Fatalf("unable to create etcd cluster: %s", err)
+		}
+
+		// get etcd cluster object
+		var ec ecv1alpha1.EtcdCluster
+		ops := []wait.Option{
+			wait.WithTimeout(1 * time.Minute),
+			wait.WithInterval(5 * time.Second),
+		}
+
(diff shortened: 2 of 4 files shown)


ivanvc:
/ok-to-test

@SDBrett, thanks for the pull request. Without deeply reviewing the code, I have an initial comment: Can you update the license header in both files? It's missing in `utils.go`, and the one from `etcd_options_test.go` says 2021 and "The Kubernetes authors" (I'm unsure if we prefer "The etcd authors").

test/e2e/etcd_options_test.go
```diff
@@ -0,0 +1,164 @@
...
+			APIVersion: "operator.etcd.io/v1alpha1",
+			Kind:       "EtcdCluster",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdClusterName,
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    1,
+			Version: "v3.5.17",
```
jmhbnz, line 62: Should we bump this to `3.5.21` now that it has been released?  More broadly should we hardcode this or pull it from a binary version or git latest release etc?

test/e2e/utils.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err := client.Resources().Get(ctx, etcdCluster.Name, etcdCluster.Namespace, &ec); err != nil {
+		return ctx, fmt.Errorf("unable to fetch etcd cluster: %s", err)
+	}
+
+	return ctx, nil
+}
+
+// Gets K8S object from cluster, accepts wait options. Returns error if resource does not exist
+func getKubernetesObject(object k8s.Object, ctx context.Context, client klient.Client, etcdCluster *ecv1alpha1.EtcdCluster, outObj k8s.Object, options ...wait.Option) (context.Context, error) {
```
jmhbnz, line 53: I would lean towards naming this `getKubernetesResource` to align with the terminology used by k8s client go.

test/e2e/etcd_options_test.go
ivanvc: Would there be an easy way to do this without automation? Because this looks very similar to what we have in the website repository.
Should we write some automation that keeps the latest version synced?

jmhbnz: An API call to https://api.github.com/repos/etcd-io/etcd/releases/latest during test execution should suffice.

SDBrett: Does it make sense to always point to the latest since users can specify the version of etcd to use?

SDBrett: What about adding an option to the make file to specify the version of etcd to test against and that could be provided as input?

test/e2e/utils.go
SDBrett: Thanks, have updated

test/e2e/etcd_options_test.go
jmhbnz: This raises an interesting question I can't recall if we have discussed previously, someone fill me in if so.  What is our support matrix of etcd versions to etcd operator versions?  From memory operator is only focused on 3.5.x+?

I guess it comes down to the intent of the test, it might be one of the following three:
1.  _"Ensure that etcd operator is always able to pass options end to end through to the latest released supported version of etcd"_
1.  _"Ensure that etcd operator is always able to pass options end to end through to a specified released version of etcd"_
1.  _"Ensure that etcd operator is always able to pass options end to end through to all supported versions of etcd"_

I don't really like option two because what if we change something in a new release of etcd and the etcd-operator needs a corresponding update but we don't detect it because the e2e tests is somewhere set to only test against an older version? 

Option three also sounds excessive.

Option one makes sense to me as over time when operator e2e tests run we will build up history of test execution against different etcd versions over time.  It might be nice to override the version somewhere at high level i.e. `Makefile` or environment variable.

ivanvc: I like option 3, for completeness, but I agree it could be excessive. However, suppose we do it like you suggest, by overriding an environment or make variable. In that case, that means that we can have this completeness in CI without needing to run local tests against every supported version.

I'm also realizing we can read/parse the etcd version from the `go.mod`, without using any API (minimizing possible flakes due to network).

SDBrett:
> /ok-to-test
> 
> @SDBrett, thanks for the pull request. Without deeply reviewing the code, I have an initial comment: Can you update the license header in both files? It's missing in `utils.go`, and the one from `etcd_options_test.go` says 2021 and "The Kubernetes authors" (I'm unsure if we prefer "The etcd authors").

Sorry missed this earlier, have made the updates

abdurrehman107 (approved):

abdurrehman107:
lgtm. Thank you @SDBrett.

Makefile
```diff
@@ -2,6 +2,8 @@
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
+# The version of ETCD to run e2e tests against
+E2E_ETCD_VERSION ?= v3.5.21
```
ivanvc, line 6: We discussed this some weeks ago. But would it be better to default to the latest required rather than having this version hardcoded here? I'm afraid we may forget to update this version when we release new versions.

```suggestion
E2E_ETCD_VERSION ?= $(shell go list -m -f {{.Version}} go.etcd.io/etcd/api/v3)
```

test/e2e/etcd_options_test.go
```diff
@@ -0,0 +1,166 @@
+/*
+Copyright 2024.
```
ivanvc, line 2: nit: 2025

```suggestion
Copyright 2025.
```

test/e2e/utils.go
```diff
@@ -0,0 +1,97 @@
+/*
+Copyright 2024.
```
ivanvc, line 2: nit: 2025

```suggestion
Copyright 2025.
```

ivanvc (commented):
I'm just coming back to this pull request after being away for two weeks. I left a question and two minor nits. Other than that, this is looking great :v:

ivanvc (approved):
LGTM. Thanks, @SDBrett.

test/e2e/etcd_options_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    1,
+			Version: os.Getenv("ETCD_VERSION"),
+			StorageSpec: &ecv1alpha1.StorageSpec{
+				AccessModes:       corev1.ReadWriteOnce,
+				VolumeSizeRequest: resource.MustParse("4Mi"),
+				VolumeSizeLimit:   resource.MustParse("4Mi"),
+			},
```
ahrtr, line 69: This e2e test focuses on the etcd options, you don't have to necessarily configure the storageSpec. Suggest to remove it keep this test as simple as possible, but not a big problem.

test/e2e/etcd_options_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Namespace: namespace,
+		},
+	}
+
+	feature.Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+
+		client := cfg.Client()
+		_ = corev1.AddToScheme(client.Resources().GetScheme())
+		_ = metav1.AddMetaToScheme(client.Resources().GetScheme())
+		_ = ecv1alpha1.AddToScheme(client.Resources().GetScheme())
```
ahrtr, line 88: You already added them into the TestMain, why add them again?

test/e2e/utils.go
ahrtr: You marked this comment resolved, but I still see 2024 instead of 2025. Probably you forgot to push your local change?

test/e2e/utils.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func setupTestRun(ctx context.Context,
+	cfg *envconf.Config,
+	client klient.Client,
+	etcdCluster *ecv1alpha1.EtcdCluster) (
+	context.Context,
+	error) {
+
+	ctx, err := envfuncs.CreateNamespace(etcdCluster.Namespace)(ctx, cfg)
```
ahrtr, line 53: The namesapce was already created in TestMain.

Also I do not see much value to create a separate function `setupTestRun`. If you insist on adding such function, then it should be reused by other test cases as well.

test/e2e/utils.go
```diff
@@ -0,0 +1,97 @@
...
+		return ctx, fmt.Errorf("unable to fetch etcd cluster: %s", err)
+	}
+
+	return ctx, nil
+}
+
+// Gets K8S object from cluster, accepts wait options. Returns error if resource does not exist
+func getKubernetesResource(
+	object k8s.Object,
+	ctx context.Context,
```
ahrtr, line 77: move ctx to the first parameter.

test/e2e/utils.go
```diff
@@ -0,0 +1,97 @@
...
+	client klient.Client,
+	etcdCluster *ecv1alpha1.EtcdCluster,
```
ahrtr, line 79: You create this generic function and you only care about the namespace & name, why pass in an etcdCluster object?

test/e2e/utils.go
```diff
@@ -0,0 +1,97 @@
...
+	outObj k8s.Object, options ...wait.Option) (
+	context.Context,
```
ahrtr, line 81: why return the passed context?

ahrtr (changes requested):
.

test/e2e/utils.go
```diff
@@ -0,0 +1,63 @@
...
+			return container, nil
+		}
+	}
+	return container, fmt.Errorf("container named %s not found", name)
+}
+
+// Gets K8S object from cluster, accepts wait options. Returns error if resource does not exist
+func getKubernetesResource(
+	ctx context.Context,
+	object k8s.Object,
```
ahrtr, line 44: Pls use `NamespacedName` instead, 

https://github.com/kubernetes/kubernetes/blob/0d9dccfc6bc22f2370650cf5b8ad99cbba74ea23/staging/src/k8s.io/apimachinery/pkg/types/namespacedname.go#L27-L30

Alternative, use `ObjectKey`

https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/client/interfaces.go#L32

test/e2e/utils.go
```diff
@@ -0,0 +1,63 @@
...
+	client klient.Client,
+	outObj k8s.Object, options ...wait.Option) error {
+
```
ahrtr, line 47: nit
```suggestion
```

test/e2e/utils.go
ahrtr: I meant to change the parameter from `k8s.Object` to `NamespacedName` or `ObjectKey`. You can even move this generic function into utils/utils.go.

ahrtr: This PR is almost ready to be merged with this comment unresolved. Pls let me know if you want to resolve it in this PR or in a followup PR.

ahrtr: Let me do it in a followup

ahrtr (approved):
Thanks for the contribution!

Issue #121, opened 2025-04-02
E2E test environment does not support podman


SDBrett:
It is challenging to perform e2e tests when using podman.

The primary issue is that `kindCluster.LoadImage` does not work with podman, instead `kindCluster.LoadImageArchive` is required but there is no make step to create archives.

I'm happy to work on adding option to make life easier for podman users.

PR #123, opened 2025-04-02
add e2e podman config

SDBrett:
Links to issue #121 

Added e2e config to make e2e testing easier for podman users.


test/e2e/e2e_suite_test.go
```diff
@@ -198,8 +205,58 @@ func TestMain(m *testing.M) {
...
+	ctx, err = loadImageArchive(*kindCluster, ctx)
+	if err != nil {
+		return ctx, err
+	}
+	return ctx, nil
+}
+
+func createImageArchive() error {
+	// Save docker image archive
+	cmd := exec.Command("make", "docker-save", fmt.Sprintf("ARCHIVE_NAME=%s", archiveName))
```
ivanvc, line 242: I tried running this locally with Podman, but I'm getting the following cleanup error:

```
2025/04/02 09:31:33 Destroying cluster...
2025/04/02 09:31:40 deleting image archive
2025/04/02 09:31:40 failed to delete image archive: remove etcdOperator.tar: no such file or directory
FAIL    go.etcd.io/etcd-operator/test/e2e       181.442s
FAIL
make: *** [Makefile:75: test-e2e] Error 1
```

Should we write the archive under a [`T.TempDir`](https://pkg.go.dev/testing#T.TempDir) directory? SO we don't need to do a cleanup later?

SDBrett: I wasn't aware of T.TempDir, I will have a look into this

SDBrett: I had a look at using T.TempDir but I wasn't sure how to use it within the scope of TestMain as that imports `*testing.M`, instead I used `os.MkdirTemp` and it is performing the cleanup function well.

test/e2e/e2e_suite_test.go
```diff
@@ -203,3 +211,49 @@ func TestMain(m *testing.M) {
...
+	ctx, err = loadImageArchive(*kindCluster, ctx, fname)
+	if err != nil {
+		return ctx, err
+	}
+	return ctx, nil
+}
+
+func createImageArchive(archivePath string) error {
+	// Save docker image archive
+	cmd := exec.Command("make", "docker-save", fmt.Sprintf("ARCHIVE_NAME=%s", archivePath))
```
jmhbnz, line 246: I think we should be passing the docker-image as defined in the test config.

```suggestion
	cmd := exec.Command("make", "docker-save", fmt.Sprintf("ARCHIVE_NAME=%s", archivePath), fmt.Sprintf("IMG=%s", dockerImage)
```

jmhbnz (commented):
This is looking almost ready to merge @SDBrett nice work.  After a test run at my end I have just one suggestion below 🙏🏻

test/e2e/e2e_suite_test.go
SDBrett: I have an change planned for after this PR is merged that will make this redundant.

The change will set `dockerImage` to `os.Getenv("IMG")`. The idea being to reduce the need to modify `e2e_suite_test.go` to suit the local environment and then revert for before committing.

jmhbnz: Happy to address it as follow-up, no problem

test/e2e/e2e_suite_test.go
```diff
@@ -72,11 +75,16 @@ func TestMain(m *testing.M) {
...
+			if containerTool == "podman" {
+				if ctx, err := useImageArchive(kindCluster, ctx); err != nil {
+					log.Println(err)
+					return ctx, err
+				}
+			} else {
+				if err := kindCluster.LoadImage(ctx, dockerImage); err != nil {
+					log.Printf("Failed to load image into kind: %s", err)
+					return ctx, err
+				}
```
ahrtr, line 87: Thanks for the enhancement.

However, I think it makes more sense to let the lower level libraries/tools (e.g `kind`, `sigs.k8s.io/e2e-framework/support/kind`) to support it instead of etcd-operator.

cc @BenTheElder

SDBrett: Before making this PR I tried to find a solution from the kind / the e2e framework but was unable to find a method to make it work. 

There has been discussion on getting kind to work with Podman for several years, including some PRs which were never merged. For the most part the discussion comes back to "use load image archive instead of load image. 

One of the latest PRs was closed by @BenTheElder with the comment to use load image archive. Ref: https://github.com/kubernetes-sigs/kind/pull/3109#issuecomment-2512262145

Additionally it does not appear that the Kind libraries expose if podman, docker or another CRI is being used.

BenTheElder: https://github.com/kubernetes-sigs/kind/issues/2038

I spoke to some redhatters about this last week. The issue is open and "help wanted", pending someone to work out the details. We've lost a lot of time to breaking changes in podman and for developing Kubernetes/Kubernetes (our first and original priority) docker is required (we use buildx there).

BenTheElder: FTR when you run with podman it should also still log about it being considered experimental due to this instability.

use image archive is a stopgap until the above linked issue is resolved. Kind is basically maintained in our spare time and we haven't personally had time to sort this out and nobody has stepped up to flesh out what the better image load UX would look like.

we know we have people with multiple runtimes installed and docker-image command behavior has been stable so we'd prefer to get it right and introduce a new command that considers the behaviors and overrides up front to minimize that headache

SDBrett: On using `LoadImageToKindClusterWithName` I don't know if that's the right approach as there are two other options which feel more consistent.

The e2e framework two functions for both loading images and image archives (both functions have the exact same code)
- https://github.com/kubernetes-sigs/e2e-framework/blob/a30da6e642c295acac9f273c3b921987f8c3eca2/pkg/envfuncs/provider_funcs.go#L135
- https://github.com/kubernetes-sigs/e2e-framework/blob/a30da6e642c295acac9f273c3b921987f8c3eca2/pkg/envfuncs/provider_funcs.go#L158

Using `LoadImageToKindClusterWithName`  feels like repeating something that already exists and is in libraries we are already using.

---

Happy to add that comment, it makes sense

---

To be honest I'm not sure how to create an e2e test this feature. I could move my changes into `utils.go` and create a `utils_test.go` to run unit  tests but I imagine that I would need to mock the interaction with the kind cluster and that's something I haven't done before.

jmhbnz: From my perspective I don't think we should require e2e testing suite features to have their own e2e tests in order for a contribution to be accepted.  e2e tests should be related to the functionality of the etcd operator.

As mentioned above in my review https://github.com/etcd-io/etcd-operator/pull/123#pullrequestreview-2745089619 I've validated the feature in my environment which is based on podman.

Merging this will reduce friction for future contributors who are also using podman.

ahrtr: > On using `LoadImageToKindClusterWithName` I don't know if that's the right approach as there are two other options which feel more consistent.
> 
> The e2e framework two functions for both loading images and image archives (both functions have the exact same code)

My point is to get the details something like `if containerTool == podman" { ...} else { ...}` encapsulated in `LoadImageToKindClusterWithName` to make the e2e test clean & more readable. Also `createImageArchive` and `loadImageArchive` should be part of the utilities (util.go)

> To be honest I'm not sure how to create an e2e test this feature.

I was thinking to get podman installed and set `CONTAINER_TOOL` as `podman`, then run the existing e2e test. Not sure how difficult it's. Since our priority/goal is to test etcd-operator, not the underneath container runtime/tool, so I am not insist on adding such workflow test.



> As mentioned above in my review [#123 (review)](https://github.com/etcd-io/etcd-operator/pull/123#pullrequestreview-2745089619) I've validated the feature in my environment which is based on podman.

I don't think this is a good justification, although I am not insist on adding e2e test asa mentioned above. Adding e2e test can prevent regression.

SDBrett: @ahrtr Write regards to the following comment, can you please confirm if my understanding is correct?

> My point is to get the details something like if containerTool == podman" { ...} else { ...} encapsulated in LoadImageToKindClusterWithName to make the e2e test clean & more readable. Also createImageArchive and loadImageArchive should be part of the utilities (util.go)

You're suggestion is replace the `if containerTool == podman" { ...} else { ...}`  section of code with a wrapper function in the `test/utils.go` file that performs relevant tasks to determine the container tool used and the perform appropriate steps to?

For example replace `if containerTool == podman" { ...} else { ...}`  with `test_utils.LoadContainerImageToKindCluster`

The `LoadContainerImageToKindCluster` function would follow the current logic.

SDBrett: @ahrtr 
> I was thinking to get podman installed and set CONTAINER_TOOL as podman, then run the existing e2e test. Not sure how difficult it's. Since our priority/goal is to test etcd-operator, not the underneath container runtime/tool, so I am not insist on adding such workflow test.

I don't think it's reasonable / feasible to install podman to a maintainers system to test the tests. I could look at some unit tests to test the basic logic, however I'm not sure of the extent of the testing as some things would require mocking things like creating the image archive process etc.

test/e2e/e2e_suite_test.go
```diff
@@ -72,10 +73,12 @@ func TestMain(m *testing.M) {
 				return ctx, err
 			}
 
-			// Load docker image into kind
-			log.Println("Loading docker image into kind cluster...")
-			if err := kindCluster.LoadImage(ctx, dockerImage); err != nil {
-				log.Printf("Failed to load image into kind: %s", err)
+			// LoadImage does not work when using podman, so LoadImageArchive is used instead
+			// refer to https://github.com/kubernetes-sigs/kind/issues/2038
```
ahrtr, line 77: Pls move this comment into `LoadContainerImageToKindCluster` (right following the` if containerTool == "podman"`)

test/utils/utils.go
```diff
@@ -247,3 +256,67 @@ func UncommentCode(filename, target, prefix string) error {
...
+		if err := kindCluster.LoadImage(ctx, dockerImage); err != nil {
+			log.Printf("Failed to load image into kind: %s", err)
+			return ctx, err
+		}
+	}
+	return ctx, nil
+}
+
+func useImageArchive(kindCluster *kind.Cluster, ctx context.Context) (context.Context, error) {
+	var err error
```
ahrtr, line 279:
```suggestion
```

test/utils/utils.go
```diff
@@ -247,3 +256,67 @@ func UncommentCode(filename, target, prefix string) error {
...
+	fname := filepath.Join(dname, archiveName)
+
+	defer func() {
+		if err := os.RemoveAll(fname); err != nil {
+			log.Printf("failed to remove %s: %s", fname, err)
+		}
+	}()
+
+	log.Println("Creating docker image archive...")
+	if err := createImageArchive(fname); err != nil {
```
ahrtr, line 294:
```suggestion
	if err = createImageArchive(fname); err != nil {
```

test/utils/utils.go
```diff
@@ -247,3 +256,67 @@ func UncommentCode(filename, target, prefix string) error {
...
+	return nil
+}
+
+func loadImageArchive(kindCluster kind.Cluster, ctx context.Context, archivePath string) (context.Context, error) {
+	// Load archive into cluster
+	if err := kindCluster.LoadImageArchive(ctx, archivePath); err != nil {
+		return ctx, err
+	}
+	return ctx, nil
+}
```
ahrtr, line 322: Minor comment: These two functions do not make much sense. I would suggest to move the implementation into `useImageArchive` directly.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/brejohns0/etcd-operator/etcd-operator
+  newTag: v0.1
```
ahrtr, line 8: can you explain why change it in this PR?

SDBrett: It was a mistake, will be removed

test/utils/utils.go
```diff
@@ -247,3 +256,68 @@ func UncommentCode(filename, target, prefix string) error {
 	// nolint:gosec
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
+
+// LoadImage does not work when using podman, so LoadImageArchive is used instead
+// refer to https://github.com/kubernetes-sigs/kind/issues/2038
+func LoadContainerImageToKindCluster(kindCluster *kind.Cluster,
+	ctx context.Context,
```
ahrtr, line 263: nit: usually context is the first parameter

test/utils/utils.go
```diff
@@ -247,3 +256,51 @@ func UncommentCode(filename, target, prefix string) error {
 	// nolint:gosec
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
+
+// LoadImage does not work when using podman, so LoadImageArchive is used instead
+// refer to https://github.com/kubernetes-sigs/kind/issues/2038
+func LoadContainerImageToKindCluster(ctx context.Context,
+	kindCluster *kind.Cluster,
+	dockerImage string,
+	containerTool string) (context.Context, error) {
```
ahrtr, line 265:
```suggestion
	containerTool string) error {
```

test/utils/utils.go
```diff
@@ -247,3 +256,51 @@ func UncommentCode(filename, target, prefix string) error {
...
+	} else {
+		if err := kindCluster.LoadImage(ctx, dockerImage); err != nil {
+			log.Printf("Failed to load image into kind: %s", err)
+			return ctx, err
+		}
+	}
+	return ctx, nil
+}
+
+func useImageArchive(ctx context.Context, kindCluster *kind.Cluster) (context.Context, error) {
```
ahrtr, line 280:
```suggestion
func useImageArchive(ctx context.Context, kindCluster *kind.Cluster) error {
```

test/utils/utils.go
ahrtr: It's a little weird to return the received context.

test/utils/utils.go
```diff
@@ -247,3 +256,51 @@ func UncommentCode(filename, target, prefix string) error {
 	// nolint:gosec
 	return os.WriteFile(filename, out.Bytes(), 0644)
 }
+
+// LoadImage does not work when using podman, so LoadImageArchive is used instead
+// refer to https://github.com/kubernetes-sigs/kind/issues/2038
+func LoadContainerImageToKindCluster(ctx context.Context,
+	kindCluster *kind.Cluster,
+	dockerImage string,
```
ahrtr, line 264:
```suggestion
	imageName string,
```

ahrtr: Also please remove the existing LoadImageToKindClusterWithName if it isn't used  
https://github.com/etcd-io/etcd-operator/blob/16920cb54a6ecf4d4a429b833d97f23765c5f24d/test/utils/utils.go#L176

test/utils/utils.go
SDBrett: True

test/utils/utils.go
```diff
@@ -247,3 +244,51 @@ func UncommentCode(filename, target, prefix string) error {
...
+	containerTool string) (context.Context, error) {
```
ahrtr, line 253: ditto https://github.com/etcd-io/etcd-operator/pull/123#discussion_r2036963870
```suggestion
	containerTool string) error {
```

test/utils/utils.go
```diff
@@ -247,3 +244,51 @@ func UncommentCode(filename, target, prefix string) error {
...
+		if err := kindCluster.LoadImage(ctx, dockerImage); err != nil {
+			log.Printf("Failed to load image into kind: %s", err)
+			return err
+		}
+	}
+	return nil
+}
+
+func useImageArchive(ctx context.Context, kindCluster *kind.Cluster) error {
+
```
ahrtr, line 269: nit
```suggestion
```

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,2 @@
 resources:
-- manager.yaml
+- manager.yaml
```
ahrtr, line 2: nit: You removed the newline character at the end of the file.

ahrtr (approved):
LGTM

It's much better now. Thanks for your contribution!

PR #139, opened 2025-05-09
Update GetKubernetesResource's signiture and move it into utils package

ahrtr:
Followup to https://github.com/etcd-io/etcd-operator/pull/120#discussion_r2061531928

cc @ivanvc @jmhbnz @abdurrehman107 @hakman @ArkaSaha30


test/utils/utils.go
```diff
@@ -291,3 +297,32 @@ func useImageArchive(ctx context.Context, kindCluster *kind.Cluster) error {
 	}
 	return nil
 }
+
+// GetKubernetesResource retrieves a Kubernetes object from the cluster.
+// It accepts wait options and returns an error if the resource cannot be found within the given conditions.
+func GetKubernetesResource(ctx context.Context, client klient.Client, objKey runtimeClient.ObjectKey,
```
ivanvc, line 303: If this function returns an error if the resource is not found, I think it could have a better name, like `IsKubernetesResourcePresent(...)` or `doesKubernetesResourceExist(...)`. I just find conflicting `Get`, and the function returns not the object, but only an error.

I know this is just an implementation of the original function, but it could be a good opportunity for a rename.

ahrtr: The returned object is `outObj` ( the 4th parameter)

jmhbnz (approved):
LGTM - Thanks @ahrtr

PR #141, opened 2025-05-12
Add `CertManager` interface functions

ArkaSaha30:
This PR will implement the interface functions required for the cert-manager certificate provider:

- `EnsureCertificateSecret`
- `ValidateCertificateSecret`
- `GetCertificateConfig`
- `DeleteCertificateSecret`
- ~`RevokeCertificate`~ : Cert-manger certificates can only be deleted, which is handled by DeleteCertificateSecret as per [official documentation](https://cert-manager.io/docs/usage/certificate/#inner-workings-diagram-for-developers)

Issue ref: https://github.com/etcd-io/etcd-operator/issues/10

Note:
Decreasing the code change volume in https://github.com/etcd-io/etcd-operator/pull/9 for simplicity and readability.


ahrtr:
@ArkaSaha30 please remove the `WIP` and ping me when this PR is ready to review.

ArkaSaha30:
> @ArkaSaha30 please remove the `WIP` and ping me when this PR is ready to review.

Sure, I will add the e2e tests and ping for review

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+	cmmeta "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	interfaces "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+)
+
+type Provider struct {
```
ahrtr, line 27:
```suggestion
type CertManagerProvider struct {
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+	client.Client
+}
+
```
ahrtr, line 30:
```suggestion
var _ Provider = (* CertManagerProvider)(nil)
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			SecretName:  secretName,
+			DNSNames:    cfg.AltNames.DNSNames,
+			IPAddresses: strings.Fields(strings.Trim(fmt.Sprint(cfg.AltNames.IPs), "[]")),
+			IssuerRef: cmmeta.ObjectReference{
+				Name: issuerName,
+				Kind: issuerKind,
+			},
+			Duration: &metav1.Duration{Duration: cfg.ValidityDuration},
```
ahrtr, line 62: why I do not see some fields (i.e. CommonName, Organization) in the Config?

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+// DNSNames and IPAddresses if not user-defined, will be set to default value in runtime:
+// fmt.Sprintf("%s-%d.%s.%s.svc.cluster.local", ec.Name, index, ec.Name, ec.Namespace)
+// returns an error if the Certificate resource cannot be created.
+func (cm *Provider) createCertificate(
+	ctx context.Context,
+	secretName string,
+	namespace string,
+	cfg *interfaces.Config) error {
+	issuerName := cfg.ExtraConfig["issuerName"].(string)
+	issuerKind := cfg.ExtraConfig["issuerKind"].(string)
```
ahrtr, line 47: probably we can define a list of constant keys (string).

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+	case *rsa.PrivateKey:
+		pub := cert.PublicKey.(*rsa.PublicKey)
+		if !key.PublicKey.Equal(pub) {
+			return errors.New("private key(RSA) does not match the public key in the certificate")
+		}
+	case *ecdsa.PrivateKey:
+		pub := cert.PublicKey.(*ecdsa.PublicKey)
+		if !key.PublicKey.Equal(pub) {
+			return errors.New("private key(ECDSA) does not match the public key in the certificate")
+		}
```
ahrtr, line 101: reference: https://github.com/etcd-io/etcd/blob/14448c52f598a923c2d40c48d3468df6aa3dedb8/server/auth/jwt.go#L161-L174

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+			secretName, namespace, valErr)
+	}
+	if checkCertSecret {
+		return fmt.Errorf("valid certificate secret: %s already present in namespace: %s , skipping Certificate creation",
+			secretName, namespace)
+	}
+	log.Printf("certificate secret: %s not present in namespace: %s , creating new Certificate",
+		secretName, namespace)
+
+	return cm.createCertificate(ctx, secretName, namespace, cfg)
```
ahrtr, line 182: If the certManager is still in progress of creating the secret, so the ValidateCertificateSecret may fail to get/read the certificate, so etcd-operator will create the certificateResource again? It looks like a race condition.

ahrtr: There should be official Cert-manager API for users (etcd-operator in this case) to check the progress.

ahrtr: https://cert-manager.io/docs/usage/certificate/#inner-workings-diagram-for-developers

We should read the Certificate resource, and check the field `reason`,
- `Pending`
- `Issued`

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func (cm *Provider) DeleteCertificateSecret(ctx context.Context, secretName string, namespace string) error {
+	secret := &corev1.Secret{}
+	err := cm.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, secret)
+	if err != nil {
+		return fmt.Errorf("failed to get secret: %w", err)
+	}
```
ahrtr, line 190: We also need to Certificate status (reason). What if it's still in progress?

pkg/certificate/cert_manager/provider.go
ahrtr: If there is something wrong on the cert-manager side, then it may get stuck on creating the secret (the reason will be "Pending" for a long long time), we need to think about how to resolve it.
- Solution 1: Let users to manually resolve it ( because the cert-manager is deployed by users), and need human intervention.When users cancel/remove the certificate resource, etcd-operator will automatically recreate it.
- Solution 2: etcd-operator automaticaly cancel it if the secret isn't ready for a long time (a timeout parameter is needed from users, and have a default value), and recreate the certificate resource.

ArkaSaha30: > why I do not see some fields (i.e. CommonName, Organization) in the Config?

Updated the missing fields.

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,313 @@
...
+		secretName, namespace)
+
+	return cm.createCertificate(ctx, secretName, namespace, cfg)
+}
+
+func (cm *CertManagerProvider) ValidateCertificateSecret(
+	ctx context.Context,
+	secretName string,
+	namespace string,
+	_ *interfaces.Config) (bool, error) {
```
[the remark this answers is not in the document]
ahrtr: After second thought, I do think we should only return `error`. If it's an `nil`, then it means validation successful, otherwise not. It makes no any sense to return another `bool` value.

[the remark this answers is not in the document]
ArkaSaha30: Yes agree, I will update the PR with the change

ahrtr:
@ArkaSaha30 I see that the workflow check still fails. Are you still working on it?

ArkaSaha30:
> @ArkaSaha30 I see that the workflow check still fails. Are you still working on it?

Yes, I'm fixing it, along with the Certificate creation checks.

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	if checkKeyPairErr := checkKeyPair(parseCert, privateKey); checkKeyPairErr != nil {
+		return false, fmt.Errorf("private key does not match certificate: %w", checkKeyPairErr)
+	}
+
+	return true, nil
+}
+
+func (cm *CertManagerProvider) DeleteCertificateSecret(ctx context.Context, secretName string, namespace string) error {
```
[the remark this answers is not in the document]
ArkaSaha30: When we delete the secret and not the certificate, cert-manager controller reconciles and creates a new secret linked to the existing certificate.

The inverse is not true, deleting the certificate does not delete the secret, so it needs to be done as an additional operation.

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,313 @@
...
+
+var _ interfaces.Provider = (*CertManagerProvider)(nil)
+
+func New(c client.Client) *CertManagerProvider {
+	return &CertManagerProvider{
+		c,
+	}
+}
+
+func (cm *CertManagerProvider) EnsureCertificateSecret(
```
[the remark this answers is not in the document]
ahrtr: The key point is that we need to ensure `EnsureCertificateSecret` is idempotent. 

```
if the certificate exist {
   check the certificate status
   if status is pending {
       return Pending
   }
 
   if status is Ready {
       validate the secret
       if successful {
              return nil
       } else {
              return error
       }
   }

    return whatever error
} else {
  // not exist
  validate the input parameter in Config
  if validation is successful {
       create the certificate
  } else {
       return error
  }
}
```

ahrtr:
All the workflow checks are green, which is good.

Please resolve all the review comments, and feedback when it's ready for next round of review.

pkg/certificate/cert_manager/provider.go
ArkaSaha30: Updated the logic to include this.

pkg/certificate/cert_manager/provider.go
ArkaSaha30: Added the check

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return nil
+}
+
+func (cm *Provider) RevokeCertificate(ctx context.Context, secretName string, namespace string) error {
+	cmCertificate := &certmanagerv1.Certificate{}
+	getCertificateErr := cm.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, cmCertificate)
+	if getCertificateErr != nil {
+		return getCertificateErr
+	}
```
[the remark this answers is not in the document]
ArkaSaha30: Yes, we can use `&certmanagerv1.Certificate{}.Status.Conditions` to check

pkg/certificate/cert_manager/provider.go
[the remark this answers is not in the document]
ArkaSaha30: Updated the `EnsureCertificateSecret` logic

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+	IssuerKindKey = "issuerKind"
+)
+
+type CertManagerProvider struct {
+	client.Client
+}
+
+var _ interfaces.Provider = (*CertManagerProvider)(nil)
+
+func New(c client.Client) *CertManagerProvider {
```
ahrtr:
```suggestion
func New(c client.Client) interfaces.Provider {
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+			}
+			err := cm.createCertificate(ctx, secretName, namespace, cfg)
+			if err != nil {
+				return err
+			}
+		}
+		return checkCertExists
+	}
+
+	log.Printf("valid certificate: %s present in namespace: %s, checking certificate status...", secretName, namespace)
```
ahrtr:
```suggestion
	log.Printf("Valid certificate: %s present in namespace: %s, checking certificate status...", secretName, namespace)
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+	cmStatus := cmCertificate.Status.Conditions
+	for _, condition := range cmStatus {
+		switch condition.Type {
+		case certmanagerv1.CertificateConditionReady:
+			log.Printf("Certificate Ready: %v (Reason: %s, Message: %s)\n",
+				condition.Status, condition.Reason, condition.Message)
+			return nil
+		case certmanagerv1.CertificateConditionIssuing:
+			return fmt.Errorf("Certificate Issuing: %v (Reason: %s, Message: %s)\n",
```
ahrtr: Define a pending error something like below,

```
var ErrPending = errors.New(".....")

// Users need to use errors.Is to check the returned error whether it's ErrPending
fmt.Errorf("xxxx: %w", ErrPending)

```

ahrtr: Please use lower case for the first letter for any returned error

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+				condition.Status, condition.Reason, condition.Message)
+		default:
+			return fmt.Errorf("Certificate status unknown: %v (Reason: %s, Message: %s)\n",
+				condition.Status, condition.Reason, condition.Message)
+		}
```
ahrtr: ditto

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+			IssuerNameKey: cmCertificate.Spec.IssuerRef.Name,
+			IssuerKindKey: cmCertificate.Spec.IssuerRef.Kind,
+		},
+	}
+
+	return cfg, nil
+}
+
+// getCertificateStatus returns the current status of the certificate creation
+func (cm *CertManagerProvider) getCertificateStatus(certificateName, namespace string, ctx context.Context) error {
```
ahrtr:
```suggestion
func (cm *CertManagerProvider) checkCertificateStatus(certificateName, namespace string, ctx context.Context) error {
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+
+func New(c client.Client) *CertManagerProvider {
+	return &CertManagerProvider{
+		c,
+	}
+}
+
+func (cm *CertManagerProvider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string,
+	cfg *interfaces.Config) error {
+	checkCertExists := cm.checkCertificateExists(secretName, namespace, ctx)
```
ahrtr: move the implementation here (remove checkCertificateExists)

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// getExtraConfigByName checks if the ExtraConfig keys has been defined with a valid string
+func getExtraConfigByName(key string, cfg *interfaces.Config) (string, error) {
+	value, isValid := cfg.ExtraConfig[key].(string)
+	if !isValid {
+		return "", fmt.Errorf("value for %s not correctly provided, try again", key)
+	}
+	return value, nil
+}
```
ahrtr: Two approaches:
- add a generic method below into the interface package
- [preferred] or just remove this function,  because it's too simple. The golang map already returns two values: value and bool
```
func (cfg *Config) GetExtraConfigByName(key string) (any, bool)
{

}
```

ArkaSaha30:
cc @ivanvc @jmhbnz @hakman

pkg/certificate/interfaces/interface.go
```diff
@@ -2,10 +2,23 @@ package certificate
...
+var ErrPending = errors.New("certificate creation pending")
+
+// ErrUnknown is returned when the Certificate status does not match the provider defined states
+var ErrUnknown = errors.New("certificate status unknown")
+
+// MaxRetries is the max number a function should be re-triggered in case of error
+const MaxRetries = 5
+
+// RetryInterval is the time to wait before the function is re-triggered
+const RetryInterval = time.Minute
```
ahrtr: 1 minute and 5 retries is too long. Based on your test, usually how long does it take to generate a secret? And Is there any best practice provided by cert-manager?

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,368 @@
...
+		} else {
+			return dSecretErr
+		}
+
+	}
+
+	return nil
+}
+
+// RevokeCertificate does not exist, certificates can only be deleted which is handled by DeleteCertificateSecret as per
```
ahrtr:
```suggestion
// RevokeCertificate isn't supported, certificates can only be deleted which is handled by DeleteCertificateSecret as per
```

test/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+		})
+
+	feature.Assess("Ensure certificate",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			cmProvider := cert_manager.New(client.Resources().GetControllerRuntimeClient())
+			err := cmProvider.EnsureCertificateSecret(ctx, cmCertificateName, cmCertificateNamespace, cmConfig)
+			if err != nil {
+				t.Fatalf("Cert-Manager Certificate could not be created: %v", err)
+			}
```
ahrtr, line 89: I think we should have a retry mechanism here.

pkg/certificate/interfaces/interface.go
ArkaSaha30: In my testing, it 2/5 times it had to retry once i.e. wait for 1 minute and retry.
Since the standard timeout for k8s resources is 5mins, I added a retry of 5 times with 1 min wait.

test/e2e/cert_manager_test.go
ArkaSaha30: Do you suggest I remove the retry logic in the `provider.go` and add it here?

ahrtr: I think we should add retry logic in both places.

pkg/certificate/interfaces/interface.go
ahrtr: setting a smaller interval (i.e. 5s), and larger retries makes more sense. 

For example, it takes the cert-manager 1min and 5 seconds, in current implementation, you will have to wait for 2 minutes.

pkg/certificate/interfaces/interface.go
```diff
@@ -2,10 +2,23 @@ package certificate
...
+	"errors"
 	"net"
 	"time"
 )
 
+// ErrPending is returned when the Certificate is not in "Ready" state
+var ErrPending = errors.New("certificate creation pending")
+
+// ErrUnknown is returned when the Certificate status does not match the provider defined states
+var ErrUnknown = errors.New("certificate status unknown")
```
ahrtr:
```suggestion
var (
// ErrPending is returned when the Certificate is not in "Ready" state
 ErrPending = errors.New("certificate creation pending")

// ErrUnknown is returned when the Certificate status does not match the provider defined states
 ErrUnknown = errors.New("certificate status unknown")
}
```

pkg/certificate/interfaces/interface.go
ahrtr:
```suggestion
const (
// MaxRetries is the max number a function should be re-triggered in case of error
 MaxRetries = 5

// RetryInterval is the time to wait before the function is re-triggered RetryInterval = time.Minute
)
```

test/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind: cmIssuerType,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cmIssuerName,
+		},
+		Spec: certv1.IssuerSpec{
+			IssuerConfig: certv1.IssuerConfig{
+				SelfSigned: &certv1.SelfSignedIssuer{},
+			}},
```
ahrtr:
```suggestion
			},	
		},
```

test/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	cmConfig := &interfaces.Config{
+		CommonName:       cmCertificateName,
+		ValidityDuration: cmCertificatealidity,
+		ExtraConfig: map[string]any{
+			"issuerName": cmIssuerName,
+			"issuerKind": cmIssuerType,
+		},
+	}
+
```
ahrtr, line 56: Create the provider only once, and reuse it later.
```suggestion
cmProvider := cert_manager.New(client.Resources().GetControllerRuntimeClient())
```

test/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+				t.Fatalf("Failed to validate Cert-Manager Certificate secret: %v", err)
+			}
+			return ctx
+		})
+
+	feature.Assess("Get certificate config",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			cmProvider := cert_manager.New(client.Resources().GetControllerRuntimeClient())
+			_, err := cmProvider.GetCertificateConfig(ctx, cmCertificateName, cmCertificateNamespace)
```
ahrtr: Can you verify the returned config match the original config that  is used to call EnsureCertificateSecret?

test/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+				t.Fatalf("Cert-Manager Certificate not found: %v", err)
+			}
+			return ctx
+		})
+
+	feature.Assess("Delete certificate secret",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			cmProvider := cert_manager.New(client.Resources().GetControllerRuntimeClient())
+			err := cmProvider.DeleteCertificateSecret(ctx, cmCertificateName, cmCertificateNamespace)
```
ahrtr, line 122: You need to verify that the certificate and secret do not exist afterwards.

pkg/certificate/interfaces/interface.go
ArkaSaha30: Updated the MaxRetries and RetryIntervals

test/e2e/cert_manager_test.go
ArkaSaha30: Since `client` is created from the `cfg *envconf.Config` passed to each Assess func which in turn is used to create the `cmProvider`, we can revisit later how to initialize it once and reuse it.

test/e2e/cert_manager_test.go
ArkaSaha30: Sure, updated with the valdiaiton check

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,375 @@
...
+				return err
+			}
+		} else {
+			return err
+		}
+	}
+
+	log.Printf("Valid certificate: %s present in namespace: %s, checking certificate status...", secretName, namespace)
+
+	for try := 1; try <= interfaces.MaxRetries; try++ {
```
ivanvc: I think we can simplify this with

```suggestion
	for try := range interfaces.MaxRetries {
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,375 @@
...
+		log.Printf("Valid certificate secret: retry attempt %v, after %v, error: %v", try, interfaces.RetryInterval, err)
+		time.Sleep(interfaces.RetryInterval)
+	}
+	if err != nil {
+		return err
+	}
+
+	certificateData, exists := secret.Data["tls.crt"]
+	if !exists {
+		return errors.New("certificate not found in secret")
```
ivanvc: Does it make sense to define these errors as exported variables? Maybe not, if we're going to be the only consumers for the errors (and in the end are just used for logging purposes).

pkg/certificate/certificate.go
```diff
@@ -14,12 +17,13 @@ const (
...
 )
 
-func NewProvider(pt ProviderType) (certInterface.Provider, error) {
+func NewProvider(pt ProviderType, c client.Client) (certInterface.Provider, error) {
 	switch pt {
 	case Auto:
 		return nil, nil // change me later
 	case CertManager:
-		return nil, nil // change me later
+
```
ivanvc: nit: I think we can remove this empty line.

ivanvc (commented):
Thanks for the pull request, Arka. I left a couple of minor comments :)

pkg/certificate/cert_manager/provider.go
ArkaSaha30: Sure, updating the code to use `range`:
```
for try := range [interfaces.MaxRetries]int{}
```

pkg/certificate/cert_manager/provider.go
ArkaSaha30: For now, we are the only consumers, and errors will be used for logging. 
Maybe if there is a future use case we can make it as exported variables?

pkg/certificate/certificate.go
ArkaSaha30: Sure!

ArkaSaha30:
> Thanks for the pull request, Arka. I left a couple of minor comments :)

Thank you, I have addressed the review comments.

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,375 @@
...
+
+	for try := 1; try <= interfaces.MaxRetries; try++ {
+		err = cm.checkCertificateStatus(secretName, namespace, ctx)
+		if err == nil {
+			break
+		}
+		// Wait for the certificate to be in "Ready" state
+		// Reference: https://cert-manager.io/docs/usage/certificate/#inner-workings-diagram-for-developers
+		log.Printf("Certificate Status: retry attempt %v, after %v, error: %v", try, interfaces.RetryInterval, err)
+		time.Sleep(interfaces.RetryInterval)
```
ahrtr: We should only retry when the error is pending, otherwise, it makes no sense to retry.

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,375 @@
...
+	for try := range [interfaces.MaxRetries]int{} {
+		err = cm.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, secret)
+		if err == nil {
+			break
+		}
+		// Wait for cert-manager reconciler to create the associated certificate secret
+		// Reference: https://cert-manager.io/docs/usage/certificate/#inner-workings-diagram-for-developers
+		log.Printf("Valid certificate secret: retry attempt %v, after %v, error: %v", try+1, interfaces.RetryInterval, err)
+		time.Sleep(interfaces.RetryInterval)
+	}
```
ahrtr: We should only retry only when the error is NotFound.

pkg/certificate/cert_manager/provider.go
ahrtr: It's OK to define errors below,
- "certificate not found in secret"
- "private key not found in secret"
- ....

ArkaSaha30: Updated the PR as per consensus

pkg/certificate/cert_manager/provider.go
ArkaSaha30: Sure, checking for `if !errors.Is(err, interfaces.ErrUnknown)` since for both `ErrPending` and `does not exist` retry will be required

pkg/certificate/cert_manager/provider.go
```diff
@@ -0,0 +1,385 @@
...
+
+var _ interfaces.Provider = (*CertManagerProvider)(nil)
+
+func New(c client.Client) interfaces.Provider {
+	return &CertManagerProvider{
+		c,
+	}
+}
+
+func (cm *CertManagerProvider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string,
```
ivanvc, line 46: Could we split this function into smaller ones, and have less logic in the top-level function? Does that make sense? It's a bit difficult to follow the function body.


```golang
func (cm *CertManagerProvider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string,
	cfg *interfaces.Config) error {
	cmCertificate := &certmanagerv1.Certificate{}
// wrap the error handling logic into a new function
    err := cm.getCertificate(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, cmCertificate)
    if err != nil {
      return err
    }

    log.Printf("Valid certificate: %s present in namespace: %s, checking certificate status...", secretName, namespace)

// wrap the error-handling logic and the wait for the ready status into a new function
    err = cm.checkCertificateStatusAndWaitForReadyStatus(secretName, namespace, ctx)
    if err != nil {
      return err
    }

	log.Printf("Certificate Status: %s ready in namespace: %s, validating associated secret...", secretName, namespace)

// wrap the error handling logic into a new function
	err = cm.validateCertificateSecret(ctx, secretName, namespace, cfg)
	if err != nil {
		return err
    }

	log.Printf("Valid certificate secret: %s already present in namespace: %s", secretName, namespace)
	return nil
}


// for example, define getCertificate as: 
func (cm *CertManagerProvider) getCertificate(ctx context.Context, objKey *client.ObjectKey, cmCertificate *crtmanagerv1.Certificate{}) error {
	err := cm.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, cmCertificate)
	if err != nil {
		if k8serrors.IsNotFound(err) {
			valErr := cm.validateCertificateConfig(ctx, namespace, cfg)
			if valErr != nil {
				return valErr
			}
			err := cm.createCertificate(ctx, secretName, namespace, cfg)
			if err != nil {
				return err
			}
		} else {
			return err
		}
	}
	return nil
}

```

ivanvc (approved):
LGTM. I left a suggestion. But I'm okay with the current state, or addressing that in a follow-up. Thanks, @ArkaSaha30 :)

PR #149, opened 2025-05-20
refactor datapersistence

SDBrett:
Refactors the data persistence test to use the `GetKubernetesResource` function.

Small adjustment to the STS lookup interval so the options only need to be declared once

E2E test ran without errors after changes made.


ahrtr (approved):

ahrtr:
thx

PR #158, opened 2025-06-09
Update "ValidateCertificateSecret" interface signature

ArkaSaha30:
This PR will update the `ValidateCertificateSecret` interface signature to return only error instead of (bool, error), which will suggest if the validation is successful or not.

cc @ahrtr @ivanvc


ahrtr (approved):
LGTM & thx

PR #180, opened 2025-07-16
feat(controller): Refactor reconcile loop and add unit tests

ballista01:
## Summary

* **Refactors `EtcdClusterReconciler.Reconcile`** into focused phases to reduce cyclomatic complexity and make behavior easier to reason about.
* **Adds a `reconcileState` container** so data gathered in earlier phases can be reused without repeated lookups.
* **Reworks the controller unit tests** to target each phase individually with fake clients instead of the previous envtest-based flow.

---

## Controller changes (`internal/controller/etcdcluster_controller.go`)

* Split the main reconcile loop into:

  * `fetchAndValidateState`
  * `bootstrapStatefulSet`
  * `performHealthChecks`
  * `reconcileClusterState`

  Each returns early with a `ctrl.Result` when it needs a requeue or encounters an error.

* Introduced `reconcileState` (cluster, StatefulSet, member list, member health) to pass context between phases.

* `fetchAndValidateState`:

  * Sets a default image registry.
  * Prepares TLS credentials.
  * Verifies StatefulSet ownership.
  * Signals requeues when dependent resources aren’t yet ready.

* `bootstrapStatefulSet`:

  * Unifies StatefulSet creation/initial scaling and the headless service check.
  * Uses `ctrl.Result.IsZero()` to decide whether to continue.

* `performHealthChecks` and `reconcileClusterState`:

  * Now operate on `memberHealth`.
  * Added comments clarifying learner promotion, scale-out/-in, and health gating.

---

## Tests (`internal/controller/etcdcluster_controller_test.go`)

* Replaced the previous integration-style test with **table-driven unit tests** using controller-runtime’s fake client.
* Added `TestFetchAndValidateState` and `TestBootstrapStatefulSet`, switching to **require** for hard preconditions and asserting on StatefulSet/Service/ConfigMap outcomes.
* Expanded coverage for bootstrap scenarios:

  * Initial creation.
  * 0→1 promotion.
  * No-op when resources exist.
* Verified cluster state remains untouched when nothing changes.

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -55,7 +55,7 @@ type reconcileState struct {
 	cluster        *ecv1alpha1.EtcdCluster      // cluster custom resource currently being reconciled
 	sts            *appsv1.StatefulSet          // associated StatefulSet for the cluster
 	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
-	healthInfos    []etcdutils.EpHealth         // health information for each etcd member
+	memberHealth   []etcdutils.EpHealth         // health information for each etcd member
 }
 
 // +kubebuilder:rbac:groups=operator.etcd.io,resources=etcdclusters,verbs=get;list;watch;create;update;patch;delete
@@ -79,12 +79,12 @@ type reconcileState struct {
 // https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile
 func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	state, res, err := r.fetchAndValidateState(ctx, req)
-	if state == nil || err != nil || res.RequeueAfter != 0 {
+	if state == nil || err != nil {
 		return res, err
 	}
 
-	if primRes, err := r.syncPrimitives(ctx, state); err != nil || primRes.Requeue || primRes.RequeueAfter != 0 { //nolint:staticcheck
-		return primRes, err
+	if bootstrapRes, err := r.bootstrapStatefulSet(ctx, state); err != nil || !bootstrapRes.IsZero() {
+		return bootstrapRes, err
 	}
 
 	if err = r.performHealthChecks(ctx, state); err != nil {
@@ -150,40 +150,41 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 	return &reconcileState{cluster: ec, sts: sts}, ctrl.Result{}, nil
 }
 
-// syncPrimitives ensures that the foundational Kubernetes objects for a cluster
-// exist and are correctly initialized. It creates the StatefulSet (initially
+// bootstrapStatefulSet ensures that the foundational Kubernetes objects for
+// a cluster exist and are correctly initialized. It creates the StatefulSet (initially
 // with 0 replicas) and the headless Service if necessary. When either resource
 // is created or the StatefulSet is scaled from zero to one replica, the returned
 // ctrl.Result requests a requeue so the next reconciliation loop can observe the
 // new state. The reconcileState is updated with the current StatefulSet.
-func (r *EtcdClusterReconciler) syncPrimitives(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
+func (r *EtcdClusterReconciler) bootstrapStatefulSet(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
+	requeue := false
 	var err error
 
-	if s.sts == nil {
+	switch {
+	case s.sts == nil:
 		logger.Info("Creating StatefulSet with 0 replica", "expectedSize", s.cluster.Spec.Size)
-		if s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 0, r.Scheme); err != nil {
+		s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 0, r.Scheme)
+		if err != nil {
 			return ctrl.Result{}, err
 		}
-		if err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{RequeueAfter: requeueDuration}, nil
-	}
+		requeue = true
 
-	if s.sts.Spec.Replicas != nil && *s.sts.Spec.Replicas == 0 {
+	case s.sts.Spec.Replicas != nil && *s.sts.Spec.Replicas == 0:
 		logger.Info("StatefulSet has 0 replicas. Trying to create a new cluster with 1 member")
-		if s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 1, r.Scheme); err != nil {
+		s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 1, r.Scheme)
+		if err != nil {
 			return ctrl.Result{}, err
 		}
-		if err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+		requeue = true
 	}
 
 	if err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
 		return ctrl.Result{}, err
+	}
+
+	if requeue {
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
 	}
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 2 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -60,121 +70,165 @@ type EtcdClusterReconciler struct {
...
+// specification. Each phase is handled by a dedicated helper method.
 //
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/reconcile
+// For more details on the controller-runtime Reconcile contract see:
+// https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile
 func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx)
+	state, res, err := r.fetchAndValidateState(ctx, req)
+	if state == nil || err != nil || res.RequeueAfter != 0 {
```
ahrtr:
```suggestion
	if state == nil || err != nil {
```

internal/controller/etcdcluster_controller.go
```diff
@@ -60,121 +70,165 @@ type EtcdClusterReconciler struct {
...
 
-		sts, err = reconcileStatefulSet(ctx, logger, etcdCluster, r.Client, 1, r.Scheme)
-		if err != nil {
+// syncPrimitives ensures that the foundational Kubernetes objects for a cluster
+// exist and are correctly initialized. It creates the StatefulSet (initially
+// with 0 replicas) and the headless Service if necessary. When either resource
+// is created or the StatefulSet is scaled from zero to one replica, the returned
+// ctrl.Result requests a requeue so the next reconciliation loop can observe the
+// new state. The reconcileState is updated with the current StatefulSet.
+func (r *EtcdClusterReconciler) syncPrimitives(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
```
ahrtr:
```suggestion
func (r *EtcdClusterReconciler) bootstrapStatefulSet(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
```

internal/controller/etcdcluster_controller.go
```diff
@@ -60,121 +70,165 @@ type EtcdClusterReconciler struct {
...
+	logger := log.FromContext(ctx)
+	var err error
+
+	if s.sts == nil {
+		logger.Info("Creating StatefulSet with 0 replica", "expectedSize", s.cluster.Spec.Size)
+		if s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 0, r.Scheme); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
```
ahrtr: remove it to simplify the source code

internal/controller/etcdcluster_controller.go
```diff
@@ -60,121 +70,165 @@ type EtcdClusterReconciler struct {
...
-	err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, etcdCluster, r.Scheme)
-	if err != nil {
+	if s.sts.Spec.Replicas != nil && *s.sts.Spec.Replicas == 0 {
+		logger.Info("StatefulSet has 0 replicas. Trying to create a new cluster with 1 member")
+		if s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 1, r.Scheme); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err = createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
+			return ctrl.Result{}, err
+		}
```
ahrtr: remove it to simplify the source code
```suggestion
```

ahrtr (commented):
The refactoring makes sense to me. Just raised several minor comments. 

Thanks. 

cc @ivanvc @ArkaSaha30 @abdurrehman107 @hakman @frederiko PTAL when you get free cycle, thx

internal/controller/etcdcluster_controller.go
ahrtr: I mean removing the call to `createHeadlessServiceIfNotExist`

internal/controller/etcdcluster_controller.go
```diff
@@ -48,6 +48,16 @@ type EtcdClusterReconciler struct {
...
 }
 
+// reconcileState holds all transient data for a single reconciliation loop.
+// Every phase of Reconcile stores intermediate information here so that
+// subsequent phases can operate without additional lookups.
+type reconcileState struct {
+	cluster        *ecv1alpha1.EtcdCluster      // cluster custom resource currently being reconciled
+	sts            *appsv1.StatefulSet          // associated StatefulSet for the cluster
+	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
+	healthInfos    []etcdutils.EpHealth         // health information for each etcd member
```
ivanvc: Reading this property name out loud makes me wonder if we could rename it to: `memberHealth`

internal/controller/etcdcluster_controller_test.go
```diff
@@ -19,91 +19,313 @@ package controller
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "etcd",
+					Namespace: "default",
+					UID:       "1",
+				},
+				Spec: ecv1alpha1.EtcdClusterSpec{Size: 1, Version: "3.5.17"},
+			},
+			req: ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
+			assert: func(t *testing.T, state *reconcileState, res ctrl.Result, err error, ec *ecv1alpha1.EtcdCluster, _ *appsv1.StatefulSet) {
+				if assert.NotNil(t, state) {
```
ivanvc: What's the value of this if statement? If it fails, either way the test will fail in the end. So, I'm trying to understand why we would want this behavior.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -19,91 +19,313 @@ package controller
...
+					},
+				},
+			},
+			req: ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
+			assert: func(t *testing.T, state *reconcileState, res ctrl.Result, err error, ec *ecv1alpha1.EtcdCluster, sts *appsv1.StatefulSet) {
+				if assert.NotNil(t, state) {
+					assert.Equal(t, ec.Name, state.cluster.Name)
+					if assert.NotNil(t, state.sts) {
+						assert.Equal(t, sts.Name, state.sts.Name)
+					}
```
ivanvc: Same question here.

internal/controller/etcdcluster_controller.go
ballista01: Thanks for the advice. I kept the self-healing logic because the headless Service is what gives the StatefulSet the stable `<pod>.<service>.<namespace>.svc.cluster.local` DNS entries that pop up in the peer/client URLs and in the TLS SANs. If we ever lose that Service (either manually or via a transient failure), we’d end up with broken endpoints unless the reconciler recreates it. The `createHeadlessServiceIfNotExist` call at the end might seem redundant but ensured self-healing in case the service get deleted accidently.

To make the intent clearer, I refactored `syncPrimitives` (and renamed to `bootstrapStatefulSet`) so that we reconcile the `StatefulSet` first and then call `createHeadlessServiceIfNotExist` exactly once at the end.

internal/controller/etcdcluster_controller_test.go
ballista01: Thanks for the question! `testify`’s `assert` helpers only record the failure and carry on, so the original `if assert.NotNil(...)` blocks were just there to stop us from dereferencing `state/state.sts` and panicing. That might seem to be confusing, so I’ve switched those checks to `require.NotNil(...)`. Now the tests fail fast when the state is missing, and the following  `assert.Equal` calls simply verify the expected fields.

internal/controller/etcdcluster_controller.go
ballista01: Removed.

internal/controller/etcdcluster_controller.go
ballista01: Done.

internal/controller/etcdcluster_controller.go
ballista01: Please see the explanation below.

internal/controller/etcdcluster_controller.go
ballista01: Renamed.

internal/controller/etcdcluster_controller_test.go
ballista01: Changed the usage to `require.NotNil(...)`

internal/controller/etcdcluster_controller.go
ballista01: I also recovered and updated some comments (mainly in `reconcileClusterState`). Let me know if we need more!

internal/controller/etcdcluster_controller.go
ahrtr: I wasn't saying to completely remove `createHeadlessServiceIfNotExist`. Instead, I was saying remove redundant invocations.  

I see that you already did the suggested change, thx

internal/controller/etcdcluster_controller_test.go
ivanvc: Thanks, it makes it more understandable :)

internal/controller/etcdcluster_controller.go
ivanvc: Thanks, that looks good.

PR #51, opened 2025-01-21
e2e test cases for the operator

abdurrehman107:
This PR refers to the issue #19. We've used the e2e-framework deliver tests.

PR #192, opened 2025-08-19
test(e2e): add scaling & KV functionality tests; extract shared helpers

ballista01:
## Summary

This PR **extends the E2E suite under `test/e2e/`** with realistic cluster-lifecycle scenarios, **PVC-backed clusters**, and shared helpers to reduce duplication and flakiness. It adds **scaling**, **pod recovery with persistent storage**, and **KV/replication** checks using `etcdctl` **from inside cluster pods**.
**No controller or production code changes** are included.

## What’s in this PR

* **`test/e2e/e2e_test.go`**

  * Keeps existing `TestInvalidClusterSize`, `TestClusterHealthy`.
  * **New:** `TestScaling` (table-driven: `ScaleInFrom3To1`, `ScaleOutFrom1To3`)

    * Verifies the controller updates **StatefulSet replicas**.
    * Checks etcd **member list** count and ensures **no learners** remain after scaling.
  * **New:** `TestPodRecovery`

    * Starts a **3-node PVC-backed** cluster, deletes one pod, waits for recreation, and **verifies it reuses the PVC**.
    * Asserts **member ID consistency** before/after recovery (no unintended re-adds).
    * Validates **replication** by `put` on one member and `get` from the recovered member.
  * **Updated:** `TestEtcdClusterFunctionality`

    * Runs on a **PVC-backed 3-node** cluster.
    * Validates endpoint health; `put/get/del`; cross-member read for **replication**; and ensures **no learners** in a healthy cluster.

* **`test/e2e/helpers_test.go`** (new)

  * **Cluster + storage**

    * `createEtcdClusterWithPVC` — creates an `EtcdCluster` with `StorageSpec` (64Mi) and PVC.
    * `getAvailableStorageClass` — resolves StorageClass via **`ETCD_E2E_STORAGECLASS` env** or default SC; **skips** if none suitable.
  * **Lifecycle / waiting**

    * `waitForSTSReadiness` — waits for **StatefulSet `ReadyReplicas`** using e2e-framework `wait` (tuned timeouts/intervals).
    * `scaleEtcdCluster` — patches `spec.size` and waits for STS readiness.
    * `cleanupEtcdCluster` — best-effort deletion.
  * **Introspection**

    * `execInPod` — runs `etcdctl` inside pods and returns stdout/stderr.
    * `getEtcdMembers` — parses `etcdctl member list -w json` into typed structs to build **name → memberID** map (more robust than string parsing).
    * `verifyPodUsesPVC` — asserts a pod mounts the expected PVC (by prefix).
  * **Types:** `EtcdMember`, `EtcdMemberList`.

## Motivation

* Address prior review feedback to **split the large refactor** into focused pieces.
* Add coverage for scenarios that require a real cluster (DNS/networking/storage) and are not reliable at unit-test level:

  * **Scaling:** keeps `spec.size`, **StatefulSet replicas**, and etcd **member list** in sync; ensures learners are promoted.
  * **Persistence & recovery:** validates **PVC reuse** and **member ID stability** across pod recreation.
  * **Data-path checks:** `etcdctl` health + KV + cross-member reads to prove **replication**.

## How it validates behavior

* Executes `etcdctl` **inside pods** (`execInPod`) to observe **internal cluster state** (health, member list, KV).
* Uses **JSON parsing** for membership (`getEtcdMembers`) to avoid brittle string checks.
* `waitForSTSReadiness` and targeted `wait.For` loops reduce flakes by waiting for **ReadyReplicas** and **pod recreation**.
* Storage-aware helpers ensure tests run with **PVC-backed** clusters and **skip gracefully** when no StorageClass is available.

## Prior work
Huge thanks to **#51** by @abdurrehman107! It inspired the feature-based layout with sigs.k8s.io/e2e-framework, the **Setup - Assess - Teardown** flow and the use of wait.For + conditions.ResourceScaled for readiness. Appreciate the groundwork!

## Notes for reviewers

* If your cluster has no default StorageClass, it's best to **`ETCD_E2E_STORAGECLASS=<your-sc>`** before running E2E.
* Timeouts are conservative (**5m** for STS readiness; **3m** for pod recreation) to minimize flakes; happy to tune based on CI signal.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -125,3 +125,263 @@ func TestClusterHealthy(t *testing.T) {
 	// 'testEnv' is the env.Environment you set up in TestMain
 	_ = testEnv.Test(t, feature.Feature())
 }
+
+func TestScaling(t *testing.T) {
+	testCases := []struct {
+		name            string
+		initialSize     int
+		scaleTo         int
+		expectedMembers int
+	}{
+		{name: "ScaleInFrom3To1", initialSize: 3, scaleTo: 1, expectedMembers: 1},
+		{name: "ScaleOutFrom1To3", initialSize: 1, scaleTo: 3, expectedMembers: 3},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			feature := features.New(tc.name)
+			etcdClusterName := fmt.Sprintf("etcd-%s", strings.ToLower(tc.name))
+
+			feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+				createEtcdClusterWithPVC(ctx, t, c, etcdClusterName, tc.initialSize)
+				waitForSTSReadiness(t, c, etcdClusterName, tc.initialSize)
+				return ctx
+			})
+
+			feature.Assess(
+				fmt.Sprintf("scale to %d", tc.scaleTo),
+				func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+					scaleEtcdCluster(ctx, t, c, etcdClusterName, tc.scaleTo)
+					// Wait until StatefulSet spec/status reflect the scaled size and are ready
+					waitForSTSReadiness(t, c, etcdClusterName, tc.scaleTo)
+					return ctx
+				},
+			)
+
+			feature.Assess("verify member list", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+				podName := fmt.Sprintf("%s-0", etcdClusterName)
+				ml := getEtcdMemberListPB(t, c, podName)
+
+				if len(ml.Members) != tc.expectedMembers {
+					t.Errorf("Expected to find %d members in member list, but got %d",
+						tc.expectedMembers, len(ml.Members))
+				}
+
+				// Verify controller promoted all learners to voting members
+				for _, m := range ml.Members {
+					if m.IsLearner {
+						t.Errorf("Found unpromoted learner after scaling completed: member %s (%d)", m.Name, m.ID)
+					}
+				}
+				return ctx
+			})
+
+			feature.Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+				cleanupEtcdCluster(ctx, t, c, etcdClusterName)
+				return ctx
+			})
+
+			_ = testEnv.Test(t, feature.Feature())
+		})
+	}
+}
+
+func TestPodRecovery(t *testing.T) {
+	feature := features.New("multi-node-pod-recovery")
+	etcdClusterName := "etcd-recovery-test"
+
+	feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		createEtcdClusterWithPVC(ctx, t, c, etcdClusterName, 3)
+		waitForSTSReadiness(t, c, etcdClusterName, 3)
+		return ctx
+	})
+
+	feature.Assess("verify multi-node cluster recovery after pod deletion",
+		func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			var sts appsv1.StatefulSet
+			if err := c.Client().Resources().Get(ctx, etcdClusterName, namespace, &sts); err != nil {
+				t.Fatalf("Failed to get StatefulSet: %v", err)
+			}
+
+			initialReplicas := *sts.Spec.Replicas
+			podName := fmt.Sprintf("%s-0", etcdClusterName)
+			targetPodName := fmt.Sprintf("%s-1", etcdClusterName)
+
+			// Verify PVC usage before test
+			verifyPodUsesPVC(t, c, targetPodName, "etcd-data-"+etcdClusterName)
+
+			// Get initial member IDs for consistency verification
+			initialMembers := getEtcdMembersName2IDMapping(t, c, podName)
+			initialMemberCount := len(initialMembers)
+
+			// Delete one pod to simulate failure
+			var pod corev1.Pod
+			if err := c.Client().Resources().Get(ctx, targetPodName, namespace, &pod); err != nil {
+				t.Fatalf("Failed to get pod: %v", err)
+			}
+
+			deletedPodUID := pod.UID
+			if err := c.Client().Resources().Delete(ctx, &pod); err != nil {
+				t.Fatalf("Failed to delete pod: %v", err)
+			}
+
+			// Wait for pod recreation
+			if err := wait.For(func(ctx context.Context) (done bool, err error) {
+				var newPod corev1.Pod
+				if err := c.Client().Resources().Get(ctx, targetPodName, namespace, &newPod); err != nil {
+					return false, nil
+				}
+				return newPod.UID != deletedPodUID && newPod.Status.Phase == corev1.PodRunning, nil
+			}, wait.WithTimeout(3*time.Minute), wait.WithInterval(10*time.Second)); err != nil {
+				t.Fatalf("Pod failed to be recreated: %v", err)
+			}
+
+			// Wait for full StatefulSet readiness
+			waitForSTSReadiness(t, c, etcdClusterName, int(initialReplicas))
+
+			// Verify PVC usage after recovery
+			verifyPodUsesPVC(t, c, targetPodName, "etcd-data-"+etcdClusterName)
+
+			// Verify member ID consistency
+			finalMembers := getEtcdMembersName2IDMapping(t, c, podName)
+			if len(finalMembers) != initialMemberCount {
+				t.Errorf("Member count changed after recovery: expected %d, got %d", initialMemberCount, len(finalMembers))
+			}
+
+			// Verify the recovered pod maintains the same member ID
+			if targetMemberID, exists := initialMembers[targetPodName]; exists {
+				if finalMemberID, exists := finalMembers[targetPodName]; exists {
+					if targetMemberID != finalMemberID {
+						t.Errorf("Member ID changed for %s: expected %d, got %d", targetPodName, targetMemberID, finalMemberID)
+					}
+				} else {
+					t.Errorf("Member %s not found after recovery", targetPodName)
+				}
+			}
+
+			// Verify cluster replication works across recovered pod
+			_, stderr, err := execInPod(t, c, podName, namespace, []string{"etcdctl", "put", "replication-test", "value"})
+			if err != nil {
+				t.Fatalf("Failed to write data after recovery: %v, stderr: %s", err, stderr)
+			}
+
+			stdout, stderr, err := execInPod(t, c, targetPodName, namespace, []string{"etcdctl", "get", "replication-test"})
+			if err != nil {
+				t.Fatalf("Failed to read data from recovered pod: %v, stderr: %s", err, stderr)
+			}
+
+			if !strings.Contains(stdout, "value") {
+				t.Errorf("Data replication to recovered pod failed. Expected 'value', got: %s", stdout)
+			}
+
+			return ctx
+		})
+
+	feature.Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		cleanupEtcdCluster(ctx, t, c, etcdClusterName)
+		return ctx
+	})
+
+	_ = testEnv.Test(t, feature.Feature())
+}
+
+func TestEtcdClusterFunctionality(t *testing.T) {
+	feature := features.New("etcd-cluster-functionality")
+	etcdClusterName := "etcd-functionality-test"
+	testKey := "test-key"
+	testValue := "test-value"
+
+	feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		createEtcdClusterWithPVC(ctx, t, c, etcdClusterName, 3)
+		waitForSTSReadiness(t, c, etcdClusterName, 3)
+		return ctx
+	})
+
+	feature.Assess("verify cluster health", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		podName := fmt.Sprintf("%s-0", etcdClusterName)
+
+		// Wait until all members are promoted to voting members (no learners),
+		// otherwise endpoint health will fail on learners.
+		waitForNoLearners(t, c, podName, 3)
+
+		// Check health for the whole cluster rather than a single member
+		command := []string{"etcdctl", "endpoint", "health", "--cluster"}
+		stdout, stderr, err := execInPod(t, c, podName, namespace, command)
+		if err != nil {
+			t.Fatalf("Failed to check endpoint health: %v, stderr: %s", err, stderr)
+		}
+
+		// Determine expected members dynamically from member list
+		ml := getEtcdMemberListPB(t, c, podName)
+		expected := len(ml.Members)
+
+		// Count healthy lines; expect all members are healthy
+		healthy := 0
+		for _, line := range strings.Split(strings.TrimSpace(stdout), "\n") {
+			if strings.Contains(line, "is healthy") {
+				healthy++
+			}
+		}
+		if healthy != expected {
+			t.Errorf("Expected %d healthy endpoints, but got %d. Output: %s", expected, healthy, stdout)
+		}
+		return ctx
+	})
+
+	feature.Assess("test data operations", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		podName := fmt.Sprintf("%s-0", etcdClusterName)
+
+		// Write key-value data
+		command := []string{"etcdctl", "put", testKey, testValue}
+		_, stderr, err := execInPod(t, c, podName, namespace, command)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -125,3 +125,118 @@ func TestClusterHealthy(t *testing.T) {
...
+
+func TestScaling(t *testing.T) {
+	testCases := []struct {
+		name            string
+		initialSize     int
+		scaleTo         int
+		expectedMembers int
+	}{
+		{"ScaleDownFrom3To1", 3, 1, 1},
+		{"ScaleUpFrom1To3", 1, 3, 3},
```
ahrtr: Also suggest to add field name
```suggestion
		{"ScaleInFrom3To1", 3, 1, 1},
		{"ScaleOutFrom1To3", 1, 3, 3},
```

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,118 @@ func TestClusterHealthy(t *testing.T) {
...
+	feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+		createEtcdCluster(ctx, t, c, etcdClusterName, 2)
+		waitForStsReady(t, c, etcdClusterName, 2)
+		// Manually add a third member as a learner
+		podName := fmt.Sprintf("%s-0", etcdClusterName)
+		command := []string{
+			"etcdctl", "member", "add", "etcd-promote-learner-2",
+			"--peer-urls=http://etcd-promote-learner-2.etcd-promote-learner.etcd-operator-system.svc.cluster.local:2380",
+			"--learner",
+		}
```
ahrtr: If TLS is enabled, the command will fail. More importantly, why adding member directly? It should be the responsibility of the controller.

ahrtr:
also cc @frederiko

test/e2e/e2e_test.go
ballista01: Changed to 
```
{name: "ScaleInFrom3To1", initialSize: 3, scaleTo: 1, expectedMembers: 1},
{name: "ScaleOutFrom1To3", initialSize: 1, scaleTo: 3, expectedMembers: 3},
``` 
https://github.com/etcd-io/etcd-operator/pull/192/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R136-R137

test/e2e/e2e_test.go
ballista01: I agree, maybe we should keep the test case in unit test rather than make it E2E.https://github.com/etcd-io/etcd-operator/blob/267dbb96e08c332cecca3bd698d64d07c7e6ac55/internal/etcdutils/etcdutils_test.go#L122
Removed.

ballista01 (commented):
<outdated comment>

test/e2e/e2e_test.go
```diff
@@ -33,7 +33,7 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/features"
 
-	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+	ec_v1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
```
ivanvc: nit: Why did you change the name of this import? If we rename, it will be inconsistent with other files.

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,241 @@ func TestClusterHealthy(t *testing.T) {
...
+				cleanupEtcdCluster(ctx, t, c, etcdClusterName)
+				return ctx
+			})
+
+			_ = testEnv.Test(t, feature.Feature())
+		})
+	}
+}
+
+// TODO: TestPodRecovery is commented out - would like maintainer feedback on expected behavior
```
ivanvc: I think this is worth discussing in a different issue.

In what scenarios would the pod get deleted? The scheduler shouldn't delete the pod, unless I'm missing/misunderstanding something.

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,113 @@
...
+			Size:    size,
+			Version: etcdVersion,
+		},
+	}
+	if err := c.Client().Resources().Create(ctx, etcdCluster); err != nil {
+		t.Fatalf("Failed to create EtcdCluster: %v", err)
+	}
+}
+
+func waitForStsReady(t *testing.T, c *envconf.Config, name string, expectedReplicas int) {
```
ivanvc: Suggestion to use Go-style for the initialism.

```suggestion
func waitForSTSReadiness(t *testing.T, c *envconf.Config, name string, expectedReplicas int) {
```

```suggestion
func waitForTheSTSToBeReady(t *testing.T, c *envconf.Config, name string, expectedReplicas int) {
```

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,113 @@
+package e2e
```
ivanvc, line 17: Could you please add the license header to this file?

test/e2e/e2e_test.go
ballista01: Changed back to the original alias for consistency.

test/e2e/helpers_test.go
ballista01: Changed to `waitForSTSReadiness`.

test/e2e/helpers_test.go
ballista01: License header added. https://github.com/etcd-io/etcd-operator/pull/192/files#diff-0a46baa7716e5cf9a396958c52643c1515750e55f917c4e20c3cf51991514707R1-R15

test/e2e/e2e_test.go
ballista01: Thanks for the review! I've created an issue at https://github.com/etcd-io/etcd-operator/issues/193 for discussion.

ballista01: Quick update: This only repros when the cluster is created without persistent storage, or when a Pod and its PVC are removed together (effective permanent loss). With PVC-backed clusters (typical in production), deleting a Pod behaves as expected: the StatefulSet recreates the same ordinal with its PVC and the member rejoins; no downscale.

If you prefer, I’ll close #193 as not reproducible for the transient case and send a small docs PR clarifying the above.

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,278 @@ func TestClusterHealthy(t *testing.T) {
...
+
+				memberLines := strings.Split(strings.TrimSpace(stdout), "\n")
+				if len(memberLines) != tc.expectedMembers {
+					t.Errorf("Expected to find %d members in member list, but got %d. Output: %s",
+						tc.expectedMembers, len(memberLines), stdout)
+				}
+
+				// Verify controller promoted all learners to voting members
+				for _, line := range memberLines {
+					if strings.Contains(line, "isLearner=true") {
```
ahrtr: Have you manually verified the output of command `etcdctl member list`?

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,225 @@
...
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    size,
+			Version: etcdVersion,
+			StorageSpec: &ecv1alpha1.StorageSpec{
+				AccessModes:       corev1.ReadWriteOnce,
+				StorageClassName:  storageClassName,
+				PVCName:           fmt.Sprintf("%s-pvc", name),
```
ahrtr: Where is the PVC coming from?

https://github.com/etcd-io/etcd-operator/blob/267dbb96e08c332cecca3bd698d64d07c7e6ac55/api/v1alpha1/etcdcluster_types.go#L144-L150

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,278 @@ func TestClusterHealthy(t *testing.T) {
...
+
+			// Clean up - delete the key
+			podName = fmt.Sprintf("%s-0", etcdClusterName)
+			command = []string{"etcdctl", "del", testKey}
+			_, stderr, err = execInPod(t, c, podName, namespace, command)
+			if err != nil {
+				t.Fatalf("Failed to delete data: %v, stderr: %s", err, stderr)
+			}
+			return ctx
+		})
```
ahrtr: If you really want to verify the data consistent, you should use `etcdctl endpoint hashkv -w json`

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,225 @@
...
+	ID         uint64   `json:"ID"`
+	Name       string   `json:"name"`
+	PeerURLs   []string `json:"peerURLs"`
+	ClientURLs []string `json:"clientURLs"`
+}
+
+// EtcdMemberList represents the output of etcdctl member list -w json
+type EtcdMemberList struct {
+	Members []EtcdMember `json:"members"`
+}
```
ahrtr: pls do not define new structs, use etcd's existing definitions for each kind of response instead.

test/e2e/e2e_test.go
ballista01: Switched to parsing etcdctl JSON with etcd’s official types instead of string matching.

test/e2e/helpers_test.go
ballista01: Corrected RWO storage to rely on StatefulSet `volumeClaimTemplates`

test/e2e/e2e_test.go
ballista01: Added a consistency check using `etcdctl endpoint hashkv -w json` executed inside pods. I compare the hash across members and expect them to match.

test/e2e/helpers_test.go
ballista01: Removed custom structs and now unmarshal into `etcdserverpb.MemberListResponse`

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,265 @@ func TestClusterHealthy(t *testing.T) {
...
+		podName := fmt.Sprintf("%s-0", etcdClusterName)
+		command := []string{"etcdctl", "endpoint", "health"}
+		stdout, stderr, err := execInPod(t, c, podName, namespace, command)
+		if err != nil {
+			t.Fatalf("Failed to check endpoint health: %v, stderr: %s", err, stderr)
+		}
+
+		if !strings.Contains(stdout, "is healthy") {
+			t.Errorf("Expected healthy endpoints, but got: %s", stdout)
+		}
```
ahrtr: You only check the health of one member. Should you check all members' health?

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,284 @@
...
+			return &nested.HashKV
+		}
+		// 3) Flat fields at element level (unknown keys ignored)
+		var flat etcdserverpb.HashKVResponse
+		if err := json.Unmarshal(elems[0], &flat); err == nil &&
+			(flat.Hash != 0 || flat.HashRevision != 0 ||
+				flat.CompactRevision != 0 || flat.Header != nil) {
+			return &flat
+		}
+	}
```
ahrtr: Why it's so complicated to parse the response? If you specify `--cluster`, then you will get all members' responses, otherwise it's just one member's response.

test/e2e/e2e_test.go
ballista01: It's now using `--cluster` to check all members' health. I also added `waitForNoLearners`.

test/e2e/helpers_test.go
ballista01: Thanks for the reminder. I've updated the usage with `--cluster` and simplified parsing.

test/e2e/e2e_test.go
```diff
@@ -125,3 +125,270 @@ func TestClusterHealthy(t *testing.T) {
...
+					scaleEtcdCluster(ctx, t, c, etcdClusterName, tc.scaleTo)
+
+					// Verify controller updated StatefulSet replicas
+					var sts appsv1.StatefulSet
+					if err := c.Client().Resources().Get(ctx, etcdClusterName, namespace, &sts); err != nil {
+						t.Fatalf("Failed to get StatefulSet: %v", err)
+					}
+					if sts.Spec.Replicas == nil || *sts.Spec.Replicas != int32(tc.scaleTo) {
+						t.Errorf("Controller failed to update StatefulSet replicas: expected %d, got %d", tc.scaleTo, *sts.Spec.Replicas)
+					}
```
ahrtr: `waitForSTSReadiness` (called in `scaleEtcdCluster`) already verified the sts. Replica.

Or you can remove the calling of `waitForSTSReadiness` in `scaleEtcdCluster`, and call `waitForSTSReadiness` directly here.
```suggestion
```

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,280 @@
...
+	stdout, stderr, err := execInPod(t, c, podName, namespace, []string{"etcdctl", "member", "list", "-w", "json"})
+	if err != nil {
+		t.Fatalf("Failed to get etcd member list: %v, stderr: %s", err, stderr)
+	}
+
+	var memberList etcdserverpb.MemberListResponse
+	if err := json.Unmarshal([]byte(stdout), &memberList); err != nil {
+		t.Fatalf("Failed to parse etcd member list JSON: %v", err)
+	}
+
```
ahrtr: You can just call `getEtcdMemberListPB` directly?

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,280 @@
...
+	var etcdCluster ecv1alpha1.EtcdCluster
+	if err := c.Client().Resources().Get(ctx, name, namespace, &etcdCluster); err == nil {
+		if err := c.Client().Resources().Delete(ctx, &etcdCluster); err != nil {
+			t.Logf("Failed to delete EtcdCluster: %v", err)
+		}
+	}
+}
+
+// getEtcdMembers retrieves the etcd cluster member list as name->ID mapping using etcd's native types
+func getEtcdMembers(t *testing.T, c *envconf.Config, podName string) map[string]uint64 {
```
ahrtr:
```suggestion
func getEtcdMembersName2IDMapping(t *testing.T, c *envconf.Config, podName string) map[string]uint64 {
```

test/e2e/e2e_test.go
ballista01: Removed `waitForSTSReadiness` from `scaleEtcdCluster`. I also added `sts.Spec.Replicas` and `sts.Status.ObservedGeneration` checks to harden `waitForSTSReadiness`.

test/e2e/helpers_test.go
ballista01: Yes, that's an overlook. Changed to `memberList := getEtcdMemberListPB(t, c, podName)`

test/e2e/helpers_test.go
ballista01: done

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,285 @@
...
+		t.Fatalf("Failed to get cluster endpoint hashkv from %s: %v, stderr: %s", podName, err, stderr)
+	}
+
+	// Expected JSON: array of objects like {"Endpoint":"...","HashKV":{...}}
+	var entries []struct {
+		Endpoint string                      `json:"Endpoint"`
+		HashKV   etcdserverpb.HashKVResponse `json:"HashKV"`
+	}
+	if err := json.Unmarshal([]byte(stdout), &entries); err != nil {
+		t.Fatalf("Failed to parse endpoint hashkv JSON: %v. Raw: %s", err, truncate(stdout, 512))
```
ahrtr: print the whole response? It won't too long.

ahrtr:
This PR high level looks good to me.

@frederiko @ivanvc do you have comment on the implementation? PTAL, thx

test/e2e/helpers_test.go
ahrtr: I mean removing the `truncate`

ballista01: done

ballista01:
> Please rebase & squash the commit, thx

on it

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,277 @@
...
+			if strings.HasPrefix(volume.PersistentVolumeClaim.ClaimName, expectedPVCPrefix) {
+				hasPVC = true
+				break
+			}
+		}
+	}
+
+	if !hasPVC {
+		t.Errorf("Pod %s does not use expected PVC with prefix %s", podName, expectedPVCPrefix)
+	}
```
ivanvc: I think this can be further simplified to:

```suggestion
	for _, volume := range pod.Spec.Volumes {
		if volume.PersistentVolumeClaim != nil {
			if strings.HasPrefix(volume.PersistentVolumeClaim.ClaimName, expectedPVCPrefix) {
				return
			}
		}
	}

	t.Errorf("Pod %s does not use expected PVC with prefix %s", podName, expectedPVCPrefix)
```

ivanvc (approved):
LGTM. Thanks, @ballista01.

test/e2e/helpers_test.go
ballista01: Applied, it's cleaner now.

ahrtr (approved):
LGTM

Thank you!

PR #194, opened 2025-08-22
Deprecate prepareOwnerReference in favor of controllerutils

frederiko:
This small refactor proposes deprecating prepareOwnerReferences in favor of controllerutils.SetControllerReference.


ahrtr (approved):
This is really a nice cleanup. 

It's be better to add an e2e test to verify that the statefulSets, headless service and configMap have the correct ownerReference set. It's OK to resolve it in a separate PR. Could you raise an issue to track it? thx

PR #227, opened 2025-10-23
Add `AutoProvider` certificate manager 

ArkaSaha30:
This PR will add `AutoProvider` certificate manager as a part of generating default self-signed certificates:
- interface functions 
- reconciler logic
- e2e tests

```
etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k get pods
NAME                   READY   STATUS    RESTARTS   AGE
etcdcluster-sample-0   1/1     Running   0          34m
etcdcluster-sample-1   1/1     Running   0          34m
etcdcluster-sample-2   1/1     Running   0          33m


etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k describe pod etcdcluster-sample-0  
Name:             etcdcluster-sample-0
Namespace:        default
Priority:         0
Service Account:  default
Node:             kind-control-plane/172.18.0.2
Start Time:       Mon, 27 Oct 2025 23:47:43 +0530
Labels:           app=etcdcluster-sample
                  apps.kubernetes.io/pod-index=0
                  controller=etcdcluster-sample
                  controller-revision-hash=etcdcluster-sample-754c4cfbcd
                  statefulset.kubernetes.io/pod-name=etcdcluster-sample-0
Annotations:      <none>
Status:           Running
IP:               10.244.0.9
IPs:
  IP:           10.244.0.9
Controlled By:  StatefulSet/etcdcluster-sample
Containers:
  etcd:
    Container ID:  containerd://faa47a05746374496ff1b3d0147a93efc4d72bc07572e68183df1df9decdcdbf
    Image:         gcr.io/etcd-development/etcd:v3.5.21
    Image ID:      gcr.io/etcd-development/etcd@sha256:fd158fbe55240e252947bbd2e8dddc217997ff43978071fac2bd202b6ad15c03
    Ports:         2379/TCP, 2380/TCP
    Host Ports:    0/TCP, 0/TCP
    Command:
      /usr/local/bin/etcd
    Args:
      --name=$(POD_NAME)
      --listen-peer-urls=http://0.0.0.0:2380
      --listen-client-urls=http://0.0.0.0:2379
      --initial-advertise-peer-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2380
      --advertise-client-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2379
    State:          Running
      Started:      Mon, 27 Oct 2025 23:47:52 +0530
    Ready:          True
    Restart Count:  0
    Environment Variables from:
      etcdcluster-sample-state  ConfigMap  Optional: false
    Environment:
      POD_NAME:       etcdcluster-sample-0 (v1:metadata.name)
      POD_NAMESPACE:  default (v1:metadata.namespace)
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-kwksc (ro)
Conditions:
  Type                        Status
  PodReadyToStartContainers   True 
  Initialized                 True 
  Ready                       True 
  ContainersReady             True 
  PodScheduled                True 
Volumes:
  server-secret:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  etcdcluster-sample-server-tls
    Optional:    false
  peer-secret:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  etcdcluster-sample-peer-tls
    Optional:    false
  kube-api-access-kwksc:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  34m   default-scheduler  Successfully assigned default/etcdcluster-sample-0 to kind-control-plane
  Normal  Pulling    34m   kubelet            Pulling image "gcr.io/etcd-development/etcd:v3.5.21"
  Normal  Pulled     34m   kubelet            Successfully pulled image "gcr.io/etcd-development/etcd:v3.5.21" in 8.646s (8.646s including waiting). Image size: 20653564 bytes.
  Normal  Created    34m   kubelet            Created container: etcd
  Normal  Started    34m   kubelet            Started container etcd

etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k get secrets
NAME                            TYPE                DATA   AGE
etcdcluster-sample-client-tls   kubernetes.io/tls   3      35m
etcdcluster-sample-peer-tls     kubernetes.io/tls   3      35m
etcdcluster-sample-server-tls   kubernetes.io/tls   3      35m

etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k describe secret etcdcluster-sample-client-tls
Name:         etcdcluster-sample-client-tls
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  kubernetes.io/tls

Data
====
ca.crt:   769 bytes
tls.crt:  769 bytes
tls.key:  365 bytes

etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k describe secret etcdcluster-sample-peer-tls  
Name:         etcdcluster-sample-peer-tls
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  kubernetes.io/tls

Data
====
tls.crt:  774 bytes
tls.key:  365 bytes
ca.crt:   774 bytes

etcd-operator on  auto-cert-provider [$?] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com 
❯ k describe secret etcdcluster-sample-server-tls
Name:         etcdcluster-sample-server-tls
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  kubernetes.io/tls

Data
====
ca.crt:   769 bytes
tls.crt:  769 bytes
tls.key:  365 bytes
```

### DIFF
--- a/test/e2e/auto_provider_test.go
+++ b/test/e2e/auto_provider_test.go
@@ -1,0 +1,237 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsV1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+	"go.etcd.io/etcd-operator/pkg/certificate"
+	"go.etcd.io/etcd-operator/pkg/certificate/auto"
+	interfaces "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+)
+
+const (
+	autoCertificateName      = "sample-cert"
+	autoCertificateNamespace = "default"
+	autoCertificateValidity  = auto.DefaultValidity
+)
+
+func TestAutoProvider(t *testing.T) {
+	feature := features.New("Auto Provider Certificate").WithLabel("app", string(certificate.Auto))
+
+	cmConfig := &interfaces.Config{
+		CommonName:       autoCertificateName,
+		ValidityDuration: autoCertificateValidity,
+	}
+
+	feature.Setup(
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			_ = appsv1.AddToScheme(client.Resources().GetScheme())
+			_ = corev1.AddToScheme(client.Resources().GetScheme())
+			_ = certv1.AddToScheme(client.Resources().GetScheme())
+			_ = apiextensionsV1.AddToScheme(client.Resources().GetScheme())
+
+			return ctx
+		})
+
+	feature.Assess("Ensure certificate",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			acProvider := auto.New(client.Resources().GetControllerRuntimeClient())
+			err := acProvider.EnsureCertificateSecret(ctx, autoCertificateName, autoCertificateNamespace, cmConfig)
+			if err != nil {
+				t.Fatalf("Auto Provider Certificate could not be created: %v", err)
+			}
+			return ctx
+		})
+
+	feature.Assess("Validate certificate secret",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			acProvider := auto.New(client.Resources().GetControllerRuntimeClient())
+			err := acProvider.ValidateCertificateSecret(ctx, autoCertificateName, autoCertificateNamespace, cmConfig)
+			if err != nil {
+				t.Fatalf("Failed to validate Auto Provider Certificate secret: %v", err)
+			}
+			return ctx
+		})
+
+	feature.Assess("Get certificate config",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			acProvider := auto.New(client.Resources().GetControllerRuntimeClient())
+			config, err := acProvider.GetCertificateConfig(ctx, cmCertificateName, cmCertificateNamespace)
+			if err != nil {
+				t.Fatalf("Auto Certificate not found: %v", err)
+			}
+			if !reflect.DeepEqual(config, cmConfig) {
+				t.Fatalf("Auto Certificate config does not match with the given config")
+			}
+			return ctx
+		})
+
+	feature.Assess("Delete certificate secret",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			acProvider := auto.New(client.Resources().GetControllerRuntimeClient())
+			err := acProvider.DeleteCertificateSecret(ctx, autoCertificateName, autoCertificateNamespace)
+			if err != nil {
+				t.Fatalf("Failed to delete Certificate secret: %v", err)
+			}
+			return ctx
+		})
+
+	feature.Assess("Verify Delete certificate",
+		func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			client := cfg.Client()
+			acProvider := auto.New(client.Resources().GetControllerRuntimeClient())
+			_, err := acProvider.GetCertificateConfig(ctx, autoCertificateName, autoCertificateNamespace)
+			if err == nil {
+				t.Fatalf("Auto Provider Certificate found, deletion failed: %v", err)
+			}
+			return ctx
+		})
+
+	_ = testEnv.Test(t, feature.Feature())
+}
+
+func TestClusterAutoCertCreation(t *testing.T) {
+	feature := features.New("cluster-auto-cert-creation")
+
+	const etcdClusterName = "etcd-cluster-auto-cert"
+	const size = 3
+
+	etcdCluster := &ecv1alpha1.EtcdCluster{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "operator.etcd.io/v1alpha1",
+			Kind:       "EtcdCluster",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdClusterName,
+			Namespace: namespace,
+		},
+		Spec: ecv1alpha1.EtcdClusterSpec{
+			Size:    size,
+			Version: etcdVersion,
+			TLS: &ecv1alpha1.TLSCertificate{
+				Provider: string(certificate.Auto),
+				ProviderCfg: ecv1alpha1.ProviderConfig{
+					AutoCfg: &ecv1alpha1.ProviderAutoConfig{
+						CommonConfig: ecv1alpha1.CommonConfig{
+							CommonName:       "etcd-operator-system",
+							ValidityDuration: "8760h",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	feature.Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+		client := cfg.Client()
+		_ = appsv1.AddToScheme(client.Resources().GetScheme())
+		_ = corev1.AddToScheme(client.Resources().GetScheme())
+		_ = certv1.AddToScheme(client.Resources().GetScheme())
+		_ = apiextensionsV1.AddToScheme(client.Resources().GetScheme())
+
+		// create etcd cluster
+		if err := client.Resources().Create(ctx, etcdCluster); err != nil {
+			t.Fatalf("unable to create etcd cluster: %s", err)
+		}
+
+		// get etcd cluster object
+		var ec ecv1alpha1.EtcdCluster
+		if err := client.Resources().Get(ctx, etcdClusterName, namespace, &ec); err != nil {
+			t.Fatalf("unable to fetch etcd cluster: %s", err)
+		}
+
+		return ctx
+	})
+
+	feature.Assess("Check certificate secrets exist",
+		func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			client := c.Client()
+			// checks if corresponding client, server, peer secrets are created in the respective namespace
+			if err := wait.For(
+				func(context.Context) (bool, error) {
+					return validateSecretExists(ctx, client, etcdClusterName, namespace, "secret")
+				},
+				wait.WithTimeout(3*time.Minute),
+				wait.WithInterval(10*time.Second),
+			); err != nil {
+				t.Fatalf("timed out waiting for certificate: %s", err)
+			}
+			return ctx
+		},
+	)
+
+	feature.Assess("Verify Data Operations",
+		func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			// verify etcdCluster is accessible via client certificate with put and get
+			verifyDataOperations(t, c, etcdClusterName, "test-key", "test-value")
+			return ctx
+		},
+	)
+
+	_ = testEnv.Test(t, feature.Feature())
+}
+
+func validateSecretExists(ctx context.Context, client klient.Client,
+	etcdClusterName, etcdClusterNamespace, resourceType string) (bool, error) {
+	clientCertName := fmt.Sprintf("%s-client-tls", etcdClusterName)
+	serverCertName := fmt.Sprintf("%s-server-tls", etcdClusterName)
+	peerCertName := fmt.Sprintf("%s-peer-tls", etcdClusterName)
+
+	var obj any
+
+	switch resourceType {
+	case "secret":
+		var secretObj corev1.Secret
+		obj = &secretObj
+	default:
+		return false, fmt.Errorf("invalid resource type: %v", resourceType)
+	}
+
+	runtimeObj, err := obj.(k8s.Object)
+	if !err {
+		return false, fmt.Errorf("object does not implement runtime.Object: %T", obj)
+	}
+
+	if err := client.Resources().Get(ctx, clientCertName, etcdClusterNamespace, runtimeObj); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get Client %s: %v", resourceType, err)
+	}
+
+	if err := client.Resources().Get(ctx, serverCertName, etcdClusterNamespace, runtimeObj); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get Server %s: %v", resourceType, err)
+	}
+
+	if err := client.Resources().Get(ctx, peerCertName, etcdClusterNamespace, runtimeObj); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get Peer %s: %v", resourceType, err)
+	}
+	return true, nil
+}


config/crd/bases/operator.etcd.io_etcdclusters.yaml
```diff
@@ -149,7 +149,8 @@ spec:
                           validityDuration:
                             description: |-
                               ValidityDuration is the expected duration until which the certificate will be valid,
-                              expects in human-readable duration: 100d12h, if empty defaults to 90d
+                              expects in human-readable duration: 100d12h, if empty defaults to 90d for cert-manager
+                              and 365d for auto as per: https://github.com/etcd-io/etcd/blob/main/client/pkg/transport/listener.go#L275
```
ivanvc: This link will go out of date quickly. Should we either:
1. Replace with a permalink (commit sha).
2. Replace with a link to some document.
3. Replace with the actual reason why it's like this.

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,295 @@
...
+			return err
+		}
+	}
+
+	err = ac.ValidateCertificateSecret(ctx, secretName, namespace, cfg)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return err
+		} else {
+			return fmt.Errorf("invalid certificate secret: %s present in namespace: %s, please delete and try again.\nError: %s",
```
ivanvc: Suggestion to wrap the underlying error:

```suggestion
			return fmt.Errorf("invalid certificate secret: %s present in namespace: %s, please delete and try again.\nError: %w",
```

internal/controller/utils.go
```diff
@@ -600,12 +600,51 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
...
+			// TODO: instead of error, set default autoConfig
+			return err
+		}
+		autoConfig := createAutoCertificateConfig(ec)
+		createCertErr := autoCert.EnsureCertificateSecret(ctx, certName, ec.Namespace, autoConfig)
+		if createCertErr != nil {
+			log.Printf("Error creating certificate: %s", createCertErr)
+		}
+		return nil
+	}
```
```
providerName := ec.Spec.TLS.Provider
if providerName == "" {
    providerName = Certificate.Auto
}

```

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (ac *AutoCertProvider) GetCertificateConfig(ctx context.Context,
+	secretName, namespace string) (*interfaces.Config, error) {
+	var autoCertSecret corev1.Secret
+	err := ac.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, &autoCertSecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get certificate: %w", err)
+	}
+
+	return &interfaces.Config{}, nil
```
ahrtr: We should get the info from the certificate included in the secret.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -98,7 +98,8 @@ type CommonConfig struct {
 	AltNames AltNames `json:"altNames,omitempty"`
 
 	// ValidityDuration is the expected duration until which the certificate will be valid,
-	// expects in human-readable duration: 100d12h, if empty defaults to 90d
+	// expects in human-readable duration: 100d12h, if empty defaults to 90d for cert-manager
+	// and 365d for auto as per: https://github.com/etcd-io/etcd/blob/main/client/pkg/transport/listener.go#L275
```
ivanvc: This link will quickly go out of date. Should we either:

1.   Replace with a permalink (commit sha).
2.  Replace with a link to some document.
3.  Replace with the actual reason why it's like this.

internal/controller/utils.go
```diff
@@ -600,12 +600,51 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
...
+	duration, err := time.ParseDuration(autoConfig.ValidityDuration)
+	if err != nil {
+		log.Printf("Failed to parse ValidityDuration: %s", err)
+	}
+
+	var getAltNames certInterface.AltNames
+	if autoConfig.AltNames.DNSNames != nil {
+		getAltNames = certInterface.AltNames{
+			DNSNames: autoConfig.AltNames.DNSNames,
+			IPs:      make([]net.IP, len(autoConfig.AltNames.DNSNames)),
```
ivanvc, line 613: What's the purpose of this? Is the plan to fill it later? I can't find it in the current code.

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,295 @@
...
+
+	interfaces "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+	"go.etcd.io/etcd/client/pkg/v3/transport"
+)
+
+const (
+	DefaultValidity = 365 * 24 * time.Hour
+)
+
+type AutoCertProvider struct {
```
ivanvc: I think this should be named only "Provider", so the import path is `auto.Provider`. And ideally CertManager's should be `cert_manager.Provider`. Removes the stutter.

test/e2e/auto_provider_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+	"go.etcd.io/etcd-operator/pkg/certificate/auto"
+	interfaces "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+)
+
+const (
+	autoCertificateName      = "sample-cert"
+	autoCertificateNamespace = "default"
+	autoCertificateValidity  = 365 * 24 * time.Hour
```
ivanvc: Why not use `auto.DefaultValidity`?

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	if checkKeyPairErr := checkKeyPair(parseCert, privateKey); checkKeyPairErr != nil {
+		return fmt.Errorf("private key does not match certificate: %w", checkKeyPairErr)
+	}
+
+	return nil
+}
+
+func (ac *AutoCertProvider) DeleteCertificateSecret(ctx context.Context, secretName, namespace string) error {
```
ivanvc: According to the documentation, we should also clean up the certificate. But I don't see any logic regarding that. Is this going to be done in the controller? Or do you think this will be added later?

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,295 @@
...
+
+	return nil
+}
+
+// createNewSecret generates or updates a Kubernetes TLS Secret
+// with a self-signed certificate in the specified namespace.
+// DNSNames and IPAddresses if not user-defined, will be set to default value in runtime:
+// fmt.Sprintf("%s-%d.%s.%s.svc.cluster.local", ec.Name, index, ec.Name, ec.Namespace)
+// minimum validityDuration is 365 days
+// as per: https://github.com/etcd-io/etcd/blob/main/client/pkg/transport/listener.go#L275
```
ivanvc: Same comment here, this link may become invalid soon.

pkg/certificate/auto/provider.go
ArkaSaha30: In the case of auto provider, we create a secret that has the certificate inside it. There is no certificate resource created.
The certificate resource will be created only in the case of cert-manager.

api/v1alpha1/etcdcluster_types.go
ArkaSaha30: Sure, updating it with the permalink(commit sha)

internal/controller/utils.go
ArkaSaha30: So we are preparing the certificate config here from the user-defined config for etcd-operator. This config will be passed to the `createCertificate` function where based on the provider `EnsureCertificateSecret`with the above mentioned config parameter will be passed to create the certificateSecret

pkg/certificate/auto/provider.go
ArkaSaha30: Sure, I am updating it for Auto provider here. I will update the same for cert-manager in a separate PR.

pkg/certificate/auto/provider.go
ivanvc: But then, we should delete the secret when cleaning up, right?

internal/controller/utils.go
ivanvc: But, this is not being used anywhere. I can see how it is filled for `CertManager`: 
https://github.com/etcd-io/etcd-operator/blob/c4d7a626747fa372168d40bfcb73612477e1f68c/pkg/certificate/cert_manager/provider.go#L217

And used: https://github.com/etcd-io/etcd-operator/blob/c4d7a626747fa372168d40bfcb73612477e1f68c/pkg/certificate/cert_manager/provider.go#L330

Unless I'm not following through, as this slice is always empty.

pkg/certificate/auto/provider.go
ArkaSaha30: > But then, we should delete the secret when cleaning up, right?

Yes correct. I will add the reconciliation logic for it, thanks

ArkaSaha30: Added it as a part of patchCertificateSecret func: https://github.com/ArkaSaha30/etcd-operator/blob/auto-cert-provider/internal/controller/utils.go#L731-L747 

Manually validated that deleting the etcdCluster objects deletes the secret as well.

internal/controller/utils.go
ivanvc: But, this is not being used anywhere. I can see how it is filled for `CertManager`: 
https://github.com/etcd-io/etcd-operator/blob/c4d7a626747fa372168d40bfcb73612477e1f68c/pkg/certificate/cert_manager/provider.go#L217

And used: https://github.com/etcd-io/etcd-operator/blob/c4d7a626747fa372168d40bfcb73612477e1f68c/pkg/certificate/cert_manager/provider.go#L330

Unless I'm not following through, as this slice is always empty.

pkg/certificate/auto/provider.go
ivanvc: Just triple checking :smile:. If we delete the `EtcdCluster` CR, and it had a certificate through `AutoProvider`, will the certificate secret get deleted? Or would it be left in the cluster?

ArkaSaha30: >  If we delete the EtcdCluster CR, and it had a certificate through AutoProvider, will the certificate secret get deleted

Yes, I tested it locally and deleting the etcdCluster CR will remove the certificate secrets.

internal/controller/utils.go
ArkaSaha30: I am using it here for creating the self-signed certificate: https://github.com/ArkaSaha30/etcd-operator/blob/auto-cert-provider/pkg/certificate/auto/provider.go#L224-L232

ArkaSaha30: But IPs are not being used, should it come as a part of the hosts?

pkg/certificate/auto/provider.go
ivanvc: Thanks for confirming.

internal/controller/utils.go
ivanvc: That's the DNS names, not the IP addresses. This slice is kept empty with a capacity of `len(autoConfig.AltNames.DNSNames)`. @ahrtr, do we need the `IPs`?

ahrtr: I think we should populate the IPs with whatever the IPs passed by users (otherwise we should remove IPAddress from the user-facing config), although usually in practice IPAddresses aren't used, because POD IPs are dynamic.

Please update the `createCMCertificateConfig` to populate the IPs with user provided values as well. It can be resolved in a separate PR.

We can think about to refactor the user interface or implementation in future when we get more feedback from real users.

ivanvc:
@ArkaSaha30, where do we stand in this pull request? Is this ready for another review?

internal/controller/utils.go
ivanvc: > I think we should populate the IPs with whatever the IPs passed by users (otherwise we should remove IPAddress from the user-facing config), although usually in practice IPAddresses aren't used, because POD IPs are dynamic.

I think we should delete the `IPAddressess` property because, as you said, this will get invalidated with pod rotations.

pkg/certificate/auto/provider.go
ArkaSaha30: Hi @ahrtr , I explored getting the original config from the secret itself but it will be a complicated one since we are manipulating the config and passing it to `transport.SelfCert()` which in turn creats the certificate.
Another workaround is to retrieve the certificateConfig from the etcdCluster object via the `certificateSecret.ownerRef.Name`, but it will cause a circular dependency

What can be the suggested solution here?
cc @ivanvc

internal/controller/utils.go
ahrtr: > I think we should delete the `IPAddressess` property because, as you said, this will get invalidated with pod rotations.

I suggest that we follow comment to proceed for now. For any user facing API change, we should get more feedback. Let's discuss it separately.

ahrtr: > I suggest that we follow comment to proceed for now

I mean https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2480818598

pkg/certificate/auto/provider.go
ahrtr: > but it will be a complicated one

Where is the complication?  I don't see a PoC so far, nor a reasonable explanation.

The easiest solution is to just save the config passed to `EnsureCertificateSecret`, and return it in `GetCertificateConfig`. The problem is that it isn't the final real configuration that we used to create the certificate.

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,297 @@
...
+	if cfg != nil {
+		if cfg.CommonName != "" {
+			hosts = append(hosts, cfg.CommonName)
+		}
+		if len(cfg.AltNames.DNSNames) > 0 {
+			hosts = append(hosts, cfg.AltNames.DNSNames...)
+		}
+	}
+
+	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, uint(validity/DefaultValidity))
```
neolit123, line 251: if you decide to use the fake host ports as we discussed during the meeting of 25 Nov, due to the bug in SelfCert not working for hosts without port, then remember to use net.JoinHostPort for adding a port. the function is IPv4/6 compatible.

ArkaSaha30: Thank you, I have updated the code.

pkg/certificate/auto/provider.go
ArkaSaha30: Sure, I have updated the `GetCertificateConfig ` to return the saved user-defined config

internal/controller/utils.go
```diff
@@ -600,13 +600,42 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
 }
 
 func createAutoCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config {
-	// TODO
-	config := &certInterface.Config{}
+	autoConfig := ec.Spec.TLS.ProviderCfg.AutoCfg
+	duration, err := time.ParseDuration(autoConfig.ValidityDuration)
+	if err != nil {
+		log.Printf("Failed to parse ValidityDuration: %s", err)
```
neolit123, line 606: should it return an error here?

internal/controller/utils.go
```diff
@@ -600,13 +600,42 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
...
+	}
+
+	var getAltNames certInterface.AltNames
```
neolit123: maybe just call it `altNames`

internal/controller/utils.go
```diff
@@ -600,13 +600,42 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
...
+		Organization:     autoConfig.Organization,
+		ValidityDuration: duration,
+		AltNames:         getAltNames,
+	}
 	return config
 }
 
 func createCertificate(ec *ecv1alpha1.EtcdCluster, ctx context.Context, c client.Client, certName string) error {
-	cert, certErr := certificate.NewProvider(certificate.ProviderType(ec.Spec.TLS.Provider), c)
+	// tls field is present but spec is empty
```
neolit123:
```suggestion
	// The TLS field is present but spec is empty
```

internal/controller/utils.go
```diff
@@ -643,40 +672,52 @@ func createCertificate(ec *ecv1alpha1.EtcdCluster, ctx context.Context, c client
 	return nil
 }
 
-func createClientCertificate(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client) error {
+func createClientCertificate(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client, scheme *runtime.Scheme) error {
 	certName := getClientCertName(ec.Name)
 	createClientCertErr := createCertificate(ec, ctx, c, certName)
+	patchCertErr := patchCertificateSecret(ctx, ec, c, scheme, certName)
```
neolit123: just use a single err and check it on every stage.

internal/controller/utils.go
```diff
@@ -643,40 +672,52 @@ func createCertificate(ec *ecv1alpha1.EtcdCluster, ctx context.Context, c client
...
-func createServerCertificate(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client) error {
+func createServerCertificate(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client, scheme *runtime.Scheme) error {
 	serverCertName := getServerCertName(ec.Name)
 	createServerCertErr := createCertificate(ec, ctx, c, serverCertName)
 	if createServerCertErr != nil {
 		return createServerCertErr
 	}
+	patchCertErr := patchCertificateSecret(ctx, ec, c, scheme, serverCertName)
+	if patchCertErr != nil {
+		return fmt.Errorf("patching certificate secret: %s with ownerReference failed: %w", serverCertName, patchCertErr)
```
neolit123: same comments as above

internal/controller/utils.go
```diff
@@ -643,40 +672,52 @@ func createCertificate(ec *ecv1alpha1.EtcdCluster, ctx context.Context, c client
...
+func createPeerCertificate(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client, scheme *runtime.Scheme) error {
 	peerCertName := getPeerCertName(ec.Name)
 	createPeerCertErr := createCertificate(ec, ctx, c, peerCertName)
 	if createPeerCertErr != nil {
 		return createPeerCertErr
 	}
+	patchCertErr := patchCertificateSecret(ctx, ec, c, scheme, peerCertName)
+	if patchCertErr != nil {
+		return fmt.Errorf("patching certificate secret: %s with ownerReference failed: %w", peerCertName, patchCertErr)
+	}
```
neolit123, line 710: same

internal/controller/utils.go
```diff
@@ -643,40 +672,52 @@ func createCertificate(ec *ecv1alpha1.EtcdCluster, ctx context.Context, c client
...
 	if ec.Spec.TLS != nil {
-		createServerCertErr := createServerCertificate(ctx, ec, c)
+		createServerCertErr := createServerCertificate(ctx, ec, c, scheme)
 		if createServerCertErr != nil {
 			err = createServerCertErr
 			logger.Error(createServerCertErr, "Error creating Server Certificate")
 
 		}
-		createPeerCertErr := createPeerCertificate(ctx, ec, c)
+		createPeerCertErr := createPeerCertificate(ctx, ec, c, scheme)
```
neolit123: is there a reason why it both logs and returns the err?
seems like it can just use the same err var for all stages and check them after another.

internal/controller/utils.go
```diff
@@ -686,3 +727,21 @@ func applyEtcdMemberCerts(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c cli
 	}
 	return err
 }
+
+func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client, scheme *runtime.Scheme, certSecretName string) error {
+	getCertSecret := &corev1.Secret{}
+	getErr := c.Get(ctx, client.ObjectKey{Name: certSecretName, Namespace: ec.Namespace}, getCertSecret)
```
neolit123: same here just call it err

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+func New(c client.Client) interfaces.Provider {
+	return &Provider{
+		c,
+	}
+}
+
+var userAutoConfig *interfaces.Config
+
+func (ac *Provider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string,
+	cfg *interfaces.Config) error {
```
neolit123, line 49: keep it on the prev line

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+	// save the user defined AutoConfig so that it can be returned from GetCertificateConfig
```
neolit123:
```suggestion
	// Save the user defined AutoConfig so that it can be returned from GetCertificateConfig
```

internal/controller/utils.go
```diff
@@ -686,3 +727,21 @@ func applyEtcdMemberCerts(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c cli
...
+	if getErr != nil {
+		return getErr
+	}
+
+	log.Printf("Setting ownerReference for certificate secret: %s", certSecretName)
+	if err := controllerutil.SetControllerReference(ec, getCertSecret, scheme); err != nil {
+		return err
+	}
+	if updateErr := c.Update(ctx, getCertSecret); updateErr != nil {
```
neolit123: err

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+		if len(cfg.AltNames.DNSNames) > 0 {
+			for _, dnsName := range cfg.AltNames.DNSNames {
+				hostPort := net.JoinHostPort(dnsName, "5500")
+				hosts = append(hosts, hostPort)
+			}
+		}
+	}
+
+	log.Printf("hosts: %v", hosts)
```
neolit123: maybe add a more detailed log msg, like "calling SelfCert with hosts: ...`

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	keyPEM, err := os.ReadFile(keyPath)
+	if err != nil {
+		return fmt.Errorf("failed to read generated key file %s: %w", keyPath, err)
+	}
+
+	caPEM, err := os.ReadFile(caPath)
+	if err != nil || len(caPEM) == 0 {
+		// use certPEM when CA file is not found or empty
```
neolit123:
```suggestion
		// Use certPEM when CA file is not found or empty
```

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+		caPEM = certPEM
+	}
+
+	// Validate cert and key pair
+	if _, err := tls.X509KeyPair(certPEM, keyPEM); err != nil {
+		return fmt.Errorf("generated keypair invalid: %w", err)
```
neolit123:
```suggestion
		return fmt.Errorf("generated keypair is invalid: %w", err)
```

neolit123 (commented):
thanks for the updates
generally LGTM, just a few minor comments from me

internal/controller/etcdcluster_controller.go
```diff
@@ -117,7 +117,7 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 
 	// Ensure the operator has TLS credentials when the cluster requests TLS.
 	if ec.Spec.TLS != nil {
-		if err := createClientCertificate(ctx, ec, r.Client); err != nil {
+		if err := createClientCertificate(ctx, ec, r.Client, r.Scheme); err != nil {
```
palnabarun: `r.Client` also has the Scheme embedded. Can't that be used instead of passing Scheme through the call chain?

internal/controller/utils.go
palnabarun: also, the Get call and the error check can be folded to the same block.

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+
+func New(c client.Client) interfaces.Provider {
+	return &Provider{
+		c,
+	}
+}
+
+var userAutoConfig *interfaces.Config
+
+func (ac *Provider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string,
```
palnabarun, line 48: Whenever the "name, namespace" pattern is used, an object key can be used. This ensures consistency. See github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/example_test.go for example.

```suggestion
func (ac *Provider) EnsureCertificateSecret(ctx context.Context, secretKey ctrlruntimeclient.ObjectKey
```

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+	} else {
+		return err
+	}
+
+	certificateData, exists := secret.Data["tls.crt"]
+	if !exists {
+		return interfaces.ErrTLSCert
+	}
+
+	decodeCertificatePem, _ := pem.Decode(certificateData)
```
palnabarun, line 107: Do we not need an err check?

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+
+func (ac *Provider) DeleteCertificateSecret(ctx context.Context, secretName, namespace string) error {
+	var secret corev1.Secret
+	err := ac.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, &secret)
+	if k8serrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
palnabarun:
```suggestion
	if err := ac.Get(ctx, client.ObjectKey{Name: secretName, Namespace: namespace}, &secret); err != nil {
	    if k8serrors.IsNotFound(err) {
		    return nil
	    }
		return err
	}
```

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+	if userAutoConfig != nil {
+		return userAutoConfig, nil
+	}
+
+	return nil, fmt.Errorf("failed to get user-defined autoConfig")
+}
+
+// parsePrivateKey parses the private key from the PEM-encoded data.
+func parsePrivateKey(privateKeyData []byte) (crypto.PrivateKey, error) {
+	block, _ := pem.Decode(privateKeyData)
```
palnabarun, line 170: same question as above

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,317 @@
...
+		return nil, errors.New("failed to decode private key: invalid PEM")
+	}
+
+	// Parse the private key from the PEM block
+	privateKey, err := x509.ParseECPrivateKey(block.Bytes)
+	if err != nil {
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse private key: %w", err)
+		}
+	}
```
palnabarun, line 179:
```suggestion
	if err != nil {
		return nil, fmt.Errorf("failed to parse private key: %w", err)
	}
```

ahrtr:
@ArkaSaha30 pls feedback after you resolve all the review comments, thx

pkg/certificate/auto/provider.go
ArkaSaha30: There is a linter restriction of 120 chars which is why I had to do the line break:

```
etcd-operator on  auto-cert-provider [$!?] via 🐳 colima via 🐹 v1.25.4 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com took 7s 
❯ make lint
/Users/arkas1/Documents/GitHub/upstream/etcd-operator/bin/golangci-lint run
pkg/certificate/auto/provider.go:48:1: The line is 126 characters long, which exceeds the maximum of 120 characters. (lll)
func (ac *Provider) EnsureCertificateSecret(ctx context.Context, secretName, namespace string, cfg *interfaces.Config) error {
^
1 issues:
* lll: 1
make: *** [lint] Error 1
```

internal/controller/utils.go
ArkaSaha30: Agree, there is no need to log the error again, as well as return the error. Fixed it.

pkg/certificate/auto/provider.go
ArkaSaha30: Since this is an interface change, I will change and refactor it for both certManager and auto certificate provider in a separate follow-up PR

pkg/certificate/auto/provider.go
ArkaSaha30: We are already checking the error in the next line. 
`pem.Decode` returns `(p *Block, rest []byte)` - pointer to decoded PEM block and the remaining bytes after the block. 
`if decodeCertificatePem == nil` checks if the decoding failed and returns error `ErrDecodeCert`

pkg/certificate/auto/provider.go
ArkaSaha30: Same `pem.Decode` error handling style: https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2596521671

internal/controller/etcdcluster_controller.go
ArkaSaha30: Agree, we can use `r.Client.Scheme()` rather than passing extra param `r.Scheme`. Updated it.

internal/controller/utils.go
ArkaSaha30: I am currently handling the scenario in `auto/provider.go` - https://github.com/ArkaSaha30/etcd-operator/blob/auto-cert-provider/pkg/certificate/auto/provider.go#L217-L220
Currently, it will log the error but fall back to the default duration of 1yr and let the user bring up the etcdCluster.
Please let me know if this is not the desired behaviour, and I will update it to throw an error.

pkg/certificate/auto/provider.go
ahrtr: Good point, Pls raise a separate ticket to track this change

internal/controller/utils.go
ahrtr: I think we should return an error here, we only use the default value if not provided. Pls raise a separate ticket to track this.

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,314 @@
...
+
+var _ interfaces.Provider = (*Provider)(nil)
+
+func New(c client.Client) interfaces.Provider {
+	return &Provider{
+		c,
+	}
+}
+
+var userAutoConfig *interfaces.Config
```
ahrtr, line 46: Why define a global variable? Please add it as a field in Provider struct (line 34 above).

ahrtr: pls also resolve this in a followup PR, thx

pkg/certificate/auto/provider.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	parseCert, err := x509.ParseCertificate(decodeCertificatePem.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse certificate: %w", err)
+	}
+
+	if parseCert.NotAfter.Before(time.Now()) {
+		return interfaces.ErrCertExpired
+	}
```
ivanvc, line 124: Would it be a good idea also to check `parseCert.NotBefore`?

ArkaSaha30: Yes! Adding it in the refactor issue: https://github.com/etcd-io/etcd-operator/issues/251

Issue #182, opened 2025-07-22
Determine final list of fields for 0.2 Status


jberkus:
Per #136 we need to decide the final list of fields for cluster status.  Here's the current proposal from @ballista01 :

```
status:
  observedGeneration: 5                    # NEW: Top-level field indicating the latest .metadata.generation processed by the controller
  currentReplicas: 3                       # Unchanged: Matches the StatefulSet replica count
  currentVersion: 3.6.0-rc.3               # Unchanged: Represents the running etcd version of the cluster
  leaderId: 975eab669fe7b560               # Unchanged: Leader member ID for status observability
  memberCount: 3                           # Unchanged: Total number of etcd members
  readyReplicas: 3

  # lastDefragTime could be optionally added if controller manages defragmentation
  # lastDefragTime: "2025-07-20T12:00:00Z"   # OPTIONAL: Cluster-level defragmentation completion time; not included in this PR

  members:
    - id: 22f4b6fa7639917a
      name: test-cluster-2
      version: 3.6.0-rc.3
      isHealthy: true
      isLearner: false                     # NEW: Added to support learner promotion logic and troubleshooting
      # REMOVED: dbSize/dbSizeInUse → considered too dynamic; suggested to be exposed via /metrics
      # clientURL and peerURL omitted here for minimal diagnostic view, can be re-added under a feature gate

    - id: 52117ab448fed65d
      name: test-cluster-1
      version: 3.6.0-rc.3
      isHealthy: true
      isLearner: false

    - id: 4f8c9e671cbd2233
      name: test-cluster-0
      version: 3.6.0-rc.3
      isHealthy: true
      isLearner: false

  conditions:
    - type: Available
      status: "True"
      reason: ClusterReady
      message: Cluster is fully available and healthy
      observedGeneration: 5
      lastTransitionTime: "2025-07-02T18:52:06Z"
      # Unchanged: Matches K8s convention for status.conditions

    - type: Progressing
      status: "False"
      reason: ReconcileSuccess
      message: Cluster reconciled to desired state
      observedGeneration: 5
      lastTransitionTime: "2025-07-02T18:52:06Z"
      # Unchanged: Indicates that spec and observed state are aligned

    - type: Degraded
      status: "False"
      reason: ClusterReady
      message: Cluster is healthy
      observedGeneration: 5
      lastTransitionTime: "2025-07-02T18:50:57Z"
      # Unchanged: Used to signal partial failures if any member is unhealthy

[...]

PR #229, opened 2025-10-23
api: add EtcdCluster status fields and CRD schema

ballista01:
Extracted from https://github.com/etcd-io/etcd-operator/pull/136 for easier review per https://github.com/etcd-io/etcd-operator/pull/136#issuecomment-3438387686. API Changes only, no logic.


ivanvc:
Link to #182

api/v1alpha1/etcdcluster_types.go
```diff
@@ -130,6 +130,82 @@ type ProviderCertManagerConfig struct {
...
+	MemberCount int32 `json:"memberCount,omitempty"`
+
+	// CurrentVersion is the observed etcd version of the cluster.
+	// This is typically derived from the version of the healthy leader or a consensus among healthy members.
+	// +optional
+	CurrentVersion string `json:"currentVersion,omitempty"`
+
+	// LeaderId is the hex-encoded ID of the current etcd cluster leader, if one exists and is known.
+	// +optional
+	LeaderId string `json:"leaderId,omitempty"`
```
ivanvc: Can we rename this to `LeaderID` to match the Go standard?

```suggestion
	// LeaderID is the hex-encoded ID of the current etcd cluster leader, if one exists and is known.
	// +optional
	LeaderID string `json:"leaderID,omitempty"`
```

ballista01: Yeah, it also matches what's in `etcdutils.go`. Thanks for the suggestion! Applied.

Issue #232, opened 2025-10-28
Cert-manager scheme registration error in etcd-operator-controller-manager


ArkaSaha30:
### Problem
In the current implementation of cert-manager provider in etcd-operator, the etcd-operator-controller-manager expects cert-manager to be installed in the desired cluster as a prerequisite. 
This prerequisite should not exist since we want to give users the flexibility to choose their own certificate manager, currently limited to: cert-manager provider and auto provider. 

### Cause
Introduced as a part of commit: https://github.com/etcd-io/etcd-operator/commit/b97e11f5ce97ac5854cf8cdeec932f2248cacb27 
Due to cert-manager scheme registration here: https://github.com/etcd-io/etcd-operator/blob/main/cmd/main.go#L52

### Error Logs
In a vanilla cluster not having cert-manager pre-installed, the etcd-operator runs into the following error, causing CrashLoopBackOff.:
```
etcd-operator on  auto-cert-provider [$!] via 🐳 colima via 🐹 v1.25.3 on ☁️  (us-east-1) on ☁️  arkasaha30@gmail.com took 4s 
❯ k logs -f etcd-operator-controller-manager-57f9c8d96c-h5wkd -n etcd-operator-system     
2025-10-28T17:26:53Z	INFO	setup	starting manager
2025-10-28T17:26:53Z	INFO	controller-runtime.metrics	Starting metrics server
2025-10-28T17:26:53Z	INFO	setup	disabling http/2
2025-10-28T17:26:53Z	INFO	starting server	{"name": "health probe", "addr": "[::]:8081"}
I1028 17:26:53.655650       1 leaderelection.go:257] attempting to acquire leader lease etcd-operator-system/cc4a0f4b.etcd.io...
I1028 17:26:53.667873       1 leaderelection.go:271] successfully acquired lease etcd-operator-system/cc4a0f4b.etcd.io
2025-10-28T17:26:53Z	INFO	Starting EventSource	{"controller": "etcdcluster", "controllerGroup": "operator.etcd.io", "controllerKind": "EtcdCluster", "source": "kind source: *v1.Certificate"}
[...]

PR #234, opened 2025-10-30
Fix cert-manager scheme registration

ArkaSaha30:
This PR will:
- fix default always install cert-manager in e2e test to install when cert-manager specific test is called optionally.
- fix default scheme registration of cert-manager

Fixes: https://github.com/etcd-io/etcd-operator/issues/232

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -334,14 +334,34 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 	return ctrl.Result{}, nil
 }
 
+// isCertManagerCRDPresent checks if cert-manager CRDs are installed in the cluster
+func isCertManagerCRDPresent(mgr ctrl.Manager) bool {
+	gvk := certv1.SchemeGroupVersion.WithKind("Certificate")
+	_, err := mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+	return err == nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *EtcdClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.Recorder = mgr.GetEventRecorder("etcdcluster-controller")
-	return ctrl.NewControllerManagedBy(mgr).
+	setupLog := ctrl.Log.WithName("setup")
+
+	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&ecv1alpha1.EtcdCluster{}).
 		Owns(&appsv1.StatefulSet{}).
 		Owns(&corev1.Service{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&certv1.Certificate{}).
-		Complete(r)
+		Owns(&corev1.ConfigMap{})
+
+	// Conditionally watch cert-manager Certificate resources if CRDs are installed
+	// This allows the controller to react to Certificate status changes when using cert-manager provider
+	if isCertManagerCRDPresent(mgr) {
+		// cert-manager CRDs are installed, add Certificate watch
+		builder = builder.Owns(&certv1.Certificate{})
+		setupLog.Info("cert-manager CRDs detected, enabling Certificate watches")
+	} else {
+		// cert-manager CRDs not installed, skip Certificate watch
+		setupLog.Info("cert-manager CRDs not detected, only auto provider will be available. Restart the controller after cert-manager CRDs are installed")
+	}
+
+	return builder.Complete(r)
 }
--- a/test/e2e/cert_manager_test.go
+++ b/test/e2e/cert_manager_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"log"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


config/crd/bases/operator.etcd.io_etcdclusters.yaml
```diff
@@ -296,8 +296,8 @@ spec:
                   CurrentVersion is the observed etcd version of the cluster.
                   This is typically derived from the version of the healthy leader or a consensus among healthy members.
                 type: string
-              leaderId:
-                description: LeaderId is the hex-encoded ID of the current etcd cluster
+              leaderID:
+                description: LeaderID is the hex-encoded ID of the current etcd cluster
```
ahrtr: for this kind of change, it should be in a separate PR, so that it can be merged soon.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: etcd-operator
-  newTag: v0.1
+  newName: arkasaha30/etcd-operator
+  newTag: auto-cert1
```
ivanvc: I think you forgot to unstage this file.

test/e2e/e2e_suite_test.go
```diff
@@ -36,7 +36,7 @@ import (
 	"sigs.k8s.io/e2e-framework/support/kind"
 
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
-	test_utils "go.etcd.io/etcd-operator/test/utils"
```
ivanvc, line 39: I don't think we should change this. There is a linter we enabled in the etcd repository that enforces `snake_case` for package names. Reference: https://go.dev/blog/package-names

test/e2e/cert_manager_test.go
```diff
@@ -23,6 +24,7 @@ import (
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 	"go.etcd.io/etcd-operator/pkg/certificate/cert_manager"
 	interfaces "go.etcd.io/etcd-operator/pkg/certificate/interfaces"
+	testUtils "go.etcd.io/etcd-operator/test/utils"
```
ivanvc: Therefore, I think this should be `test_utils`.

config/manager/kustomization.yaml
ArkaSaha30: Yes, updated it. Thanks!

test/e2e/cert_manager_test.go
ArkaSaha30: Sure, updated the PR, thanks!

pkg/certificate/auto/provider.go
```diff
@@ -40,7 +41,8 @@ type Provider struct {
 
 var _ interfaces.Provider = (*Provider)(nil)
 
-func New(c client.Client) interfaces.Provider {
+func New(c client.Client, scheme *runtime.Scheme) interfaces.Provider {
```
ivanvc: If we don't need the scheme, I suggest prefixing it with an underscore or replacing it with `_`.

```suggestion
func New(c client.Client, _scheme *runtime.Scheme) interfaces.Provider {
```

pkg/certificate/cert_manager/provider.go
```diff
@@ -37,7 +38,10 @@ type CertManagerProvider struct {
 
 var _ interfaces.Provider = (*CertManagerProvider)(nil)
 
-func New(c client.Client) interfaces.Provider {
+func New(c client.Client, scheme *runtime.Scheme) interfaces.Provider {
```
ivanvc: Ditto.

pkg/certificate/certificate.go
```diff
@@ -18,12 +19,12 @@ const (
...
 
-func NewProvider(pt ProviderType, c client.Client) (certInterface.Provider, error) {
+func NewProvider(pt ProviderType, c client.Client, scheme *runtime.Scheme) (certInterface.Provider, error) {
 	switch pt {
 	case Auto:
-		return auto.New(c), nil
+		return auto.New(c, scheme), nil
 	case CertManager:
-		return certManager.New(c), nil
+		return certManager.New(c, scheme), nil
```
ivanvc: I'm confused. Why are we introducing this new `scheme` parameter, but we're not using it in either `auto` or `certManager` functions?

ArkaSaha30: While testing, I was facing some scheme failures, so I refactored the code by passing the scheme in the entire call chain(probably the flaky test, which was fixed last week) 
But now, after removing the redundant scheme refs, it is working correctly for both auto and cert-manager.

Thank you for flagging this!!

internal/controller/etcdcluster_controller.go
```diff
@@ -337,11 +337,25 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
...
 		Owns(&corev1.Service{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&certv1.Certificate{}).
-		Complete(r)
+		Owns(&corev1.ConfigMap{})
+
+	// Conditionally watch cert-manager Certificate resources if CRDs are installed
+	// This allows the controller to react to Certificate status changes when using cert-manager provider
+	certList := &certv1.CertificateList{}
+	if err := mgr.GetClient().List(context.Background(), certList); err == nil {
```
ahrtr: How about adding a function something like below?

```
func isCertManagerCRDPresent(mgr manager.Manager) bool {
	gvk := certv1.SchemeGroupVersion.WithKind("Certificate")
	_, err := mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
	return err == nil
}
```

ArkaSaha30: Thank you, I have updated the PR with this

ahrtr (approved):

PR #238, opened 2025-11-08
dependency: Bump to Go version 1.25.4

ronaldngounou:
Contributes to https://github.com/etcd-io/etcd/issues/20891

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.25.3-bookworm@sha256:ee420c17fa013f71eca6b35c3547b854c838d4f26056a34eb6171bba5bf8ece4 AS builder
+FROM golang:1.25.4-bookworm@sha256:7419f544ffe9be4d7cbb5d2d2cef5bd6a77ec81996ae2ba15027656729445cc4 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.25.3-bookworm@sha256:ee420c17fa013f71eca6b35c3547b854c838d4f26056a34eb6171bba5bf8ece4 AS builder
+FROM golang:1.25.4-bookworm@sha256:<TO_REPLACE> AS builder
```
ivanvc: Why did you leave it like this? The index digest should be `sha256:7419f544ffe9be4d7cbb5d2d2cef5bd6a77ec81996ae2ba15027656729445cc4`

ronaldngounou: I was not sure where to find the index digest. 

This was my thought process:

I have checked:

Docker hub, [hub.docker.com/_/golang](https://hub.docker.com/_/golang) which led me to this part of the repo where the digest are hosted. [docker-library/golang@fca8040/1.25/bookworm/Dockerfile](https://github.com/docker-library/golang/blob/fca8040176e4e0c3d409c87d58fd8a81b16f2429/1.25/bookworm/Dockerfile)
However, I am not sure which digest to select as the architecture differs (amd64, armhf, arm64,...)

Thank you for the reviews.

ahrtr (approved):

PR #246, opened 2025-11-27
Change imagePullPolicy to IfNotPresent

nwnt:
By default, kubernetes sets the imagePullPolicy to Always.  I think we don't always need to pull every time.  For development, it should help with removing the need for an external image registry.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -64,6 +64,7 @@ spec:
           - --leader-elect
           - --health-probe-bind-address=:8081
         image: controller:latest
+        imagePullPolicy: IfNotPresent
         name: manager
         securityContext:
           allowPrivilegeEscalation: false


config/manager/manager.yaml
```diff
@@ -64,6 +64,7 @@ spec:
           - --leader-elect
           - --health-probe-bind-address=:8081
         image: controller:latest
+        imagePullPolicy: IfNotPresent
```
ahrtr, line 67: The image above is controller:latest, I think the `always` makes more sense. It should be good to use the default value.

ahrtr:
It seems that it's useful for local test. @hakman wdyt?

config/manager/manager.yaml
hakman: I agree, as long as `etcd-operator` is in development, and image tag is `latest`, image pull policy should be `Always`.

nwnt: Maybe this is a somewhat different concern.  The major pain point is the project seems to explicitly require a container registry, but in practice we don't need it.  The current approach requires that the (latest) image is always pulled from the registry.  I think if we can move towards using setting a kind cluster and pushing the built image to it, it should be faster and easier.

ahrtr: @nwnt Is this causing any issue right now? It seems the e2e tests are not affected.

Or you just see this issue when you tried to manually deploy etcd-operator in your local K8s cluster?  If it's true, then I'd like keep it as it's (do not change anything).

ArkaSaha30:
> > to build the image and upload it to the local kind
> 
> 
> 
> I believe the existing e2e test already supports this. cc @ArkaSaha30 @abdurrehman107 

Yes, it currently supports.

Issue #251, opened 2025-12-09
Refactor certificate management code bits


ArkaSaha30:
This issue is to track the refactoring the certificate management code bits in follow-up PRs:

- [ ] Return and handle error for `ValidityDuration` check. 
        refs: https://github.com/etcd-io/etcd-operator/blob/main/internal/controller/utils.go#L572-L574
                 https://github.com/etcd-io/etcd-operator/blob/main/internal/controller/utils.go#L605-L607
                 https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2603878163
         **Fix PR: https://github.com/etcd-io/etcd-operator/pull/253**

- [ ] Refactor Auto Cert Provider struct to include `userAutoConfig`
        refs: https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/auto/provider.go#L34-L46
                 https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2603890659
         **Fix PR: https://github.com/etcd-io/etcd-operator/pull/255**

- [x] Add `parseCert.NotBefore` check for the CertManager and Auto provider validity.
        refs: https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/cert_manager/provider.go#L145-L147
                 https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/auto/provider.go#L122-L124
                 https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2603993085
        **Fix PR: https://github.com/etcd-io/etcd-operator/pull/254**

- [x] Refactor interfaces definition and its usages to pass `ctrlruntimeclient.ObjectKey` instead of `secretName , namespace` combination. Refs: https://github.com/etcd-io/etcd-operator/pull/227#discussion_r2581641996 
**Fix PR: https://github.com/etcd-io/etcd-operator/pull/252**

   - EnsureCertificateSecret: https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/interfaces/interface.go#L79
   - ValidateCertificateSecret: https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/interfaces/interface.go#L94C2-L94C27
   - DeleteCertificateSecret: https://github.com/etcd-io/etcd-operator/blob/main/pkg/certificate/interfaces/interface.go#L107C2-L107C25
[...]

PR #253, opened 2025-12-14
Handle invalid ValidityDuration user input

ArkaSaha30:
This PR will handle a scenario of invalid `ValidityDuration` user input for the following certificate providers and throw an error.
In case `ValidityDuration` is not defined by the user, it will default to the corresponding default values:

- Cert-manager Provider, default 90days
- Auto Provider, default 365days

Fixes: https://github.com/etcd-io/etcd-operator/issues/251

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -596,7 +596,12 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) (*certInterface.Confi
 			IPs:      make([]net.IP, len(cmConfig.AltNames.DNSNames)),
 		}
 	} else {
-		defaultDNSNames := []string{fmt.Sprintf("%s.svc.cluster.local", cmConfig.CommonName)}
+		// Use wildcard DNS for the cluster's headless service to cover all pods
+		// This allows the certificate to work for pod-0, pod-1, etc.
+		defaultDNSNames := []string{
+			fmt.Sprintf("*.%s.%s.svc.cluster.local", ec.Name, ec.Namespace),
+			fmt.Sprintf("%s.%s.svc.cluster.local", ec.Name, ec.Namespace),
+		}
 		getAltNames = certInterface.AltNames{
 			DNSNames: defaultDNSNames,
 		}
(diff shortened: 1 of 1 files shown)


internal/controller/utils.go
```diff
@@ -596,14 +604,22 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) *certInterface.Config
...
+	// Set default duration to 365 days for auto provider if not provided
+	var duration time.Duration
+	if autoConfig.ValidityDuration == "" {
+		duration = certInterface.DefaultAutoValidity
+	} else {
+		var err error
+		duration, err = time.ParseDuration(autoConfig.ValidityDuration)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse ValidityDuration: %w", err)
+		}
```
ahrtr: Please add a function something like below, and reuse it for both `createCMCertificateConfig` and `createAutoCertificateConfig`
```
func parseValidityDuration(customizedDuration string, defaultDuration time.Duration) (time.Duration, error)
```

pkg/certificate/auto/provider.go
```diff
@@ -252,7 +248,7 @@ func (ac *Provider) createNewSecret(ctx context.Context, secretKey client.Object
 
 	log.Printf("calling SelfCert with hosts: %v", hosts)
 
-	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, uint(validity/DefaultValidity))
+	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, uint(validity/interfaces.DefaultAutoValidity))
```
ivanvc: I think that dividing by `interfaces.DefaultAutoValidity` may be error-prone in the future (i.e., if someone changes the `DefaultAutoValidity` to anything other than one year). I think this division needs to be fixed to be by `365 * 24 * time.Hour`, as `transport.SelfCert(...)` expects this to be years.

It could be another constant defined in this function.

internal/controller/utils.go
ArkaSaha30: Sure, updated the PR.

pkg/certificate/auto/provider.go
ArkaSaha30: Sure, updated the PR

ArkaSaha30: I'll debug and fix the failing test as well

ivanvc (commented):
LGTM. Thanks, Arka.

ivanvc (approved):

PR #255, opened 2025-12-14
Include userAutoConfig in Auto Provider struct

ArkaSaha30:
This PR will refactor AutoCert Provider struct to include userAutoConfig.

Part of: https://github.com/etcd-io/etcd-operator/issues/251


pkg/certificate/auto/provider.go
```diff
@@ -33,22 +33,22 @@ const (
 
 type Provider struct {
 	client.Client
+	userAutoConfig *interfaces.Config
```
ahrtr:
```suggestion
	config *interfaces.Config
```

Issue #197, opened 2025-08-27
Add e2e test to verify that the statefulSets, headless service and configMap have the correct ownerReference


frederiko:
As a follow up of https://github.com/etcd-io/etcd-operator/pull/194 proposed clean up, add e2e to ensure that EtcdCluster resource owns the resources that it created.

PR #268, opened 2026-01-19
Verify OwnerReference

GunaKKIBM:
https://github.com/etcd-io/etcd-operator/issues/197 

A separate test to verify owner reference didn't seem necessary, given that successful creation of the statefulset, service, any secondary resource here for the operator also means that the owner reference has to be correct, I have added the assert condition in the existing test case itself.

### DIFF
--- a/internal/controller/utils_test.go
+++ b/internal/controller/utils_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -168,6 +169,9 @@ func TestCreateHeadlessServiceIfNotExist(t *testing.T) {
 			"app":        "test-etcd",
 			"controller": "test-etcd",
 		}, service.Spec.Selector)
+		// Verify service is controlled by EtcdCluster
+		require.Len(t, service.OwnerReferences, 1)
+		require.Equal(t, service.OwnerReferences[0].Name, ec.Name)
 	})
 
 	t.Run("does not create service if it already exists", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/controller/utils_test.go
```diff
@@ -168,6 +168,9 @@ func TestCreateHeadlessServiceIfNotExist(t *testing.T) {
 			"app":        "test-etcd",
 			"controller": "test-etcd",
 		}, service.Spec.Selector)
+		// Verify OwnerReferences
+		assert.Len(t, service.OwnerReferences, 1)
+		assert.Equal(t, service.OwnerReferences[0].Name, "test-etcd")
```
ahrtr: minor comment
```suggestion
		assert.Equal(t, service.OwnerReferences[0].Name, ec.Name)
```

internal/controller/utils_test.go
```diff
@@ -168,6 +168,9 @@ func TestCreateHeadlessServiceIfNotExist(t *testing.T) {
 			"app":        "test-etcd",
 			"controller": "test-etcd",
 		}, service.Spec.Selector)
+		// Verify OwnerReferences
+		assert.Len(t, service.OwnerReferences, 1)
```
ahrtr: the test failure is related this change.

ahrtr: Did you figure out what's the root cause for previous failure?

ahrtr: I see that you changed to below.  I recall previous failure showed that the service.OwnerReferences is empty? Shouldn't it always non-empty? 

```
		if (assert.Len(t, service.OwnerReferences, 1)) {
			assert.Equal(t, service.OwnerReferences[0].Name, ec.Name)
		}
```

GunaKKIBM: The previous failure was here, https://github.com/etcd-io/etcd-operator/pull/268/files#diff-429334b066a97ed558217b47f55a1fc71836df740243033b9a4a6ac1829b12e5R488

The ownerreferences wasn't checked against updatedConfigmap, it was getting checked against configMap, which obviously wasn't the updated one after calling applyEtcdClusterState

ahrtr: @GunaKKIBM did you see this question?

ahrtr: What is the difference between 

```
		if (assert.Len(t, service.OwnerReferences, 1)) {
			assert.Equal(t, service.OwnerReferences[0].Name, ec.Name)
		}
```

and 

```
		assert.Len(t, service.OwnerReferences, 1)
		assert.Equal(t, service.OwnerReferences[0].Name, ec.Name)
		
```
?

For the former case, if  `len(service.OwnerReferences)` isn't 1, will it fail the test case as well?

GunaKKIBM: `if (assert.Len(t, service.OwnerReferences, 1)) {
	assert.Equal(t, service.OwnerReferences[0].Name, ec.Name)
}`
		
This is safe, because let's say owner references is empty, yes the first assert fails

`assert.Len(t, service.OwnerReferences, 1)`

but the next assert crashes, because of this indexing - `service.OwnerReferences[0].Name`

Hence, it would be better to run verify the owner reference name only if it exists

ahrtr: > but the next assert crashes, because of this indexing - `service.OwnerReferences[0].Name`

then just change assert to require

ahrtr: > > but the next assert crashes, because of this indexing - `service.OwnerReferences[0].Name`
> 
> then just change assert to require

and remove the if statement

GunaKKIBM: Done! Thank you!

ahrtr (approved):
LGTM & thx

internal/controller/utils_test.go
mspoonauer: Removed the redundant `is_blocked` field and `is_blocked()` method. `block_disposition` is now the single source of truth for enforcement outcome. `response` is still set when PX returns a block response (both Serve and Suppress paths).

Issue #130, opened 2025-04-10
Local docker restart


antondalgren:
Running a 3 node cluster with the following setup causes a infinite crash loop when the local docker environment restarts, I'm using a mac. 
```yaml
apiVersion: operator.etcd.io/v1alpha1
kind: EtcdCluster
metadata:
  name: etcd-cluster
spec:
  version: v3.5.16
  size: 3
```

logs
```
kubectl logs pods/etcd-cluster-0
{"level":"info","ts":"2025-04-10T09:44:41.554777Z","caller":"flags/flag.go:113","msg":"recognized and used environment variable","variable-name":"ETCD_DATA_DIR","variable-value":"/var/lib/etcd"}
{"level":"info","ts":"2025-04-10T09:44:41.554839Z","caller":"flags/flag.go:113","msg":"recognized and used environment variable","variable-name":"ETCD_INITIAL_CLUSTER","variable-value":"etcd-cluster-0=http://etcd-cluster-0.etcd-cluster.default.svc.cluster.local:2380,etcd-cluster-1=http://etcd-cluster-1.etcd-cluster.default.svc.cluster.local:2380,etcd-cluster-2=http://etcd-cluster-2.etcd-cluster.default.svc.cluster.local:2380"}
{"level":"info","ts":"2025-04-10T09:44:41.554844Z","caller":"flags/flag.go:113","msg":"recognized and used environment variable","variable-name":"ETCD_INITIAL_CLUSTER_STATE","variable-value":"existing"}
{"level":"warn","ts":"2025-04-10T09:44:41.554866Z","caller":"embed/config.go:689","msg":"Running http and grpc server on single port. This is not recommended for production."}
{"level":"info","ts":"2025-04-10T09:44:41.554883Z","caller":"etcdmain/etcd.go:73","msg":"Running: ","args":["/usr/local/bin/etcd","--name=etcd-cluster-0","--listen-peer-urls=http://0.0.0.0:2380","--listen-client-urls=http://0.0.0.0:2379","--initial-advertise-peer-urls=http://etcd-cluster-0.etcd-cluster.default.svc.cluster.local:2380","--advertise-client-urls=http://etcd-cluster-0.etcd-cluster.default.svc.cluster.local:2379"]}
{"level":"warn","ts":"2025-04-10T09:44:41.554907Z","caller":"embed/config.go:689","msg":"Running http and grpc server on single port. This is not recommended for production."}
[...]

Issue #26, opened 2025-01-13
Enhance the reconcile process to fix the potential inconsistency between etcd cluster and statefulset


ahrtr:
Fix the following TODO item,

https://github.com/etcd-io/etcd-operator/blob/f975f942f6e482ca3257bc906087be98cf06c0d1/internal/controller/etcdcluster_controller.go#L144-L158

Please also read https://github.com/etcd-io/etcd-operator/pull/17#discussion_r1908515471

We also need to add e2e tests to verify it. We can leverage the [gofailpoint](https://github.com/etcd-io/gofail) to intentionally crash the etcd-operator right **after** updating the etcd cluster but **before** updating the statefulset.

ahrtr:
cc @gdasson

soma00333:
Hi @ahrtr 
Good day
I'm working here https://github.com/etcd-io/etcd-operator/pull/50
Please assign to me

soma00333:
Hi @ahrtr 
I hope you’re doing well.
Regarding the discussion in https://github.com/etcd-io/etcd-operator/issues/19 about removing ginkgo, I have a question:
	1.	Limit the scope of https://github.com/etcd-io/etcd-operator/issues/26 to `etcdcluster_controller.go` and create a separate issue for the corresponding e2e tests
	2.	Wait for https://github.com/etcd-io/etcd-operator/issues/19 to be resolved before proceeding with https://github.com/etcd-io/etcd-operator/issues/26

Which approach would be better?
The `etcdcluster_controller.go` has already been fixed.
https://github.com/etcd-io/etcd-operator/pull/50/files#diff-7e023adf1c245d768f95e3f174cab882db08efd47023a9ce81e29a3a7767c12a

abdurrehman107:
@soma00333 we can have #19 PR by Sunday (26th January) eve. 
cc @ahrtr

soma00333:
@abdurrehman107 
I got it. I appreciate your hard work.
Take care, and have a good weekend

abdurrehman107:
You too @soma00333

soma00333:
[note]
After the discussion in https://github.com/etcd-io/etcd-operator/pull/50, we decided to create a new PR to change only `internal/controller/etcdcluster_controller.go` (To enhance the logic. No gofail this time.)
The PR is here https://github.com/etcd-io/etcd-operator/pull/53

ivanvc:
cc. @nwnt

PR #271, opened 2026-01-20
Setup gofail for e2e tests

nwnt:
Set up `gofail`, the prerequisite for #26 and #130.  

cc @ahrtr @ivanvc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,8 +85,19 @@ test: manifests generate fmt vet envtest ## Run tests.
 # - PROMETHEUS_INSTALL_SKIP=true
 # - CERT_MANAGER_INSTALL_SKIP=true
 .PHONY: test-e2e
-test-e2e: generate fmt vet kind ## Run the e2e tests. Expected an isolated environment using Kind.
+test-e2e: generate fmt vet kind gofail-enable ## Run the e2e tests. Expected an isolated environment using Kind.
 	ETCD_VERSION="$(E2E_ETCD_VERSION)" PATH="$(LOCALBIN):$(PATH)" go test ./test/e2e/ -v
+	$(MAKE) gofail-disable
+
+.PHONY: gofail-enable
+gofail-enable: gofail
+	$(GOFAIL) enable ./internal/controller/
+	go mod tidy
+
+.PHONY: gofail-disable
+gofail-disable: gofail
+	$(GOFAIL) disable ./internal/controller/
+	go mod tidy
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter
@@ -206,6 +217,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 CRD_REF_DOCS ?= $(LOCALBIN)/crd-ref-docs
 KIND ?= $(LOCALBIN)/kind
+GOFAIL ?= $(LOCALBIN)/gofail
 
 ## Tool Versions
 KUSTOMIZE_VERSION = $(shell cd tools/mod && go list -m -f {{.Version}} sigs.k8s.io/kustomize/kustomize/v5)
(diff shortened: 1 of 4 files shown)


Dockerfile.e2e
```diff
@@ -0,0 +1,41 @@
...
+
+# Copy the go source
+COPY cmd/main.go cmd/main.go
+COPY api/ api/
+COPY internal/ internal/
+COPY pkg/ pkg/
+
+# Enable gofail
+RUN gofail enable internal/controller
+RUN go get go.etcd.io/gofail/runtime
```
ahrtr: Probably `go mod tidy` makes more sense. It will automatically update the go.mod/go.sum file.

internal/controller/etcdcluster_controller.go
```diff
@@ -292,6 +292,9 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 
 		logger.Info("Learner member added successfully", "peerURLs", peerURL)
 
+		// gofail: var exceptionAfterMemberAdd struct{}
+		// return ctrl.Result{}, fmt.Errorf("failpoint after adding a member")
```
ahrtr: probably we can inject a panic failpoint to make etcd-operator panicking

internal/controller/etcdcluster_controller.go
```diff
@@ -312,6 +315,9 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 			return ctrl.Result{}, err
 		}
 
+		// gofail: var exceptionAfterMemberDelete struct{}
+		// return ctrl.Result{}, fmt.Errorf("failpoint after removing a member")
```
ahrtr: ditto

Dockerfile.e2e
nwnt: Done. By the way, I had it like that by following the doc [here](https://github.com/etcd-io/gofail/blob/master/doc/design.md#step-2-build-your-application-with-failpoints).  Should that be updated then?  I can send a PR to do so.

internal/controller/etcdcluster_controller.go
nwnt: Yep, removed the custom return statement.  The follow up PR will setup the failpoint with the panic gofail term.

internal/controller/etcdcluster_controller.go
nwnt: same as above.

ahrtr:
Overall looks good. Only one comment, I don't think it is good to add a separate Dockerfile.e2e for test only. Ideally, we should instrument the source code before building the image.

We can update the Makefile something like below. But etcd-operator doesn't depend on gofail (go.mod doesn't have entry `go.etcd.io/gofail v0.2.0`), we should resolve this.
```
docker-build: gofail-enable
```

nwnt:
Yep, I removed Dockerfile.e2e, and wrapped `gofail-enable` and `gofail-disable` around `test-e2e` instead.  This should make it not pollute the code base with the generated code from `gofail` when testing e2e locally.

internal/controller/etcdcluster_controller.go
```diff
@@ -292,6 +292,8 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 
 		logger.Info("Learner member added successfully", "peerURLs", peerURL)
 
+		// gofail: var exceptionAfterMemberAdd struct{}
```
ahrtr: Minor comment: We should add failpoint in the PR when you add the e2e test which uses the failpoint.

internal/controller/etcdcluster_controller.go
```diff
@@ -312,6 +314,8 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 			return ctrl.Result{}, err
 		}
 
+		// gofail: var exceptionAfterMemberDelete struct{}
```
ahrtr: ditto

internal/controller/etcdcluster_controller.go
nwnt: Yep, done.

Issue #265, opened 2026-01-06
Return all health errors instead of just the first one


ahrtr:
During each reconciliation process, etcd-operator performs health check on all etcd members. Currently it just returns the first error if any. We should return all errors instead.

https://github.com/etcd-io/etcd-operator/blob/6791fdc26be39a802205f703975c30c5ddca45d0/internal/controller/utils.go#L543-L549

GunaKKIBM:
@ahrtr , can this issue be closed. We have already implemented publishing status as well. https://github.com/etcd-io/etcd-operator/issues/135. I do not have the permissions to close it. thank you!

PR #274, opened 2026-01-23
Return all health errors instead

GunaKKIBM:
https://github.com/etcd-io/etcd-operator/issues/265 
Return the health status of all members instead of just one.

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -20,6 +20,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


internal/controller/utils.go
```diff
@@ -540,15 +540,16 @@ func healthCheck(sts *appsv1.StatefulSet, lg klog.Logger) (*clientv3.MemberListR
 		return memberlistResp, nil, err
 	}
 
+	var memberErrors = make([]error, 3)
```
```suggestion
	var memberErrors []error
```

internal/controller/utils.go
```diff
@@ -540,15 +540,16 @@ func healthCheck(sts *appsv1.StatefulSet, lg klog.Logger) (*clientv3.MemberListR
...
 		if !healthInfo.Health {
 			// TODO: also update metrics?
-			return memberlistResp, healthInfos, errors.New(healthInfo.String())
+			memberErrors = append(memberErrors, errors.New(healthInfo.String()))
 		}
 		lg.Info(healthInfo.String())
 	}
 
-	return memberlistResp, healthInfos, nil
+	return memberlistResp, healthInfos, errors.Join(memberErrors...)
```
neolit123: please use the error aggregate from here:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L47

and then call Error() on it when returning.

errors.Join is not nice for controllers since it splits them on new lines.

internal/controller/utils.go
GunaKKIBM: Done

Issue #168, opened 2025-06-24
Upgrade across patches or one minor version


ivanvc:
This is a priority for 0.2.0. We need to support a minor or patch version upgrade. Add e2e tests, and verify that it is indeed an upgrade and satisfies the previous two statements.

ahrtr:
We need to add validation:
- [allowed]: upgrade across patches of the minor version: 3.5.20 -> 3.5.21
- [allowed]: upgrade across one minor version: 3.5.x -> 3.6
  - A special case: when upgrading from 3.5.x to 3.6.x, users should upgrade to 3.5.20+ first. Refer to https://etcd.io/blog/2025/upgrade_from_3.5_to_3.6_issue/
- [Not allowed]: Upgrade across multiple minor version isn't allowed.

Also add e2e test.

Ab-hishek:
@ahrtr How are these upgrades check related to etcd-operator? I mean what role does etcd-operator play here in allowing/disallowing the upgrades? Is it that the version field in the etcdCluster spec (https://github.com/etcd-io/etcd-operator/blob/8af4ef07a0d41989e951eda4531d91f66e1dfaa2/api/v1alpha1/etcdcluster_types.go#L39) needs to have such validations?

Lets say the statefulset replica(s) owned by etcdCluster was at version X initially. And then a user made an update to that version field of etcdCluster CR to update it from X to Y. And based on the new update version (Y), we will need to verify the current running version(X) and compare it with the updated version?

Ab-hishek:
Seeing this image: https://github.com/etcd-io/etcd-operator/blob/main/docs/design/reconcile_loop_v0.1.0.png, I feel my understanding is correct. But had a question which is mentioned in slack: https://kubernetes.slack.com/archives/C07DU767BEE/p1753179771886479?thread_ts=1753164208.804879&cid=C07DU767BEE

ivanvc:
Hi, @Ab-hishek. I replied to your question on Slack. Let me know if it's clear or if you need further clarification.

ahrtr:
Several things to clarify (also open to discussion),
- We only allow upgrade across one minor version (or patches version of the same minor version), we can get `currentVersion` from the status (to be added in https://github.com/etcd-io/etcd-operator/issues/182), and get the `targetVesion` from CR EtcdCluster.
[...]

PR #278, opened 2026-02-04
Add upgrade path docs and validation

neolit123:
```
add initial docs/upgrade.md
```

```

Add a new util function that can compare a current and target
etcd versions. Use the function to validate if the current
STS version can be upgraded to the target version
from the EtcdCluster version.

If the versions cannot be parsed, assume that no validation
will be performed.

Add UTs.

Add a couple of TODOs that we may want to make this
configurable from the CRD and also use the cluster Status
to check the current version instead of the STS image version.
```

xref
https://github.com/etcd-io/etcd-operator/issues/168#issuecomment-3843253447

but i don't know if i will have more time to work on this, so someone might have to fork my commit from this PR.

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
@@ -33,6 +34,7 @@ import (
 
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
 	"go.etcd.io/etcd-operator/internal/etcdutils"
+	etcdversions "go.etcd.io/etcd/api/v3/version"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
@@ -144,6 +146,52 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		if err := checkStatefulSetControlledByEtcdOperator(ec, sts); err != nil {
 			logger.Error(err, "StatefulSet is not controlled by this EtcdCluster resource")
 			return nil, ctrl.Result{}, err
+		}
+
+		// If the version to be reconciled is unsupported, throw an error.
+		if len(sts.Spec.Template.Spec.Containers) == 0 {
+			logger.Error(err, "StatefulSet has no containers yet")
+			return &reconcileState{cluster: ec, sts: sts}, ctrl.Result{}, nil
+		}
+		stsImage := sts.Spec.Template.Spec.Containers[0].Image
+		// Note: createOrPatchStatefulSet() only supports the "registry-path:image" format.
+		// TODO: switch to using the version from ec.Status eventually.
+		//   https://github.com/etcd-io/etcd-operator/pull/278/changes#r2764796805
+		idx := strings.Index(stsImage, ":")
+		if idx == -1 {
+			logger.Error(err, "could not extract image version from StatefulSet image",
+				"image", stsImage)
+			return &reconcileState{cluster: ec, sts: sts}, ctrl.Result{}, nil
+		}
+		currentVersion := stsImage[idx+1:]
+		targetVersion := ec.Spec.Version
+
+		// Only handle cases when there is a version change.
+		if currentVersion != targetVersion {
+			// TODO: consider adding an option in the CRD called allowCustomImageUpgrade to make
+			// the behavior here optional:
+			//   https://github.com/etcd-io/etcd-operator/pull/278/changes#r2764717418
+			canParse, err := validateEtcdUpgradePath(etcdversions.AllVersions, currentVersion, targetVersion)
+			if !canParse {
+				logger.Info("error when parsing reconcile versions; it is your responsibility "+
+					"to validate if the upgrade path is supported",
+					"current", currentVersion,
+					"target", targetVersion,
+					"error", err,
+				)
+				return &reconcileState{cluster: ec, sts: sts}, ctrl.Result{}, nil
+			} else {
+				if err != nil {
+					logger.Error(err, "unsupported upgrade path between current and target versions",
+						"current", currentVersion,
+						"target", targetVersion,
+					)
+					return nil, ctrl.Result{}, err
+				}
+				logger.Info("upgrade path between current and target versions is supported",
+					"current", currentVersion,
+					"target", targetVersion)
+			}
 		}
 	}
 
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -145,6 +146,34 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
+		}
+		stsImage := sts.Spec.Template.Spec.Containers[0].Image
+		// Note: createOrPatchStatefulSet() only supports the "registry-path:image" format.
+		idx := strings.Index(stsImage, ":")
+		if idx == -1 {
+			logger.Error(err, "could not extract image version from StatefulSet image",
+				"image", stsImage)
+			return &reconcileState{cluster: ec, sts: sts}, ctrl.Result{}, nil
+		}
+		currentVersion := stsImage[idx+1:]
```
neolit123, line 166: should i be taking the current version from ec.Status instead?

internal/controller/utils.go
```diff
@@ -776,3 +777,30 @@ func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c c
...
+	}
+	targetVer, err = k8sverion.ParseGeneric(target)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse target version %s: %w", target, err)
+	}
+	// Major version upgrades are undefined.
+	if targetVer.Major() != currentVer.Major() {
+		return true, nil
+	}
+	if targetVer.Minor() > currentVer.Minor()+1 || targetVer.Minor() < currentVer.Minor()-1 {
```
neolit123: should it error if target is < current?
i.e. downgrade

docs/upgrade.md
```diff
@@ -0,0 +1,15 @@
...
+paths](https://etcd.io/docs/v3.7/upgrades/) as the main
+etcd project.
+
+If you are using the official etcd images for your etcd cluster
+managed by the etcd-operator, the operator will perform validation
+for the upgrade path and follow the same rules mentioned above.
+
+However, if you are using custom etcd images, no validation
+will be performed and you have to manually ensure that the upgrade
+path is supported.
```
ahrtr: For custom etcd images, if they have tags ( [Semantic Versioning](https://semver.org/) ), I think we also need to validate the upgrade per the same rules. If they don't have tags, then etcd-operator can't validate it, and  it's users' responsibility to ensure it's supported.

internal/controller/utils.go
ahrtr: etcd-operator doesn't support downgrade for now.

We need to ensure
-  either `target` and `current` are the same minor version, and target > current
- or `target` is the next major/minor compared to `current`. We can use the definitions below,

https://github.com/etcd-io/etcd/blob/7c22682124d0bf5081df448a81e19bd1df2e91dc/api/version/version.go#L38-L47

internal/controller/utils.go
```diff
@@ -776,3 +777,30 @@ func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c c
 
 	return nil
 }
+
+// validateTargetVersion checks if the current and target versions align
+// with the official upgrade paths for etcd. If one of the versions cannot
+// be parsed canParse is false.
+func validateTargetVersion(current, target string) (canParse bool, err error) {
```
ahrtr: The first return bool: indicate whether etcd-operator should allow the upgrade
The second return err: indicate whether there is any error during validation, e.g error parsing the version

etcd-operator should only depend on the first returned bool to allow or reject the upgrade. The second return error is only for log/debugging purpose. 

Probably we should add a field/flag something `allowCustomImageUpgrade` in CRD to let users configure the behaviour.  


The rough logic should be something like below,

```
if there is any error parsing the versions {
    if allowCustomImageUpgrade==true {
           allow the upgrade
   } else {
           reject the upgrade
   }
} else {
    if the versions meet the require {
         allow the upgrade
    } else {
         reject the upgrade
    }
}

```

internal/controller/etcdcluster_controller.go
ahrtr: It's a good point. Using the version from sts should be fine for now.

After this PR gets merged, I think we need to do some overall correction on the behaviour.  We should keep using the version from ec.Status in the following stages:
- Exception handling
- Promote learner
- Scale Out & In

Refer to https://github.com/etcd-io/etcd-operator/blob/main/docs/design/reconcile_loop_v0.1.0.png

We only consider using the version from ec.Spec during the real Upgrade stage.

internal/controller/utils.go
neolit123: i will leave a TODO for allowCustomImageUpgrade and this can be a separate PR.

docs/upgrade.md
neolit123: updated.

internal/controller/etcdcluster_controller.go
neolit123: left a TODO for this.

internal/controller/utils.go
neolit123: done.

docs/upgrade.md
```diff
@@ -0,0 +1,15 @@
+# Upgrading etcd clusters
+
+## Supported upgrade paths
+
+The etcd-operator follows the same rules in terms of [supported upgrade
+paths](https://etcd.io/docs/v3.7/upgrades/) as the main etcd project.
```
ahrtr:
```suggestion
paths](https://etcd.io/docs/latest/upgrades/) as the main etcd project.
```

internal/controller/etcdcluster_controller.go
```diff
@@ -145,6 +147,50 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
+		targetVersion := ec.Spec.Version
+		// TODO: consider adding an option in the CRD called allowCustomImageUpgrade to make
+		// the behavior here optional:
+		//   https://github.com/etcd-io/etcd-operator/pull/278/changes#r2764717418
+		canParse, err := validateEtcdUpgradePath(etcdversions.AllVersions, currentVersion, targetVersion)
```
ahrtr: suggest that we compare the version to check whether the version changes or not first, and only call `validateEtcdUpgradePath` if the version changes.  Otherwise, we will see a "Upgrade ..." log message in each reconciliation workflow.

internal/controller/utils.go
```diff
@@ -776,3 +777,51 @@ func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c c
...
+	}
+	targetVer, err = semver.NewVersion(target)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse target version %s: %w", target, err)
+	}
+
+	// If current and target versions are the same, consider the upgrade valid.
+	if currentVer.Major == targetVer.Major && currentVer.Minor == targetVer.Minor {
+		return true, nil
+	}
```
ahrtr: In this case, we also need to check the patch version, we don't support downgrade for now.

internal/controller/etcdcluster_controller.go
neolit123: instead i added `areEqual` to the function as another returned arg.
otherwise the version parsing becomes external to the function and i thought that's less preferred, code org wise.

internal/controller/etcdcluster_controller.go
```diff
@@ -145,6 +147,50 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
+		// the behavior here optional:
+		//   https://github.com/etcd-io/etcd-operator/pull/278/changes#r2764717418
+		canParse, areEqual, err := validateEtcdUpgradePath(etcdversions.AllVersions, currentVersion, targetVersion)
+		if !canParse {
+			logger.Info("error when parsing reconcile versions; it is your responsibility "+
+				"to validate if the upgrade path is supported",
+				"current", currentVersion,
+				"target", targetVersion,
+				"error", err,
+			)
```
ahrtr: Minor comment: The log message may not accurate.  Imagine a case: it's a custom image without tag, and currentVersion == targetVersion. In such case, it isn't an upgrade at all.

neolit123: good point!

i'm wrapping all of this in:

```
		// Only handle cases when there is a version change.
		if currentVersion != targetVersion {
```

and removing the `areEqual` arg.

i.e. don't handle any validation if the string versions are equal.

internal/controller/utils.go
```diff
@@ -776,3 +777,58 @@ func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c c
...
+		}
+		if idx+1 >= len(etcdVersions) {
+			return true, fmt.Errorf("current version %s is the latest version, "+
+				"no valid target versions available", currentVer)
+		}
+		validTargetVer = &etcdVersions[idx+1]
+		if targetVer.Major == validTargetVer.Major && targetVer.Minor == validTargetVer.Minor {
+			isTargetValid = true
+			break
+		}
```
ahrtr: It seems that you missed the case that the targetVer is the same major.minor as currentVer?

ahrtr: e.g. from 3.5.25 -> 3.5.26?

```suggestion
		if targetVer.Major == validTargetVer.Major && targetVer.Minor == validTargetVer.Minor {
			isTargetValid = true
		}
		break
```

neolit123: updated. i simplified the checks and added the patch level tests.

PR #323, opened 2026-03-26
fix: correct the behavior of certificate creation based on specified provider

anhtranbk:
Fixes https://github.com/etcd-io/etcd-operator/issues/314

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -592,6 +592,9 @@ func parseValidityDuration(customizedDuration string, defaultDuration time.Durat
 
 func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) (*certInterface.Config, error) {
 	cmConfig := ec.Spec.TLS.ProviderCfg.CertManagerCfg
+	if cmConfig == nil {
+		return nil, fmt.Errorf("cert-manager configuration is not present")
+	}
 
 	// Set default duration to 90 days for cert-manager if not provided
 	duration, err := parseValidityDuration(cmConfig.ValidityDuration, certInterface.DefaultCertManagerValidity)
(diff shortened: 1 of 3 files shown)


ahrtr:
@ArkaSaha30 can you take a look at this PR? thx

internal/controller/utils_test.go
```diff
@@ -901,3 +904,188 @@ func TestValidateEtcdUpgradePath(t *testing.T) {
...
+		{
+			name: "auto config with all fields set",
+			ec: &ecv1alpha1.EtcdCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-cluster",
+					Namespace: "test-namespace",
+				},
+				Spec: ecv1alpha1.EtcdClusterSpec{
+					TLS: &ecv1alpha1.TLSCertificate{
+						Provider: "auto",
```
ArkaSaha30: nit: can we use `string(certInterface.Auto)` instead of `"auto"`

internal/controller/utils_test.go
```diff
@@ -901,3 +904,188 @@ func TestValidateEtcdUpgradePath(t *testing.T) {
...
+		{
+			name: "cert-manager config with all fields set",
+			ec: &ecv1alpha1.EtcdCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-cluster",
+					Namespace: "test-namespace",
+				},
+				Spec: ecv1alpha1.EtcdClusterSpec{
+					TLS: &ecv1alpha1.TLSCertificate{
+						Provider: "cert-manager",
```
ArkaSaha30: Similarly here: `string(certInterface.CertManager)` instead of `"cert-manager"`

internal/controller/utils.go
```diff
@@ -632,6 +635,15 @@ func createCMCertificateConfig(ec *ecv1alpha1.EtcdCluster) (*certInterface.Confi
 
 func createAutoCertificateConfig(ec *ecv1alpha1.EtcdCluster) (*certInterface.Config, error) {
 	autoConfig := ec.Spec.TLS.ProviderCfg.AutoCfg
+	// Set default values for auto configuration if not present
+	if autoConfig == nil {
+		autoConfig = &ecv1alpha1.ProviderAutoConfig{
+			CommonConfig: ecv1alpha1.CommonConfig{
+				CommonName:       fmt.Sprintf("%s.%s.svc.cluster.local", ec.Name, ec.Namespace),
```
ArkaSaha30: nit: since we are using `svc.cluster.local` in multiple places(again while setting defaultDNSNames), can we make it as a constant and use it?

ArkaSaha30 (commented):
Thank you for the fix PR.
Overall LGTM with small nits.

internal/controller/utils_test.go
anhtranbk: Do you mean `string(certificate.Auto)` ?

internal/controller/utils.go
anhtranbk: updated. Not fully sure about the constant name. Happy to adjust based on your suggestion.

ahrtr (approved):
LGTM & thx

PR #330, opened 2026-04-07
Fix: Add TLS volume mounts to etcd container

LostGhost1:
The etcd operator was creating TLS volumes (server-secret, peer-secret)
but never mounting them to the etcd container. This caused etcd to fail
to start with 'no such file or directory' errors when trying to access
the TLS certificates.

Changes:
- Add TLS volume mounts to the etcd container when TLS is enabled
- Mount server-secret to /etc/etcd/server-secret
- Mount peer-secret to /etc/etcd/peer-secret
- Fix storage volume mount to append instead of overwrite existing mounts

This fixes the issue where etcd pods would crash loop because they
couldn't access the TLS certificates even though the volumes were
created.

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -188,6 +188,26 @@ func createOrPatchStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1a
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


LostGhost1:
I'll come back with a better quality change

PR #352, opened 2026-06-02
fix(cert-manager): resolve reconciliation issues with enableCertificateOwnerRef

### LINKED ISSUE #331: Cannot reconcile ETCD cluster when using cert-manager configured with `enableCertificateOwnerRef=true`
# Description
I have identified an issue in the ETCD operator during the reconciliation process of a cluster when using cert-manager as the TLS certificate provider. When cert-manager is configured to automatically manage TLS certificate generation, the reconciliation flow of the ETCD operator attempts to patch the Kubernetes `Secret` resources generated from `Certificate` resources. This patch operation is intended to add an owner reference so that the Secrets are marked as managed by the ETCD controller.

However, when cert-manager is installed with the Helm chart option `enableCertificateOwnerRef=true`, cert-manager already sets its own owner reference on the generated `Secret`resources. This leads to a conflict when the ETCD operator tries to modify the same field, causing the reconciliation process to fail.

# Environment
Kubernetes version: v1.35.1
ETCD Operator version: v0.2.0
cert-manager version: v1.20.0

# Steps to Reproduce
* Install cert-manager using Helm with `enableCertificateOwnerRef: true`
* Deploy the ETCD operator using the YAML file from the releases page.
* Create an ETCD cluster configured to use cert-manager for TLS certificates.
* Observe the reconciliation process.

# Expected Behavior
The ETCD operator should successfully reconcile the cluster even when cert-manager manages the Secrets, without conflicting over owner references.

# Actual Behavior
The reconciliation fails when attempting to patch Secrets generated by cert-manager due to conflicting owner references. Error example:

```
2026-04-08T10:06:14Z    ERROR    Reconciler error    {"controller": "etcdcluster", "controllerGroup": "operator.etcd.io", "controllerKind": "EtcdCluster", "EtcdCluster": {"name":"ptf0000000","namespace"
:"sare0000"}, "namespace": "sare0000", "name": "ptf0000000", "reconcileID": "ed373930-eda5-436a-a2a4-a0ec622a7e50", "error": "patching certificate secret: ptf0000000-server-tls with ownerReference faile
d: Object sare0000/ptf0000000-server-tls is already owned by another Certificate controller ptf0000000-server-tls"}
```

# Proposed solutions

#### Conditional ownership handling
The ETCD operator should detect if a Secret is already managed by cert-manager and avoid overriding or adding conflicting owner references.

#### Configurable behavior via spec parameter
Provide a configuration flag in the ETCD resource to disable `Secret` ownership patching when external certificate managers (like cert-manager) are used.

valen-mascarenhas14:
Fixes https://github.com/etcd-io/etcd-operator/issues/331

### Problem
The operator cannot reconcile ETCD clusters when using cert-manager configured with `enableCertificateOwnerRef=true`. This manifests in two ways:

1. **Certificate creation fails** due to invalid duration parsing:
   ```
   ERROR Failed to create Client Certificate. error: "failed to parse ValidityDuration: time: unknown unit \"d\" in duration \"365d\""
   ```

2. **Owner reference conflicts** when cert-manager's `enableCertificateOwnerRef=true` sets Certificate as the controller owner of TLS secrets, preventing the operator from managing them.

### Root Causes

**Duration Parsing Issue:**
Go's `time.ParseDuration` doesn't support day units (`d`), but the sample configuration and API documentation use day-based durations like `"365d"` and `"100d12h"`.

**Owner Reference Issue:**
When `enableCertificateOwnerRef=true`, cert-manager sets itself as the controller owner of TLS secrets. The operator was attempting to overwrite this ownership, causing reconciliation failures.

### Solution

**1. Duration Parsing Fix:**
Enhanced `parseValidityDuration` to automatically convert day units to hours:
- `"365d"` → `"8760h"` (365 × 24 hours)
- `"100d12h"` → `"2412h"` (100 × 24 + 12 hours)
- Maintains backward compatibility with hour-based durations

**2. Owner Reference Fix:**
Modified secret ownership logic to:
- Skip ownership patching when a controller owner already exists
- Preserve cert-manager's ownership of TLS secrets
- Log informational messages instead of errors


internal/controller/utils.go
```diff
@@ -578,12 +578,54 @@ func getPeerCertName(etcdClusterName string) string {
...
+			} else if ch == 'd' {
+				// Convert days to hours
+				if numStr.Len() > 0 {
+					days, err := strconv.ParseFloat(numStr.String(), 64)
+					if err != nil {
+						return 0, fmt.Errorf("failed to parse day value in ValidityDuration: %w", err)
+					}
+					hours := days * 24
+					result.WriteString(fmt.Sprintf("%gh", hours))
+					numStr.Reset()
```
GunaKKIBM: I think this chunk of code can be made very simple. Just trimming the last character and converting the remaining to Integer, should help. 

Please try and check!

GunaKKIBM:
@valen-mascarenhas14, could you please add the test results here? Pod logs should be good.

internal/controller/utils.go
valen-mascarenhas14: @GunaKKIBM  The current logic was kept generic to support duration strings like `23d4h` too

internal/controller/utils.go
```diff
@@ -578,12 +578,54 @@ func getPeerCertName(etcdClusterName string) string {
...
+			ch := durationStr[i]
+			if ch >= '0' && ch <= '9' || ch == '.' {
+				numStr.WriteByte(ch)
+			} else if ch == 'd' {
+				// Convert days to hours
+				if numStr.Len() > 0 {
+					days, err := strconv.ParseFloat(numStr.String(), 64)
+					if err != nil {
+						return 0, fmt.Errorf("failed to parse day value in ValidityDuration: %w", err)
+					}
```
ArkaSaha30: Can we use regex to simplify the parsing logic? 
Something like `regexp.MustCompile(`([0-9.]+)(d)`)` to get the no of days and then calculate the hours?

internal/controller/utils.go
```diff
@@ -578,12 +578,54 @@ func getPeerCertName(etcdClusterName string) string {
...
+			} else {
+				// Other unit (h, m, s, etc.)
+				if numStr.Len() > 0 {
+					result.WriteString(numStr.String())
+					numStr.Reset()
+				}
+				result.WriteByte(ch)
+			}
+		}
+		// Append any remaining number
```
ArkaSaha30: This block can also be omitted if we use regex instead of index parsing.

ArkaSaha30 (changes requested):
Can we validate if the cascading cleanup happens on EtcdCluster deletion when the certificate secret is having cert-manager ownerRef?

Manual validation should be good for now.

internal/controller/utils.go
valen-mascarenhas14: Thanks, have used regex to simplify the parsing logic

valen-mascarenhas14:
> Can we validate if the cascading cleanup happens on EtcdCluster deletion when the certificate secret is having cert-manager ownerRef?
Manual validation should be good for now.

@ArkaSaha30 I noticed that cascading cleanup would fail when the certificate secret had cert-manager ownerRef. I've made the code changes for the same .  Below is how it looks after the code changes. PTAL.

```
valen@valens-MacBook-Air etcd-operator % kubectl apply -f config/samples/sample_certmanager_etcdcluster.yaml                                                 

clusterissuer.cert-manager.io/selfsigned created
etcdcluster.operator.etcd.io/etcdcluster-sample-cert created
valen@valens-MacBook-Air etcd-operator % kubectl get certificates -A 
NAMESPACE   NAME                                 READY   SECRET                               AGE
default     etcdcluster-sample-cert-client-tls   True    etcdcluster-sample-cert-client-tls   14s
default     etcdcluster-sample-cert-peer-tls     True    etcdcluster-sample-cert-peer-tls     14s
default     etcdcluster-sample-cert-server-tls   True    etcdcluster-sample-cert-server-tls   14s
valen@valens-MacBook-Air etcd-operator % kubectl delete -f config/samples/sample_certmanager_etcdcluster.yaml

clusterissuer.cert-manager.io "selfsigned" deleted
etcdcluster.operator.etcd.io "etcdcluster-sample-cert" deleted from default namespace
valen@valens-MacBook-Air etcd-operator % kubectl get certificates -A                                         
No resources found
```

internal/controller/utils.go
```diff
@@ -578,19 +579,37 @@ func getPeerCertName(etcdClusterName string) string {
...
-	duration, err := time.ParseDuration(customizedDuration)
+
+	// Convert days to hours using regex since time.ParseDuration doesn't support 'd' unit
+	// Matches patterns like "365d", "100.5d", etc. and converts them to hours
+	// Example: "365d" -> "8760h", "100d12h" -> "2400h12h" (which time.ParseDuration sums to 2412h)
+	dayRegex := regexp.MustCompile(`([0-9.]+)d`)
+	durationStr := dayRegex.ReplaceAllStringFunc(customizedDuration, func(match string) string {
+		// Extract the numeric part (remove 'd')
+		daysStr := match[:len(match)-1]
+		days, err := strconv.ParseFloat(daysStr, 64)
```
GunaKKIBM, line 596: Can this be documented somewhere? how the end user needs to provide duration, and what they can expect, with a few unique examples
Thank you!

PR #361, opened 2026-06-09
Close DB connection post health check

GunaKKIBM:
While running some performance test (Integrated pprof and checked number of goroutines/heap usage)
With creation of every CR (new etcd cluster), the number of goroutines were increasing. This turned out to be because of the open DB connection.

### DIFF
--- a/internal/etcdutils/etcdutils.go
+++ b/internal/etcdutils/etcdutils.go
@@ -153,10 +153,17 @@ func ClusterHealth(eps []string) ([]EpHealth, error) {
 				healthCh <- EpHealth{Ep: ep, Health: false, Error: err.Error()}
 				return
 			}
+			defer func() {
+				err := cli.Close()
+				if err != nil {
+					lg.Warn("failed to close etcd client", zap.String("endpoint", ep), zap.Error(err))
+				}
+			}()
 			startTs := time.Now()
 			// get a random key. As long as we can get the response
 			// without an error, the endpoint is health.
 			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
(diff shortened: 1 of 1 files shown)


internal/etcdutils/etcdutils.go
hakman, line 155:
```suggestion
			}
			defer func() {
				if cerr := cli.Close(); cerr != nil {
					lg.Warn("failed to close etcd client", zap.String("endpoint", ep), zap.Error(cerr))
				}
			}()
```

internal/etcdutils/etcdutils.go
```diff
@@ -157,6 +157,14 @@ func ClusterHealth(eps []string) ([]EpHealth, error) {
...
 			// without an error, the endpoint is health.
 			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer func() {
+				err := cli.Close()
+				if err != nil {
+					cancel()
+					return
+				}
+				cancel()
+			}()
```
hakman, line 161:
```suggestion
			defer cancel()
```

hakman (changes requested):

internal/etcdutils/etcdutils.go
```diff
@@ -153,10 +153,18 @@ func ClusterHealth(eps []string) ([]EpHealth, error) {
 				healthCh <- EpHealth{Ep: ep, Health: false, Error: err.Error()}
 				return
 			}
+			defer func() {
+				err := cli.Close()
+				if err != nil {
+					lg.Warn("failed to close etcd client", zap.String("endpoint", ep), zap.Error(err))
+					return
```
hakman: @GunaKKIBM why is return needed?
```suggestion
```

ahrtr (approved):

PR #363, opened 2026-06-15
Replace StatefulSets with a customized controller

### LINKED ISSUE #362: Replace Statefulsets with customized controller
## Problem
Currently etcd-operator uses StatefulSets to manage the etcd member/PODs. In happy cases, it's working well so far. But when it comes to repair a failed member/POD, it's hard to do it with StatefulSet.

Statefulsets assigns an ordinal Index for each POD, and always scales down from the highest ordinal. If member 1 is corrupt but member 0 and 2 are healthy, Statefulsets forces you to remove member-2 first.

Also currently the controller manages `Statefulset.spec.replicas`, the StatefulSet only manages "how many pods to run".  It knows nothing about the etcd membership.  This introduces a reconciliation gap: we must constantly reconcile STS-replica-count vs etcd-member-count.

## Proposal

The solution is to introduce a customzed controller to manage the etcd members/PODs. We discussed multiple solutions multiple times in our community meetings. 
- we manage POD directly, each etcd member maps 1-to-1 to a Pod owned directly by EtcdCluster.
- we manage deployment or statefulset directly, and each etcd member maps to 1-to-1 to a Statefulset or deployment owned by EtcdCluster. For each statefulset or deployment, the replica is 1.

We believe the first solution is the most flexible. Ideally, we should deliver a PoC for each solution and decide to use which one later. But we might not have that bandwidth. Let's start from solution 1, and deliver a PoC for it first. If it's confirmed that it doesn't need huge effort and easy to understand & maintain. We might proceed with it.

Note we haven't released v1.0 yet, so we're happy to discard the existing design and implementation and start over in future if necessary.

## Requirements

We don't need to deliver a complete & perfect solution for now. Instead, let's do it step by step. The first step is just to replace the StatefulSet with a customized controller as mentioned in solution 1 above. We don't add new feature in the first step, we just keep the existing functionalities:
- **Users are able to create a brand new etcd cluster of any size**
- **Users are able scale in & out an existing etcd cluster**
- Users are able to configure the certificate providers: auto, cert-manager or disabled
- Users are able to configure a storageClass to integrate with an existing CSI driver
- Users are able to customze the etcd options

cc @hakman @ivanvc @jberkus @nwnt

ahrtr:
Fix https://github.com/etcd-io/etcd-operator/issues/362

Each etcd member maps 1-to-1 to a Pod owned directly by EtcdCluster. All startup configuration (initial cluster state, initial cluster peers, data dir) is embedded in the Pod spec's env vars, not in a shared ConfigMap. A Pod is immutable once created — if a member needs to be replaced, the old Pod (and its PVC) is deleted and a fresh Pod is created.

This is a AI-generated first version.

@nwnt do you have bandwidth to continue to finish this PR? Pls feel free to fork a new branch and continue to work on it.

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -53,11 +52,9 @@ type EtcdClusterReconciler struct {
 }
 
 // reconcileState holds all transient data for a single reconciliation loop.
-// Every phase of Reconcile stores intermediate information here so that
-// subsequent phases can operate without additional lookups.
 type reconcileState struct {
-	cluster        *ecv1alpha1.EtcdCluster      // cluster custom resource currently being reconciled
-	sts            *appsv1.StatefulSet          // associated StatefulSet for the cluster
+	cluster        *ecv1alpha1.EtcdCluster      // cluster CR being reconciled
+	pods           []*corev1.Pod                // member pods owned by this cluster, sorted by ordinal
 	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
 	memberHealth   []etcdutils.EpHealth         // health information for each etcd member
 }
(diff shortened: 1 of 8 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -208,93 +176,77 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
+			if err := createMemberPod(ctx, logger, r.Client, s.cluster, nextOrdinal, r.Scheme); err != nil {
 				return ctrl.Result{}, err
 			}
 		} else {
-			logger.Info("An etcd member was removed from the cluster, but the StatefulSet hasn't scaled in yet")
-			newReplicaCount := targetReplica - 1
-			logger.Info("Decreasing StatefulSet replicas to remove the unneeded Pod.", "oldReplicaCount", targetReplica, "newReplicaCount", newReplicaCount)
-			if _, err := reconcileStatefulSet(ctx, logger, s.cluster, r.Client, newReplicaCount, r.Scheme); err != nil {
+			// A member was removed from etcd but the pod was not yet deleted.
+			logger.Info("Deleting pod for already-removed etcd member")
```
GunaKKIBM, line 247: Apologies, how do we know, which is the pod/member which has actually failed, is relying on the last member the right approach?

nwnt:
Yes I will have a look tonight

internal/controller/etcdcluster_controller.go
ahrtr: Good point. For now, we follow the same logic as Statefulset. Eventually we will need to compare etcd member names and POD names and figure out which POD to remove.

ahrtr:
@nwnt do you have any update on this PR?

nwnt:
Yes, I'm hoping to show this what I have end-to-end during the call tomorrow.  Fixed many if not most of the issues in my local now.  There are a few items left, but I should  be ready by then.

ahrtr:
superseded by https://github.com/etcd-io/etcd-operator/pull/404

PR #367, opened 2026-06-15
update controller-gen annotations of base crd

valen-mascarenhas14:
Updating the crd base annotation as controller-tools version is updated to `v0.21.0`

https://github.com/etcd-io/etcd-operator/blob/ddb834e92a45121addf5f7b672821fb4cd67e29b/tools/mod/go.mod#L12


valen-mascarenhas14:
@ahrtr Could you PTAL ?

Issue #373, opened 2026-06-17
TLS: operator's etcd client (health checks, membership) is cleartext-only


xrl:
### Problem

The operator's own etcd client is cleartext-only. `MemberList` / `ClusterHealth` / `AddMember` / `PromoteLearner` / `RemoveMember` in `internal/etcdutils/etcdutils.go` build `clientv3.Config` from endpoints only, with no TLS. Once etcd is configured to serve TLS, the operator can no longer perform health checks or membership operations against its own cluster.

This is the third step of wiring TLS into the data path (after mounting certs and emitting TLS flags/scheme).

### Proposed change

Thread an optional `*tls.Config` through the `etcdutils` functions (nil ⇒ cleartext, preserving today's behavior; non-nil ⇒ set `clientv3.Config.TLS`). The controller builds it once per reconcile from the `{name}-client-tls` secret (already created by `createClientCertificate`) and passes `nil` whenever `spec.tls == nil`.

Single trailing `*tls.Config` parameter keeps the diff minimal and nil-default-compatible across the 5 in-tree call sites. Unit test asserts the nil arg reproduces today's cleartext config; the live TLS path is covered by an e2e follow-up. PR to follow.

Issue #372, opened 2026-06-17
TLS: etcd runs cleartext — URLs hardcoded to http:// and no TLS flags emitted


xrl:
### Problem

Even when certificates are provisioned (and mounted), etcd is started in **cleartext**:
- `defaultArgs` hardcodes `http://` on the listen/advertise/peer URLs (`internal/controller/utils.go:82-85`) and emits no TLS flags.
- the endpoint generators hardcode `http://` (`peerEndpointForOrdinalIndex` `:425`, `clientEndpointForOrdinalIndex` `:481`, `clientEndpointsFromStatefulsets` `:495`).

So `spec.tls` provisions certs that are never used by the data path.

### Proposed change

When `spec.tls != nil`: switch the URL scheme to `https://` and append `--cert-file`/`--key-file`/`--trusted-ca-file`/`--peer-cert-file`/`--peer-key-file`/`--peer-trusted-ca-file`/`--client-cert-auth`/`--peer-client-cert-auth` (pointing at the mounted paths). Crucially, derive the scheme from a **single source of truth** shared by both the etcd args and every endpoint generator, so the operator's computed endpoints can never disagree with the scheme etcd actually serves.

Unit tests assert the `tls == nil` path is byte-identical to today (no regression for existing cleartext clusters) and the `tls != nil` path emits the expected flags + `https://`.

### Note on migration

TLS is treated as a **create-time** property: this targets standing up a *new* TLS-enabled cluster, not live-flipping a running cleartext cluster (a member-by-member peer-TLS roll is a separate, harder problem). Depends on the cert-mount change (mounts must land first, or etcd crashloops referencing unmounted files). PR to follow.

Issue #371, opened 2026-06-17
TLS: provisioned server/peer cert secrets are never mounted into the etcd container


xrl:
### Problem

When `spec.tls` is set, `createOrPatchStatefulSet` adds the `server-secret` and `peer-secret` as pod **Volumes** (`internal/controller/utils.go:195-211`), but the etcd container has no corresponding **VolumeMounts** — the only mount on the container is the data dir, gated on `StorageSpec` (`utils.go:257`). The provisioned certificates are therefore never visible to the etcd process.

This is the first of a short series to wire TLS into the etcd data path (the operator currently provisions cert-manager Certificates but runs etcd in cleartext): **mount certs → emit TLS flags/scheme → give the operator's own client TLS.**

### Proposed change

Add `VolumeMount`s for the server/peer cert secrets (e.g. `/etc/etcd/server-tls/`, `/etc/etcd/peer-tls/`), gated on `spec.tls != nil`. With no TLS flags yet, mounting unused cert files is a safe no-op, so this is independently shippable. Unit test asserts the mounts are present when TLS is set and absent otherwise. PR to follow.

Issue #370, opened 2026-06-17
Operator silently swallows the client-certificate provisioning error


xrl:
### Problem

In `fetchAndValidateState` (`internal/controller/etcdcluster_controller.go:138`), a failure from `createClientCertificate` is logged and then **swallowed** — the reconcile proceeds as if the client certificate exists. When provisioning genuinely fails (issuer not ready, RBAC, transient API error), the operator continues without a usable client cert, the error never surfaces in status, and any subsequent cert-dependent operation fails downstream with a less obvious symptom.

### Proposed fix

On `createClientCertificate` failure, requeue with backoff (`ctrl.Result{RequeueAfter: ...}`) instead of logging-and-proceeding, so provisioning is retried and the cluster converges once the issuer is ready.

Small, self-contained change (~8 lines) with a unit test (fake client fails the Certificate create → assert requeue, not proceed). I'd like to send a PR for this.

PR #375, opened 2026-06-17
feat: mount server/peer cert secrets into the etcd container

xrl:
## What

Mount the already-provisioned server and peer cert secrets into the etcd container.

When `spec.tls` is set, the operator already provisions the server/peer
`Certificate`s and adds their secret `Volume`s to the StatefulSet pod spec
(`createOrPatchStatefulSet`). However, the etcd container never mounts those
volumes, so the certificates are present in the pod spec but **invisible to
etcd**. This PR adds the missing `VolumeMount`s.

## Change

- Add `VolumeMount`s on the etcd container for the server and peer cert secrets:
  - `server-secret` → `/etc/etcd/server-tls/`
  - `peer-secret` → `/etc/etcd/peer-tls/`
- Gated on `spec.tls != nil`, and appended **independently of `StorageSpec`** so
  a TLS cluster without persistent storage still receives the cert mounts.
- The data-dir mount inside the `StorageSpec != nil` branch now **appends**
  rather than assigns the `VolumeMounts` slice, so the cert mounts are preserved
  when both TLS and storage are configured.
- Volume names reuse the existing cert `Volume` names; paths are factored into
  constants so the follow-up TLS-flags change references exactly the same paths.

## Why this is safe / why now

Mounting cert files with **no TLS flags** is a no-op: etcd ignores files it is
not told to use. This keeps the change independently shippable and the cluster
always bootable. It is the first step of wiring TLS into the etcd data path; the
flags that consume these mount paths (`--cert-file`, `--peer-cert-file`, …,
plus the `https://` scheme flip) land in a separate follow-up PR so each step
stays independently reviewable.

## Migration note

The StatefulSet pod template is immutable for existing clusters, so this takes
effect for **new or recreated clusters**. TLS is a create-time property — a
running cleartext cluster should be migrated to a fresh TLS cluster rather than
flipped in place.

## Test

Extended the controller unit tests with `TestCreateOrPatchStatefulSetTLSCertMounts`,
which renders the StatefulSet through the real `createOrPatchStatefulSet` path
(controller-runtime fake client) and asserts both cert `VolumeMount`s are present
with the correct names and mount paths when `spec.tls != nil`, absent when nil,
and that the data-dir mount coexists with the cert mounts when `StorageSpec` is
also set. Verified the test fails against the unmodified code and passes after
the change.

Refs #371

---

### Related work — etcd TLS data-path enablement

The operator provisions cert-manager certificates but currently runs etcd in cleartext. This PR is one step of a small, dependency-ordered series wiring TLS through the data path (`→` marks this PR):

| | Step | Change | Issue | PR |
|--|------|--------|-------|----|
|   | T0 | stop swallowing the client-certificate error | #370 | #374 |
| → | T2 | mount server/peer cert secrets into the etcd container | #371 | #375 |
|   | T3 | emit etcd TLS flags + https endpoints (single scheme source-of-truth) | #372 | _pending_ |
|   | T4 | thread an optional `*tls.Config` through the operator's etcd client | #373 | _pending_ |
|   | — | kind stress/e2e harness (the TLS end-to-end test will extend it) | — | #369 |

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -37,6 +37,14 @@ import (
 const (
 	etcdDataDir = "/var/lib/etcd"
 	volumeName  = "etcd-data"
+
+	// TLS cert secret volume names and mount paths. The volume names must match
+	// the cert Volumes added in createOrPatchStatefulSet; the mount paths are the
+	// locations the etcd TLS flags will reference once those flags land.
+	serverCertVolumeName = "server-secret"
+	peerCertVolumeName   = "peer-secret"
+	serverCertMountPath  = "/etc/etcd/server-tls/"
+	peerCertMountPath    = "/etc/etcd/peer-tls/"
 )
 
 type etcdClusterState string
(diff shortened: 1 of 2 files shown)


xrl:
Superseding this PR. The TLS work was reshaped into independent `spec.tls.{peer,client}` surfaces (#376), and the per-surface cert **VolumeMount**s this PR added now live there — server secret mounted iff the `client` surface is set, peer secret iff `peer` is set, appended (not assigned) so they coexist with the storage data-dir mount. The standalone single-toggle approach here is subsumed by that breaking alpha API change, so closing in favor of #376. Leaving the branch in place for reference. Thanks!

Issue #381, opened 2026-06-18
Expose CR-configurable per-cluster domain metrics on the operator's Prometheus endpoint


xrl:
TLDR;

The operator already serves a controller-runtime `/metrics` endpoint, but it only exposes generic controller-runtime/Go metrics. Operators running many `EtcdCluster` resources have no first-class, per-cluster signal about the health of the etcd clusters the operator manages (member counts desired vs ready, quorum, leader changes, learner count, TLS state, reconcile timings/errors).

### Proposal

Add an `internal/metrics` package that defines and registers custom Prometheus collectors against the controller-runtime global registry, so they are served on the existing `/metrics` endpoint with no extra HTTP plumbing. Every series is labelled with the cluster `namespace` and `name`.

Proposed metrics (subsystem `etcd_operator_cluster_`):

- `member_count_desired` (gauge) — `spec.size`
- `member_count` (gauge) — members from the member-list API (`status.memberCount`)
- `member_count_ready` (gauge) — StatefulSet `readyReplicas`
- `member_count_healthy` (gauge) — healthy members from status
- `learner_count` (gauge) — non-voting members
- `has_quorum` (gauge 1/0) — healthy members form a strict majority
- `has_leader` (gauge 1/0) — a leader is known
- `tls_enabled` (gauge 1/0) — cluster configured for TLS
- `leader_changes_total` (counter) — observed leader-ID transitions
- `reconcile_duration_seconds` (histogram) — reconcile wall-clock
- `reconcile_errors_total` (counter) — reconciles returning an error

### CR configurability

Add `spec.metrics` to `EtcdCluster`:

- `spec.metrics.enabled` (`*bool`, default true) — toggle/scope the operator domain metrics per cluster; disabling drops that cluster's series.
- `spec.metrics.podMonitor` — when enabled, the operator reconciles a prometheus-operator `PodMonitor` (owned by the `EtcdCluster`) selecting the etcd member pods so Prometheus scrapes etcd's own `/metrics`. Addressed via unstructured so the operator takes no hard dependency on the prometheus-operator module and degrades gracefully (logs + skips) when the CRD is absent.

### Testing

[...]

PR #379, opened 2026-06-18
feat: configurable reconcile worker pool (--max-concurrent-reconciles)

xrl:
controller-runtime defaults to a single reconcile worker, so one slow cluster blocks progress on every other cluster. This adds a `--max-concurrent-reconciles` flag (default 5) threaded into `SetupWithManager` via `builder.WithOptions`; each `EtcdCluster` keeps its own workqueue key, so concurrency only parallelizes distinct clusters and a value `<= 0` falls back to a single worker. Documented in `docs/operator-flags.md`, with `test/e2e/STRESS.md` recording the measured spinup-burst budget behind the stress-tier batching. Tested with an envtest-backed assertion that the pool size is threaded for widened, single-worker, and non-positive values.

---

### PR series — operability fixes & TLS

Small single-purpose PRs from live kind-cluster testing of the operator. Each stands alone unless an *After* is listed. → = this PR.

| | PR | Lands | After |
|---|----|-------|-------|
| | **Reviewable now — small, any order** | | |
| 🟢 | #374 | Requeue instead of swallowing the client-cert provisioning error | — |
| 🟢 | #391 | Grant `events.k8s.io` RBAC so operator Events are actually recorded | — |
| 🟢 | #392 | Correlate `members[]`/`leaderID` from one health snapshot (consistent leader) | — |
| 🟢 | #393 | Propagate user-supplied `altNames.ipAddresses` into certificates | — |
| 🟢 | #394 | Accept day-suffix `validityDuration` (`365d`, `100d12h`) as documented | — |
| 🟢 | #395 | Surface early reconcile errors as a `Degraded` condition (was empty status) | — |
| 🟢→ | #379 | Configurable reconcile worker pool (`--max-concurrent-reconciles`) | — |
| 🟢 | #369 | kind-based stress/scale e2e harness (1/3/7 members, churn, quorum watcher) | — |
| 🟢 | #398 | `podTemplate.spec` scheduling fields: topologySpread, resources, priorityClass, schedulerName | — |
| 🟢 | #397 | First-class Helm chart covering the full `config/` surface | — |
| 🟢 | #399 | Quorum-aware PodDisruptionBudget per EtcdCluster | — |
| 🟢 | #400 | Scale-in removes the right member and transfers leadership first | — |
| 🟢 | #401 | Backend quota + auto-compaction spec fields, NOSPACE alarm surfacing | — |
| 🟢 | #402 | Reconcile-loop gates replace the blocking StatefulSet-ready wait | — |
[...]

PR #378, opened 2026-06-18
test(e2e): multi-member TLS quorum + PeerCANotShared

xrl:
## What

The headline end-to-end proof that the reshaped TLS stack (#376) works as a **real multi-member cluster**, plus its negative companion.

- **`TestTLSMultiMemberQuorum`** — stands up a 3-member `EtcdCluster` with **both** TLS surfaces (peer + client) against a two-tier shared-CA cert-manager issuer (SelfSigned → CA `Certificate` `isCA:true` → CA `ClusterIssuer`) and asserts the cluster forms and holds quorum:
  - all 3 members become voting members (no stuck learners), queried over mTLS;
  - the `EtcdCluster` reports `TLSReady=True`;
  - peer traffic is `https` (peer cert flags + https peer/client URLs on the STS);
  - an in-pod mTLS put on member-0 replicates to member-2;
  - a cleartext (`http`) attempt **and** an `https`-without-client-cert attempt both **fail**, proving TLS + client-cert-auth are genuinely enforced (not theater).
- **`TestTLSPeerCANotShared`** — a cluster whose peer surface uses a bare self-signed leaf issuer (client surface on the shared CA) must surface `TLSReady=False`/`PeerCANotShared` and hold that signal stable (never falsely flip to True). It deliberately does **not** assert non-formation: the operator mounts a **single shared peer secret** into every member, so a self-signed-leaf peer cert still forms a healthy quorum — `PeerCANotShared` is a conservative warning, not a hard barrier.

New TLS-aware e2e helpers (the shared cleartext helpers cannot talk to a client-cert-auth server): `ensureClusterCAIssuer` (webhook-retry on create), `tlsEtcdctl`, `getEtcdMemberListTLS`, `waitForNoLearnersTLS`, `waitForTLSReadyCondition`, `waitForSTSCreated`. Both tests PASS on kind (~424s wall-clock); test-only change, no controller code touched.

## Stacking

**Stacked on #377 → #376.** Review/merge after #377; asserts the `TLSReady` condition that #377 adds and exercises the reshape in #376. This branch contains #376 + #377's commits and will shrink to its own e2e test once they merge.

---

### PR series — operability fixes & TLS

Small single-purpose PRs from live kind-cluster testing of the operator. Each stands alone unless an *After* is listed. → = this PR.

| | PR | Lands | After |
|---|----|-------|-------|
| | **Reviewable now — small, any order** | | |
| 🟢 | #374 | Requeue instead of swallowing the client-cert provisioning error | — |
[...]

PR #377, opened 2026-06-18
feat(status): add TLSReady condition and TLS lifecycle events

xrl:
## What

Surface the TLS lifecycle on the `EtcdCluster` CR. The independence reshape (#376) wired the `EventRecorder` and the `buildClientTLSConfig`/`validateTLS` surfaces but emitted **zero Events** and had **no TLS status condition** — so a misconfigured TLS spec failed silently. This closes that observability gap.

- **`TLSReady` status condition** in `updateConditions`, reflecting both peer and client surfaces with a per-surface reason vocabulary: `TLSNotConfigured` (omitted when both surfaces nil), `TLSReady`, `IssuerNotFound`, `PeerCANotShared`, `ClientServerCAMismatch`, `ClientCertificateError`, `SurfaceNotReady`. False with the specific reason on each failure; True only when every configured surface is healthy.
- **CR Events** at the controller boundary (Recorder-free helpers return a typed verdict the boundary maps to Events): `Warning` on each failure with the matching reason, a single `Normal` `TLSReady` on transition into ready.
- Implements the runtime checks the reshape deferred (not CEL-expressible): **`PeerCANotShared`** detects a self-signed-leaf cert-manager peer issuer that cannot establish shared peer trust — turning a silent cap-at-1-member hazard into a visible signal; client-surface issuer existence and operator-client secret usability are checked from cluster objects. `ClientServerCAMismatch` is in the vocabulary but unreachable from a single client surface (server + operator-client share one issuer by construction) — left as a documented TODO.

Unit tests drive every state→reason mapping, the peer-CA predicate against representative issuer inputs, and Event emission via `record.FakeRecorder`. `go build`/`go vet` clean.

## Stacking

**Stacked on #376** (independent peer/client TLS). Review/merge after #376; this branch contains #376's commits and will shrink to its own two commits once #376 merges.

---

### PR series — operability fixes & TLS

Small single-purpose PRs from live kind-cluster testing of the operator. Each stands alone unless an *After* is listed. → = this PR.

| | PR | Lands | After |
|---|----|-------|-------|
| | **Reviewable now — small, any order** | | |
| 🟢 | #374 | Requeue instead of swallowing the client-cert provisioning error | — |
| 🟢 | #391 | Grant `events.k8s.io` RBAC so operator Events are actually recorded | — |
| 🟢 | #392 | Correlate `members[]`/`leaderID` from one health snapshot (consistent leader) | — |
[...]

PR #382, opened 2026-06-18
feat: CR-configurable per-cluster domain metrics on the Prometheus /metrics endpoint

xrl:
Refs #381

## What

Exposes custom **per-cluster domain metrics** for every reconciled `EtcdCluster` on the operator's existing Prometheus `/metrics` endpoint, and makes Prometheus observability **configurable via the CR** (`spec.metrics`).

The operator already serves a controller-runtime metrics endpoint backed by a global registry; this PR registers custom collectors against that same registry, so there is no new HTTP server or port. Every series is labelled with the cluster `namespace` and `name`, so a single operator managing many clusters yields one time series per cluster.

## Metrics (subsystem `etcd_operator_cluster_`)

| metric | type | meaning |
|---|---|---|
| `member_count_desired` | gauge | desired members (`spec.size`) |
| `member_count` | gauge | members from the etcd member-list API (`status.memberCount`) |
| `member_count_ready` | gauge | ready etcd pods (StatefulSet `readyReplicas`) |
| `member_count_healthy` | gauge | members reported healthy |
| `learner_count` | gauge | non-voting (learner) members |
| `has_quorum` | gauge 1/0 | healthy members form a strict majority |
| `has_leader` | gauge 1/0 | a leader is currently known |
| `tls_enabled` | gauge 1/0 | cluster configured for TLS |
| `leader_changes_total` | counter | observed leader-ID transitions |
| `reconcile_duration_seconds` | histogram | reconcile wall-clock |
| `reconcile_errors_total` | counter | reconciles returning an error |

Gauge values are derived from the **freshly-computed `EtcdClusterStatus`** at the end of `updateStatus` — no extra API calls. Reconcile duration/errors are observed in `Reconcile`. Leader changes are tracked per cluster (the first observation and leaderless windows do not count, so operator restarts don't inflate the counter). When a cluster is deleted, its series are dropped to bound cardinality.

## CR config knob (`spec.metrics`)

```yaml
spec:
  metrics:
    enabled: true            # default true; false drops this cluster's series
    podMonitor:
      enabled: true          # emit a prometheus-operator PodMonitor for the etcd member pods
      interval: "30s"
      port: "client"
```

- `spec.metrics.enabled` (`*bool`, default **true**) toggles/scopes the operator domain metrics per cluster.
- `spec.metrics.podMonitor`, when enabled, makes the operator reconcile a `PodMonitor` (owned by the `EtcdCluster`, so GC'd on delete) selecting the etcd member pods (`app`/`controller` labels) so Prometheus scrapes etcd's own `/metrics`.

### Why unstructured for the PodMonitor

The PodMonitor is created via `unstructured` so the operator takes **no hard dependency** on the prometheus-operator Go module and does not require its CRDs to be installed unless a user opts in. If the CRD is absent, the operator logs and skips PodMonitor reconciliation without failing the rest of the reconcile (`meta.IsNoMatchError`).

## Files

- **NEW** `internal/metrics/` — collector definitions, registration (`MustRegister`/`RegisterTo`), `RecordClusterMetrics`, `ObserveReconcile`, `DeleteClusterMetrics`, per-cluster leader tracker.
- **NEW** `internal/controller/podmonitor.go` — PodMonitor reconcile/cleanup.
- `api/v1alpha1/etcdcluster_types.go` — `MetricsSpec` / `PodMonitorSpec` + `MetricsEnabled()`/`PodMonitorEnabled()` helpers (CRD/deepcopy regenerated via the pinned controller-gen).
- `internal/controller/etcdcluster_controller.go` — minimal, localized instrumentation (time the loop, record metrics after status, reconcile PodMonitor, drop series on delete) + PodMonitor RBAC marker.
- `cmd/main.go` — register collectors at startup.
- **NEW** `test/e2e/metrics_test.go` — compile-verified e2e that creates a cluster with `spec.metrics.podMonitor` and scrapes the operator `/metrics`.
- `config/samples/sample_metrics_etcdcluster.yaml` — example CR.

## Testing

- Unit tests (`internal/metrics/metrics_test.go`) register collectors to an **isolated** registry, drive cluster state, scrape via `prometheus/client_golang/testutil`, and assert exact values + labels — including quorum/leader edge cases, leader-change counting across leaderless windows, reconcile error counting, and series deletion (and leader-tracker reset) on cluster delete. Spec-helper defaults are covered too.
- `go build ./...`, `go vet ./...`, `go vet ./test/e2e/...`, and the controller envtest suite all pass.
- No live KinD run in this PR; the e2e test is authored to compile in CI and run under `make test-e2e`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

| | PR | Lands | After |
|---|----|-------|-------|
| 🟢 | #393 | Propagate user-supplied `altNames.ipAddresses` into certificates | — |
| ⚪→ | #382 | Per-cluster domain metrics on the operator `/metrics` endpoint | — |
| ⚪ | #384 | EtcdCluster admission webhooks (consolidating with #328) | #363 |

🟢 ready · ⚪ draft


xrl:
Converting to draft to keep the review queue focused while #376 and the StatefulSet rework (#363) are in flight.

Issue #380, opened 2026-06-18
EtcdCluster admission webhooks are scaffolded but never registered


xrl:
## Summary

The operator wires a `webhook.NewServer(...)` into the manager in `cmd/main.go`, and the kustomize scaffolding for `config/webhook` + `config/certmanager` markers exists in `config/default/kustomization.yaml` and `config/crd/kustomization.yaml`. However, **no admission webhook is actually implemented or registered** for `EtcdCluster`. The webhook server starts but serves nothing.

As a result, malformed `EtcdCluster` specs are only caught (if at all) later during reconciliation — e.g. an even `size` (which can never form a quorum), an unparseable `version`, an unsupported skip-minor upgrade/downgrade, or an incoherent TLS surface. The user gets no synchronous, actionable feedback at `kubectl apply` time.

Today the controller logs upgrade-path problems during reconcile (`validateEtcdUpgradePath` in `internal/controller/utils.go`) but still proceeds; there is no admission-time gate.

## Proposal

Implement and **register** admission webhooks for `EtcdCluster`:

**Validating** (`ValidateCreate` / `ValidateUpdate`):
- Reject even `size` (etcd needs an odd member count for a stable quorum).
- Reject `size < 1`.
- Validate `version` parses as semver and (on update) that the current→target transition is a supported single-minor, non-downgrade path — reusing the existing upgrade-path logic.
- Validate TLS surface coherence (provider must be one of the known providers; the matching provider config block must be present; cert-manager provider requires `issuerKind`/`issuerName`).
- Immutable-field guards (e.g. `storageSpec` is immutable once set).

**Defaulting** (`Default`):
- Default an empty `tls.provider` to `auto` when a TLS block is present.

Every rejection must carry a crisp, actionable message telling the user exactly what is wrong and how to fix it.

Wire `SetupWebhookWithManager` in `cmd/main.go`, regenerate `config/webhook` + `config/certmanager` via `make manifests`, and add unit + envtest + e2e coverage (including assertions on the exact error-message text).

I'll open a PR referencing this issue.

PR #328, opened 2026-04-05
Validation admission policy

GunaKKIBM:
https://github.com/etcd-io/etcd-operator/issues/326

Issue #326, opened 2026-03-31
Add Validation Admission Policy / Validation admission webhook to avoid updates/delete on secondary resources


GunaKKIBM:
Currently, any delete/update on the secondary resources by the user - statefulset/configmap/service is allowed. Add a VAP to avoid the same

PR #384, opened 2026-06-18
feat: implement and register EtcdCluster admission webhooks

xrl:
## Summary

The manager already constructs a webhook server (`webhook.NewServer(...)` in `cmd/main.go`) and the kustomize scaffolding carries `[WEBHOOK]`/`[CERTMANAGER]` markers, but **no admission webhook was ever implemented or registered** for `EtcdCluster`. The server started and served nothing, so malformed specs were only (partially) caught later during reconciliation, with no synchronous, actionable feedback at `kubectl apply` time.

This PR implements and **registers** a validating + defaulting admission webhook for `EtcdCluster`.

Refs #380

## What it does

`SetupWebhookWithManager` is called from `cmd/main.go` (a minimal, localized edit, skippable with `ENABLE_WEBHOOKS=false` for local `make run` without serving certs).

### Validating (`ValidateCreate` / `ValidateUpdate`)
Every rejection returns a crisp, actionable message telling the user exactly what is wrong and how to fix it:

- **`spec.size`** — must be `>= 1` and **odd**. An even-sized etcd cluster tolerates no more failures than the next-smaller odd size while being more likely to lose quorum; the message names the two nearest valid sizes.
- **`spec.version`** — must be valid semver. On **update**, the current→target transition must be a supported **single-minor, non-downgrade** path, reusing the same ordered version table (`etcd/api/v3/version.AllVersions`) the controller uses at reconcile time.
- **`spec.tls`** — `provider` must be one of `{auto, cert-manager}`; the provider-specific config block must be present/complete (cert-manager requires `issuerName`, and `issuerKind` if set must be `Issuer`/`ClusterIssuer`); the `auto` provider must not carry cert-manager config.
- **`spec.storageSpec`** — immutable once set (changing/removing the PVC template after the StatefulSet exists cannot be reconciled in place).

### Defaulting (`Default`)
- `spec.tls.provider` defaults to `auto` when a TLS block is present without an explicit provider (matches the documented behaviour of `TLSCertificate.Provider`).

### Sample rejection messages
```
spec.size: Invalid value: 4: size must be an odd number so the cluster can form a
  majority quorum; got 4. An even-sized etcd cluster tolerates no more failures than
  the next-smaller odd size while being more likely to lose quorum. Use 3 or 5 instead.

spec.version: Invalid value: "latest": version "latest" is not a valid semantic version
  (expected MAJOR.MINOR.PATCH, e.g. "3.6.1"): latest is not in dotted-tri format.

spec.version: Invalid value: "3.7.1": upgrading from version 3.5.1 to version 3.7.1 is
  not allowed (skips a minor version). etcd only supports sequential single-minor
  upgrades and forbids downgrades; upgrade one minor version at a time
  (current "3.5.1" -> target "3.7.1").

spec.tls.provider: Unsupported value: "vault": supported values: "auto", "cert-manager"
```

### Logging
The webhook logs every admission decision (admit / default / reject) with structured context — object name/namespace, operation, and per-field reason/remediation — so an operator reading the webhook-server logs can reconstruct exactly why any request was accepted or rejected.

## Manifests
`make manifests` generates `config/webhook/manifests.yaml`; this PR adds the standard `config/webhook` (service + kustomizeconfig) and `config/certmanager` (self-signed Issuer + serving `Certificate`) scaffolds, the `manager_webhook_patch.yaml`, and enables the `[WEBHOOK]`/`[CERTMANAGER]` sections in `config/default/kustomization.yaml` including cert-manager CA injection. `kustomize build config/default` resolves the service name/namespace and `cert-manager.io/inject-ca-from` annotations correctly.

## Testing
- **Unit** (`api/v1alpha1/etcdcluster_webhook_test.go`): asserts the **exact** rejection-message text for size/version/TLS/upgrade/immutability and the defaulting behaviour.
- **Envtest** (`api/v1alpha1/webhook_envtest_test.go`): installs the generated webhook config, runs a manager serving the webhook over the envtest-provisioned cert, applies invalid `EtcdCluster`s against a real API server and asserts rejection + message; confirms defaulting and a valid single-minor upgrade are admitted.
- **E2E** (`test/e2e/webhook_test.go`, new file): live apply-invalid assertions via the e2e-framework feature harness.

`go build ./...`, `go vet ./...`, `golangci-lint`, the unit tests, and the envtest webhook tests all pass; the e2e file compiles (`go test -c ./test/e2e/`). The e2e file is self-contained (its own helper) to avoid conflicts with other in-flight e2e PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

| | PR | Lands | After |
|---|----|-------|-------|
| 🟢 | #393 | Propagate user-supplied `altNames.ipAddresses` into certificates | — |
| ⚪ | #382 | Per-cluster domain metrics on the operator `/metrics` endpoint | — |
| ⚪→ | #384 | EtcdCluster admission webhooks (consolidating with #328) | #363 |

🟢 ready · ⚪ draft

### DIFF
--- a/api/v1alpha1/etcdcluster_webhook.go
+++ b/api/v1alpha1/etcdcluster_webhook.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/coreos/go-semver/semver"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -140,8 +141,7 @@ func (v *EtcdClusterCustomValidator) ValidateUpdate(
 	log := etcdclusterlog.WithValues("name", ec.Name, "namespace", ec.Namespace, "operation", "update")
 	log.Info("validating EtcdCluster on update")
 
-	errs := validateEtcdClusterSpec(&ec.Spec)
-	errs = append(errs, validateEtcdClusterUpdate(&oldEc.Spec, &ec.Spec)...)
+	errs := validateEtcdClusterUpdate(&oldEc.Spec, &ec.Spec)
 	return finish(log, ec, errs)
 }
 
@@ -175,7 +175,8 @@ func finish(log logr, ec *EtcdCluster, errs field.ErrorList) (admission.Warnings
 // exact message text.
 // ---------------------------------------------------------------------------
 
-// validateEtcdClusterSpec runs all create-time (and update-time) spec invariants.
+// validateEtcdClusterSpec runs all create-time spec invariants. On update only the
+// subset of these that changed is re-checked; see validateEtcdClusterUpdate.
 func validateEtcdClusterSpec(spec *EtcdClusterSpec) field.ErrorList {
 	errs := make(field.ErrorList, 0, 3)
 	specPath := field.NewPath("spec")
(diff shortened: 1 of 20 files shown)


xrl:
Converting to draft — this overlaps #326 and the ValidatingAdmissionPolicy approach in #328. Happy to consolidate with that effort rather than run a competing implementation.

PR #387, opened 2026-06-18
feat: automatic quorum-loss disaster recovery

xrl:
Detects sustained quorum loss on an `EtcdCluster` (majority of members unreachable / no electable leader for a sustained window) and automatically recovers by rebuilding from a surviving member (snapshot-restore / `--force-new-cluster`), then re-adds members one at a time via the existing learner path. Guarded to trigger only on true quorum loss (not transient single-member failures), idempotent, with a `Recovering` status Condition + Events and detailed logging.

Includes unit tests for the detection + recovery state transitions, and a self-contained e2e (`test/e2e/quorum_recovery_test.go`) that breaks a 3-node cluster and asserts recovery + data integrity. The e2e is authored and compile-verified; the live kind run is still pending (Docker was unavailable in the build environment).

Refs #383

---

### PR series — operability fixes & TLS

Small single-purpose PRs from live kind-cluster testing of the operator. Each stands alone unless an *After* is listed. → = this PR.

| | PR | Lands | After |
|---|----|-------|-------|
| | **Reviewable now — small, any order** | | |
| 🟢 | #374 | Requeue instead of swallowing the client-cert provisioning error | — |
| 🟢 | #391 | Grant `events.k8s.io` RBAC so operator Events are actually recorded | — |
| 🟢 | #392 | Correlate `members[]`/`leaderID` from one health snapshot (consistent leader) | — |
| 🟢 | #393 | Propagate user-supplied `altNames.ipAddresses` into certificates | — |
| 🟢 | #394 | Accept day-suffix `validityDuration` (`365d`, `100d12h`) as documented | — |
| 🟢 | #395 | Surface early reconcile errors as a `Degraded` condition (was empty status) | — |
| 🟢 | #379 | Configurable reconcile worker pool (`--max-concurrent-reconciles`) | — |
| 🟢 | #369 | kind-based stress/scale e2e harness (1/3/7 members, churn, quorum watcher) | — |
| 🟢 | #398 | `podTemplate.spec` scheduling fields: topologySpread, resources, priorityClass, schedulerName | — |
| 🟢 | #397 | First-class Helm chart covering the full `config/` surface | — |
| 🟢 | #399 | Quorum-aware PodDisruptionBudget per EtcdCluster | — |
| 🟢 | #400 | Scale-in removes the right member and transfers leadership first | — |
| 🟢 | #401 | Backend quota + auto-compaction spec fields, NOSPACE alarm surfacing | — |
[...]

PR #386, opened 2026-06-18
feat: EtcdBackup CR — on-demand snapshots to object storage (S3/GCS)

xrl:
Adds an `EtcdBackup` custom resource for on-demand etcd snapshots to object storage.

- `EtcdBackup` CRD + reconciler: exec `etcdctl snapshot save` in a member pod, stream via `io.Pipe` (bounded memory) to an object store.
- Cloud SDKs isolated behind a `pkg/objectstore.Store` factory registry (S3 + S3-compatible, GCS); the controller imports no cloud SDK directly.
- Credentials ambient (IRSA / Workload Identity) or optional `secretRef`.
- Shared object-store layer: `pkg/objectstore/*` and `internal/controller/objectstore_creds.go` (`resolveStoreCredentials` / `validateDestination`).

Note: `objectstore.Store.Download()` / `ErrNotFound` are intentionally present on the interface for the stacked follow-up.

**EtcdRestore was split into a stacked follow-up PR** so this PR is backup-only and independently reviewable/mergeable.

PR #393, opened 2026-07-02
fix: propagate user-supplied altNames.ipAddresses to certificates

xrl:
Setting spec.tls.*.providerCfg.*.altNames.ipAddresses had no effect: the controller filled AltNames.IPs with nil placeholders and never read the user-supplied field, so issued certificates carried no IP SANs on either provider. The controller now passes user IPs through on both the custom and default DNS-name paths (rejecting nil/empty entries with an explicit error), the cert-manager provider builds Certificate.Spec.IPAddresses from ip.String() instead of fmt.Sprint, and the auto provider appends the IPs to the transport.SelfCert hosts list so they land in the issued certificate's IP SANs. Verified with go build ./... and go test ./internal/... ./api/... ./pkg/...; new tests assert the emitted Certificate spec (cert-manager) and the parsed x509 IP SANs (auto), and were confirmed to fail against the pre-fix code.

---

| | PR | Lands | After |
|---|----|-------|-------|
| 🟢→ | #393 | Propagate user-supplied `altNames.ipAddresses` into certificates | — |
| ⚪ | #382 | Per-cluster domain metrics on the operator `/metrics` endpoint | — |
| ⚪ | #384 | EtcdCluster admission webhooks (consolidating with #328) | #363 |

🟢 ready · ⚪ draft

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -590,6 +590,17 @@ func parseValidityDuration(customizedDuration string, defaultDuration time.Durat
 	return duration, nil
 }
 
+// validateAltNameIPs rejects nil/empty entries so that invalid IP addresses
+// never reach certificate providers.
+func validateAltNameIPs(ips []net.IP) error {
+	for i, ip := range ips {
+		if len(ip) == 0 {
+			return fmt.Errorf("altNames.ipAddresses[%d] is not a valid IP address", i)
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


xrl:
Pushed 0f5b174 to fix the golangci-lint prealloc failure: the ipStrings slice in the cert-manager provider is now preallocated with capacity len(cfg.AltNames.IPs).

PR #404, opened 2026-07-07
Replace StatefulSets with a customized controller

nwnt:
Based on the original [PR](https://github.com/etcd-io/etcd-operator/pull/363#issuecomment-4896827597)

I have tested all the changes in my local and they passed.  Will be covering the additional required changes on top of the original PR in the call today.  

@ahrtr @ivanvc

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -53,11 +52,9 @@ type EtcdClusterReconciler struct {
 }
 
 // reconcileState holds all transient data for a single reconciliation loop.
-// Every phase of Reconcile stores intermediate information here so that
-// subsequent phases can operate without additional lookups.
 type reconcileState struct {
-	cluster        *ecv1alpha1.EtcdCluster      // cluster custom resource currently being reconciled
-	sts            *appsv1.StatefulSet          // associated StatefulSet for the cluster
+	cluster        *ecv1alpha1.EtcdCluster      // cluster CR being reconciled
+	pods           []*corev1.Pod                // member pods owned by this cluster, sorted by ordinal
 	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
 	memberHealth   []etcdutils.EpHealth         // health information for each etcd member
 }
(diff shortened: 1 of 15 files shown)


nwnt:
The diff between this PR and the original one can be found [here ](https://github.com/etcd-io/etcd-operator/pull/404/changes#diff-72dd89d095e0c9a6d5e10dfb59c8cebc958b7426).

internal/controller/etcdcluster_controller.go
```diff
@@ -208,93 +176,82 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
 
-	case s.sts.Spec.Replicas != nil && *s.sts.Spec.Replicas == 0:
-		logger.Info("StatefulSet has 0 replicas. Trying to create a new cluster with 1 member")
-		s.sts, err = reconcileStatefulSet(ctx, logger, s.cluster, r.Client, 1, r.Scheme)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		requeue = true
+	// Service must exist before pods start so that headless DNS resolves.
+	if err := createHeadlessServiceIfNotExist(ctx, logger, r.Client, s.cluster, r.Scheme); err != nil {
```
ahrtr, line 193: Do we still need headless service after removing statefulSet? Or is it just an improper function name?

internal/controller/utils.go
```diff
@@ -46,296 +46,287 @@ const (
 	etcdClusterStateExisting etcdClusterState = "existing"
 )
 
-func reconcileStatefulSet(ctx context.Context, logger logr.Logger, ec *ecv1alpha1.EtcdCluster, c client.Client, replicas int32, scheme *runtime.Scheme) (*appsv1.StatefulSet, error) {
+// memberPodName returns the deterministic name for an etcd member pod.
+// The naming convention mirrors StatefulSet so that headless-service DNS is identical.
+func memberPodName(clusterName string, ordinal int) string {
```
ahrtr: The `internal/controller/utils.go` is a little overloaded. I think we need to extract all POD related stuff into a separate file, or even separate package. If we move it into separate package, we need to do some abstraction to ensure it's business agnostic. 

in the first step, probably we can just move them into a separate file, and consider refactoring it in next step if possible.

internal/controller/etcdcluster_controller.go
ahrtr: We still need headless service, regardless of statefulset being used or not.

nwnt: Yes, was about to reply that we need it for the FQDN `hostname.subdomain.namespace.svc.cluster.local` for each pod.

ahrtr:

internal/controller/utils.go
nwnt: Done, move everything related into `pods.go`.  I have also moved the related tests to `pods_test.go`.

ahrtr:
It seems that the pipelines are not triggered if you raise PR against customized-controller. Let's still raise PR against main, once it's reviewed I will manually merge the PR to customized-controller.

ahrtr:
Or raise a PR to add customized-controller into https://github.com/kubernetes/test-infra/blob/master/config/jobs/etcd/etcd-operator-presubmits.yaml

nwnt:
Both are done.

internal/controller/pods.go
```diff
@@ -0,0 +1,377 @@
...
+	if err != nil {
+		return -1
+	}
+	return ordinal
+}
+
+func nextPodOrdinal(currentOrdinals []int, expectedReplica int) int {
+	sort.Ints(currentOrdinals)
+	for i := range expectedReplica {
+		if i >= len(currentOrdinals) || i != currentOrdinals[i] {
```
ahrtr, line 65: What's the case for the `i != currentOrdinals[i] ` ?

ahrtr (commented):
cc @GunaKKIBM PTAL

ahrtr:
Let's try to get this PR merged later today. Overall looks good.

We can consider some refactoring later when this is completely done.

internal/controller/pods.go
nwnt: It's for filling the gap if one is found.  For example, if the current ordinals are {0,1,3}, when `i` is 2 it will see that `currentOrdinals[i]` is 3, so 2 is the gap and hence the next ordinal to be used.

nwnt: the current ordinal slice is always sorted before entering this loop.

ahrtr: > if the current ordinals are {0,1,3}

In which case we may run into this situation? If no, we'd better remove it for better readability

nwnt: Yes, we do.  I added it because the logic from the original PR would fail one of the e2e tests as described [here](https://docs.google.com/document/d/1KaNYN_-nUMln2eiOrA808N63OBY-iYp-3l6fAg1plDA/edit?tab=t.0#heading=h.vgdns7iox97b).

Here is the code from the original PR.

```go
			logger.Info("Creating pod for already-registered etcd member")
			nextOrdinal := int(currentPodCount)
```

Supposed that we have `etcd-0`, `etcd-1`, and `etcd-2`, if `etcd-1` is deleted, the reconciliation loop logic would propose that `etcd-2` should be created because the count is 2.  The added function instead identifies the gap, and proposes that the first ordinal in the gap should be used.  In that same scenario, `1` would be the next ordinal compared to `2` in the original PR.

ahrtr: We might need to revisit this later. There are three variables that we need to consider to decide the ordinal,
- EtcdCluster.Spec.Size
- etcd member count
- real POD count

We need to differentiate two different cases
- one POD is somehow deleted or crashed
- normal scale in/out case, or upgrade case

nwnt: Yes, I agree.  I came up with this just to pass the existing e2e tests and didn't mean it to be an ultimate solution for the problem.  Your cases are something we want to provide as a guarantee that etcd-operator will behave in a certain way.

ahrtr:
FYI. Before I manually merge each PR into customized-controller, I will rebase it based on main first, to ensure there is no any conflict between `customized-controller` and `main`.

ahrtr:
@nwnt I just manually merge this PR into branch `customized-controller`. Also raised a comment https://github.com/kubernetes/test-infra/pull/37411#issuecomment-4918756119 PTAL

Issue #357, opened 2026-06-06
Refactoring and incremental changes to Etcd Cluster Status reporting


GunaKKIBM:
https://github.com/etcd-io/etcd-operator/pull/350/ - Additional refactoring and unaddressed comments in this PR needs to taken up in this issue.

cc: @ahrtr @ivanvc

PR #405, opened 2026-07-07
Modify updateConditions to use linearizable Get

cmallikarjunah:
Ref Issue: https://github.com/etcd-io/etcd-operator/issues/357

NOTE:
Added a separate function IsClusterAvailable() instead of modifying ClusterHealth(). ClusterHealth uses Serializable Get and is needed to check health against each member by other methods.  
IsClusterAvailable issues a linearizable Get against each provided endpoint in turn and returns true as soon as any one responds successfully.

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -481,24 +481,21 @@ func (r *EtcdClusterReconciler) updateConditions(s *reconcileState) {
 		ObservedGeneration: s.cluster.Generation,
 		LastTransitionTime: now,
 		Reason:             "ClusterNotReady",
-		Message:            "Etcd cluster is not yet available",
+		Message:            "Cluster is not yet available",
 	}
 
-	if s.memberListResp != nil && len(s.memberListResp.Members) > 0 {
-		healthyCount := 0
-		for _, health := range s.memberHealth {
-			if health.Health {
-				healthyCount++
-			}
-		}
-
-		quorum := (len(s.memberListResp.Members) / 2) + 1
-		if healthyCount >= quorum {
+	if s.sts != nil && s.memberListResp != nil && len(s.memberListResp.Members) > 0 {
+		// Issue a linearizable read against the cluster endpoints. A
+		// linearizable Get requires a Raft leader and quorum acknowledgement,
+		// so a successful response is sufficient proof that the cluster is
+		// available — no manual (N/2)+1 arithmetic needed.
+		eps := clientEndpointsFromStatefulsets(s.sts)
+		if etcdutils.IsClusterAvailable(eps) {
 			availableCondition.Status = metav1.ConditionTrue
 			availableCondition.Reason = "ClusterAvailable"
-			availableCondition.Message = fmt.Sprintf("Etcd cluster has %d/%d healthy members with quorum", healthyCount, len(s.memberListResp.Members))
+			availableCondition.Message = fmt.Sprintf("Cluster is available: linearizable read succeeded (%d members)", len(s.memberListResp.Members))
 		} else {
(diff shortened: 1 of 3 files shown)


internal/etcdutils/etcdutils.go
```diff
@@ -200,6 +200,55 @@ func ClusterHealth(eps []string) ([]EpHealth, error) {
...
+		cli, err := clientv3.New(cfg)
+		if err != nil {
+			continue
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		_, err = cli.Get(ctx, "health")
+		cancel()
+		_ = cli.Close()
+
+		if err == nil || errors.Is(err, rpctypes.ErrPermissionDenied) {
```
GunaKKIBM, line 236: It's better if this part of the code is reviewed by @ahrtr . Could you please check if this is the right way to check the cluster health?

internal/controller/etcdcluster_controller.go
```diff
@@ -481,24 +481,21 @@ func (r *EtcdClusterReconciler) updateConditions(s *reconcileState) {
...
+		// available — no manual (N/2)+1 arithmetic needed.
+		eps := clientEndpointsFromStatefulsets(s.sts)
+		if etcdutils.IsClusterAvailable(eps) {
 			availableCondition.Status = metav1.ConditionTrue
 			availableCondition.Reason = "ClusterAvailable"
-			availableCondition.Message = fmt.Sprintf("Etcd cluster has %d/%d healthy members with quorum", healthyCount, len(s.memberListResp.Members))
+			availableCondition.Message = fmt.Sprintf("Cluster is available: linearizable read succeeded (%d members)", len(s.memberListResp.Members))
 		} else {
-			availableCondition.Message = fmt.Sprintf("Etcd cluster has %d/%d healthy members, quorum requires %d", healthyCount, len(s.memberListResp.Members), quorum)
+			availableCondition.Message = fmt.Sprintf("Cluster is unavailable: no member responded to a linearizable read (%d members)", len(s.memberListResp.Members))
```
GunaKKIBM: I think it would be good to keep how many PRs out of all are available. With this change only total members is being published as a part of the statues. 

Probably, if the IsClusterAvailable function can return that healthycount, that could be better.

internal/etcdutils/etcdutils.go
```diff
@@ -200,6 +200,55 @@ func ClusterHealth(eps []string) ([]EpHealth, error) {
...
+			// Linearizable read succeeded (or auth denied but quorum confirmed).
+			return true
+		}
+
+		// No leader elected or leader not yet established, so fail fast.
+		if errors.Is(err, rpctypes.ErrNoLeader) || errors.Is(err, rpctypes.ErrNotLeader) {
```
GunaKKIBM, line 242: If the cluster logic is right, it is necessary to return the error or log it here. Please take care of the same.

internal/controller/etcdcluster_controller.go
cmallikarjunah: @GunaKKIBM Have made the changes. Given that ClusterHealth() runs anyway and its s.memberHealth count is already computed, so I'm reusing s.memberHealth to log the healthy member count.

internal/etcdutils/etcdutils.go
cmallikarjunah: I'm handling `rpctypes.ErrPermissionDenied` error type for an unauthenticated cli.Get() use case.
Basically in an auth-enabled etcd cluster every client request should pass credentials in its config. In case of the cli.Get() the member did process the linearizable read all the way through Raft — it contacted the leader, got quorum acknowledgement, applied the read index — but then rejected the request at the authorization layer because the client doesn't have permission to read the "health" key. 

Should I keep this use case handled here?
@ahrtr @GunaKKIBM

PR #421, opened 2026-07-13
fix: healthCheck returns error if MemberList fails

silentred:
MemberList may fail if DNS of headless service is not ready yet. This may cause controller to delete pod-0.

/cc @ahrtr @nwnt


ahrtr (approved):
Good catch. thx

Issue #411, opened 2026-07-07
Support config update


ahrtr:
We depend on the config hash to check whether there is any change on the config.

The related code is [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L434).

silentred:
The `etcdOptions` are applied only when the Pod is initially created. Subsequent updates to `etcdOptions` will not affect existing Pods.

```
# EtcdCluster spec looks like:
spec:
  version: v3.5.21
  size: 3
  etcdOptions:
  - "--enable-pprof"

# pod args include "--enable-pprof"
kubectl get pod etcdcluster-sample-0 -oyaml | grep args -C10
spec:
  containers:
  - args:
    - --name=$(POD_NAME)
    - --listen-peer-urls=http://0.0.0.0:2380
    - --listen-client-urls=http://0.0.0.0:2379
    - --initial-advertise-peer-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2380
    - --advertise-client-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2379
    - --enable-pprof
    command:
    - /usr/local/bin/etcd
```

ahrtr:
> Subsequent updates to `etcdOptions` will not affect existing Pods.

Does the existing 0.2.0 support this?

nwnt:
It does, yes.  The reason for the regression is the new custom controller doesn't reconcile the content of the pod like the controller manager does for StatefulSets.  

I was thinking about this yesterday, and I think we found another important reason why a custom controller is needed is we can control the order of pod rerolling.  Without that, in the worst case scenario, updating the pod's definition can end up making the target etcd cluster kill the leader n times, where n is the number of members in the cluster.

nwnt:
assign @nwnt

silentred:
> updating the pod's definition can end up making the target etcd cluster kill the leader n times, where n is the number of members in the cluster

See your point. With the feature [Partitioned rolling updates](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions) of Statefulset, the solution could be as follow:
1. imagine a 3-node cluster, having pod-0, pod-1, pod-2
2. check leader's ordinal, let's say it is 1
[...]

PR #422, opened 2026-07-13
Add hashing to etcd pod

nwnt:
Ref: #411

The original migration PR doesn't implement the drift detection like the StatefulSets controlller does.  This is needed so that each reconciliation loop can then observe the drift and act accordingly by rolling the pods.

Remaining items to be implemented (likely in separate PRs):
  - Controller to update the pod metadata if the metadata drift is detected.  The pod metadata drift should not require a pod reroll.
  - Implement controller logic to roll the etcd member pods when a drift is detected based on the calculated hash.

cc @ahrtr @silentred

### DIFF
--- a/internal/controller/etcdcluster.go
+++ b/internal/controller/etcdcluster.go
@@ -1,0 +1,30 @@
+package controller
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+
+	"k8s.io/utils/dump"
+
+	"go.etcd.io/etcd-operator/api/v1alpha1"
+)
+
+const (
+	hashLength = 12
+)
+
+func EtcdClusterHash(ec *v1alpha1.EtcdCluster) string {
+	clusterSpec := ec.DeepCopy().Spec
+	// size is not used for calculating the hash
+	clusterSpec.Size = 0
+	clusterSpec.EtcdOptions = createArgs(ec.Name, ec.Spec.EtcdOptions)
+	// we don't want to roll etcd pods when metadata is changed.
+	// instead, we'd do in-place update for them.
+	if clusterSpec.PodTemplate != nil {
+		clusterSpec.PodTemplate.Metadata = nil
+	}
+	deterministicString := dump.ForHash(clusterSpec)
+	hasher := sha256.New()
+	hasher.Write([]byte(deterministicString))
+	return hex.EncodeToString(hasher.Sum(nil))[:hashLength]
+}
--- a/internal/controller/pods.go
+++ b/internal/controller/pods.go
@@ -218,6 +218,10 @@ func defaultArgs(name string) []string {
 	}
 }
 
+const (
+	HashMetadataKey = "operator.etcd.io/spec-hash"
+)
+
 // buildMemberPod constructs the Pod object for a single etcd member.
 func buildMemberPod(ec *ecv1alpha1.EtcdCluster, podName string, initialClusterState etcdClusterState, initialCluster string) *corev1.Pod {
 	// Start with custom labels then overwrite with the mandatory defaults so
@@ -228,12 +232,13 @@ func buildMemberPod(ec *ecv1alpha1.EtcdCluster, podName string, initialClusterSt
 	}
 	maps.Copy(labels, etcdClusterLabels(ec))
 
-	// Annotations are purely from the PodTemplate; nil when not provided.
-	var annotations map[string]string
+	// Apply annotations from EtcdCluster and add additional annnotations required by etcd operator
+	annotations := make(map[string]string)
 	if ec.Spec.PodTemplate != nil && ec.Spec.PodTemplate.Metadata != nil &&
 		len(ec.Spec.PodTemplate.Metadata.Annotations) > 0 {
-		annotations = ec.Spec.PodTemplate.Metadata.Annotations
+		maps.Copy(annotations, ec.Spec.PodTemplate.Metadata.Annotations)
 	}
+	annotations[HashMetadataKey] = EtcdClusterHash(ec)
 
 	envVars := []corev1.EnvVar{
 		{
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -27,6 +27,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
 	"sigs.k8s.io/e2e-framework/klient/wait"
@@ -35,6 +36,7 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/features"
 
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+	"go.etcd.io/etcd-operator/internal/controller"
 )
 
 var etcdVersion = os.Getenv("ETCD_VERSION")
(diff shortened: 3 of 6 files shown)


internal/controller/pods.go
```diff
@@ -234,6 +238,7 @@ func buildMemberPod(ec *ecv1alpha1.EtcdCluster, podName string, initialClusterSt
 		len(ec.Spec.PodTemplate.Metadata.Annotations) > 0 {
 		annotations = ec.Spec.PodTemplate.Metadata.Annotations
 	}
+	annotations[HashMetadataKey] = EtcdClusterHash(ec)
```
ahrtr, line 241: How will you use this annotation?

ahrtr:
Can you explain why adding a hash annotation is a precondition of this change?

When users update the CR EtcdCluster, then the controller will get notified and executed automatically. Not sure why adding an annotation.

nwnt:
This is the same mechanism employed by the controller manager to detect the configuration drifts in sts or replicaset. Basically, we don't want to roll the pods for all the changes such as status or metadata updates. It's quite hard to reconcile the running pod configuration vs the state of EtcdCluster CR at afterwards. I think it'd also be more deterministic to calculate the hash from the CR at the time the pod is created and check with any future updates to the CR. The drift can then be detected by comparing the hash annotated to the pods and the hash calculated from the controller is notified of a potential related change.

internal/controller/etcdcluster.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const (
+	hashLength = 12
+)
+
+func EtcdClusterHash(ec *v1alpha1.EtcdCluster) string {
+	cluster := ec.DeepCopy().Spec
+	// size is not used for calculating the hash
+	cluster.Size = 0
```
ahrtr: I didn't notice that you only calculate the hash of spec. It's good. Pls rename cluster to `clusterSpec`.

Why size is not used for calculating hash?

internal/controller/etcdcluster.go
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2026.
```
ahrtr: I see that some files have copyright header, some not. Also the copy right header isn't consistent with etcd's main repo.

Suggest that we don't add it for now. We can add them separate in one PR if needed in future.

internal/controller/pods.go
nwnt: Here is the flow I thought about over the weekend:  the controller list the pods with the label `app=etcdcluster.Name` to get the existing pods.  Then within that list, it checks how many of them have the same hash as the one calculated from the current etcdcluster CR.  

And here is what I think the current reconliation loop should be adjusted:

1. The first check for pod count vs member count can be adjusted to so that
  - if the member count is higher, we can add the new pod in like before.
  - if the member count is lower, we can remove a pod with the mismatched hash first.
  - if the member count is the same, but some pods have an incorrect hash, they need to be deleted, which will then be filled by the next reconciliation loop.
  - If the member count is the same and all pods have the matching hash, continue to the next part.
2. Wait for the leaners to be promoted.

Apart from more logic that needs to be added so that leader is rerolled last and minimizing leader moves, I think the rest of the reconciliation logic would stay the same.

internal/controller/etcdcluster.go
nwnt: I think the size should only be used by the reconciliation logic to determine if pod should be added or removed, but not as part of the 'fingerprint' validation.  If we included the size in the hash calculation, scaling in or out would mean the existing pods are always invalidated, hence needs to be rerolled, which is not what we want.  

k8s `controller-manager` doesn't include `replicas` of ReplicaSet or StatefulSet in the hash calculation either.

nwnt: Done, renamed to `clusterSpec`.

nwnt:
Example of the annotation:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    operator.etcd.io/spec-hash: e477d3fc9b05
  creationTimestamp: "2026-07-13T22:56:40Z"
  labels:
    app: etcd-cluster-auto-cert
    controller: etcd-cluster-auto-cert
  name: etcd-cluster-auto-cert-0
  namespace: etcd-operator-system
```

internal/controller/pods.go
silentred: > How will you use this annotation?

Can we calculate the hash of both EtcdCluster and Pod in runtime, for reducing the overhead of maintaining a new annotation?

> the current reconliation loop should be adjusted

I have not think it through, but my intuition is that 
1. Under any circumstances, the quorum must be prioritized for protection. So the member count(or replica count?) should be satisfied first.
2. With quorum hold, the mismatched pod should be deleted, one pod at a time. Then next reconciliation will refill the member pods.

ahrtr: > Can we calculate the hash of both EtcdCluster and Pod in runtime, for reducing the overhead of maintaining a new annotation?

Calculating `EtcdCluster`'s hash at runtime should be OK.  I think calculating Pod's hash right before creating the Pod is a good timing as long as we treat Pod as an immutable resource. 

> And here is what I think the current reconliation loop should be adjusted:

I feel like that we should generate a new high level workflow for v0.3.0 (I will try to deliver a PR this week) based on [reconcile_loop_v0.1.0.png](https://github.com/etcd-io/etcd-operator/blob/main/docs/design/reconcile_loop_v0.1.0.png). Roughly, in each reconcile loop, we should resolve any exceptions (see below) first if any before performing any normal reconciliation operations,
- verification: block any invalid requests
- [exception] fix unhealthy member/POD. It's next step goal, we can just add a placeholder
- [exception] fix POD which has mismatched conf
- [normal operation] scale in or scale out
- [normal operation] upgrade if needed

internal/controller/etcdcluster.go
ahrtr: > If we included the size in the hash calculation, scaling in or out would mean the existing pods are always invalidated, hence needs to be rerolled, which is not what we want.

ack

internal/controller/etcdcluster.go
```diff
@@ -0,0 +1,30 @@
...
+	clusterSpec := ec.DeepCopy().Spec
+	// size is not used for calculating the hash
+	clusterSpec.Size = 0
+	clusterSpec.EtcdOptions = createArgs(ec.Name, ec.Spec.EtcdOptions)
+	// we don't want to roll etcd pods when metadata is changed.
+	// instead, we'd do in-place update for them.
+	if clusterSpec.PodTemplate != nil {
+		clusterSpec.PodTemplate.Metadata = nil
+	}
+	deterministicString := dump.ForHash(clusterSpec)
```
ahrtr, line 26: Just checked, the dump.ForHash forts the keys, so it will have the same hash if users only change the order of the fields. Good!

https://github.com/kubernetes/utils/blob/cf1189d6abe3a7522bbd54a1bc1068a0a56011c8/dump/dump.go#L33

ahrtr (approved):
LGTM, thx

@silentred any comment on the test stuff?

internal/controller/etcdcluster.go
ahrtr: can you remove copyright header for now?

test/e2e/e2e_test.go
```diff
@@ -413,3 +427,74 @@ func TestEtcdClusterFunctionality(t *testing.T) {
...
+				if !ok {
+					t.Errorf("pod %s does not have its cluster hash: annotations: %v", pod.Name, pod.Annotations)
+				}
+				if podAnnotation != clusterHash {
+					t.Errorf("pod %s has a mismatched cluster hash: want %s, got %s: pod annotations: %v",
+						pod.Name, clusterHash, podAnnotation, pod.Annotations)
+				}
+			}
+			return ctx
+		})
```
silentred, line 487: Missing ` _ = testEnv.Test(t, feature.Feature())` ?

test/e2e/e2e_test.go
```diff
@@ -344,6 +346,18 @@ func TestEtcdClusterFunctionality(t *testing.T) {
...
 	})
 
+	feature.Assess("verify all Pods have the etcd-cluster hash annotation",
+		func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			var pods corev1.PodList
+			if err := c.Client().Resources().List(
+				ctx,
+				&pods,
+				resources.WithLabelSelector(fmt.Sprintf("app=%s", etcdClusterName))); err != nil {
+				t.Fatalf("unable to retrieve the member pods with labels %s and annotation %s", "fix", "me")
```
silentred: Don't forget checking annotation `controller.HashMetadataKey` later.

internal/controller/pods.go
```diff
@@ -229,11 +233,12 @@ func buildMemberPod(ec *ecv1alpha1.EtcdCluster, podName string, initialClusterSt
 	maps.Copy(labels, etcdClusterLabels(ec))
 
 	// Annotations are purely from the PodTemplate; nil when not provided.
-	var annotations map[string]string
+	annotations := make(map[string]string)
 	if ec.Spec.PodTemplate != nil && ec.Spec.PodTemplate.Metadata != nil &&
 		len(ec.Spec.PodTemplate.Metadata.Annotations) > 0 {
 		annotations = ec.Spec.PodTemplate.Metadata.Annotations
```
silentred: `annotations[HashMetadataKey] = EtcdClusterHash(ec)` changes the data of `ec`, which smells not good.

How about copying the whole map?
```
maps.Copy(annotations, ec.Spec.PodTemplate.Metadata.Annotations)
```

internal/controller/pods.go
```diff
@@ -229,11 +233,12 @@ func buildMemberPod(ec *ecv1alpha1.EtcdCluster, podName string, initialClusterSt
 	maps.Copy(labels, etcdClusterLabels(ec))
 
 	// Annotations are purely from the PodTemplate; nil when not provided.
```
silentred: `Annotations are purely from the PodTemplate` is not true now.

internal/controller/pods.go
nwnt: Not sure if I follow the comment.  The proposed change doesn't make any changes to the data of `ec`, it only calculates the hash from `ec` and adds the hash as another annotation entry.  `make` is needed so that we can ensure the map is always initialized and not nil (ran into an issue during e2e from inserting into a nil map).

I think we could use what you propose inside the if clause as a refactoring, but that's outside of the scope of this PR.  I can send a follow up PR after this one gets merged though.

test/e2e/e2e_test.go
nwnt: Thanks, forgot to remove this part after moving it to a separate test altogether.

test/e2e/e2e_test.go
nwnt: Done.

internal/controller/etcdcluster.go
nwnt: Yep, sorry I forgot to remove from the test file.  Done.

internal/controller/pods.go
silentred:
```
annotations = ec.Spec.PodTemplate.Metadata.Annotations
```
`annotations` is now a pointer to `ec.Spec.PodTemplate.Metadata.Annotations`
```
annotations[HashMetadataKey] = EtcdClusterHash(ec)
```
`ec.Spec.PodTemplate.Metadata.Annotations` adds a key-value of HashMetadataKey

nwnt: thanks, addressed it now.

silentred (approved):
LGTM, Thanks!

internal/controller/etcdcluster.go
ahrtr: I just realized that we should exclude the version from the hash as well, as we handle upgrade separately.

ahrtr: > as we handle upgrade separately.

as mentioned in https://github.com/etcd-io/etcd-operator/pull/422#discussion_r3578113762

ahrtr:

internal/controller/etcdcluster.go
nwnt: Oh ok.  Raised https://github.com/etcd-io/etcd-operator/pull/427 for this.

PR #427, opened 2026-07-15
Exclude etcd version from spec hash calculation

nwnt:
Addressing this comment from the hash calculation PR https://github.com/etcd-io/etcd-operator/pull/422#discussion_r3580434361

cc @ahrtr


ahrtr (approved):
LGTM & thx

Issue #409, opened 2026-07-07
Regression: Certificate feature is broken


ahrtr:
The duration something like "365d" isn't supported, so it breaks the certificate feature.

silentred:
It seems the `main` branch has the same problem, because `time.ParseDuration()` does not recognize `d`, https://github.com/golang/go/issues/17767 .

https://github.com/etcd-io/etcd-operator/blob/cf269b5de2313641194a3c6de1fd5c4eea39d522/internal/controller/utils.go#L582-L591

silentred:
I noticed another issue: when deploying the etcdcluster using `config/samples/sample_autocert_etcdcluster.yaml`, the actual etcd pod args don't include TLS, even though the TLS secret was generated successfully.

```
# secrets look fine
kubectl get secrets -A
NAMESPACE     NAME                            TYPE                            DATA   AGE
default       etcdcluster-sample-client-tls   kubernetes.io/tls               3      99m
default       etcdcluster-sample-peer-tls     kubernetes.io/tls               3      99m
default       etcdcluster-sample-server-tls   kubernetes.io/tls               3      99m

# no TLS args
kubectl get pod etcdcluster-sample-0 -oyaml | grep args -C10
spec:
  containers:
  - args:
    - --name=$(POD_NAME)
    - --listen-peer-urls=http://0.0.0.0:2380
    - --listen-client-urls=http://0.0.0.0:2379
    - --initial-advertise-peer-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2380
    - --advertise-client-urls=http://$(POD_NAME).etcdcluster-sample.$(POD_NAMESPACE).svc.cluster.local:2379
    command:
    - /usr/local/bin/etcd
```

AI analysis:
- static values returned by `func defaultArgs(name string) []string`
- no arg configuration for cert-file、key-file、trusted-ca-file、client-cert-auth
- missing container VolumeMounts for server-secret and peer-secret
- `peerEndpointForOrdinalIndex` and `clientEndpointForOrdinal` return only `http` endpoint.

ahrtr:
> the actual etcd pod args don't include TLS, even though the TLS secret was generated successfully

Thanks, this should be the reason why certificate feature is broken. We need to resolve this. cc @nwnt

ahrtr:
@silentred do you have bandwidth and interested in handling this ticket?

silentred:
[...]

PR #428, opened 2026-07-15
fix(tls): support auto and cert-manager TLS configs for customized controller

silentred:
# Summary

This PR wires up TLS support on the customized controller so that TLS-enabled etcd clusters can be created successfully. It targets the customized-controller branch and addresses #409.

I tested both the `auto` and `cert-manager` TLS configuration types, and both can create a TLS-enabled etcd cluster correctly.

# Notes

1. validityDuration cannot use the d suffix (e.g. 365d)

time.ParseDuration cannot parse strings like 365d; v0.2.0 produces the same error. This is by design in the Go standard library — see [golang/go#17767](https://github.com/golang/go/issues/17767).

To make the sample YAML work out of the box, I changed validityDuration to 8760h (= 365 × 24h). With this value, the sample YAML creates an etcd cluster successfully.

If support for parsing the `d` suffix is desired, it can be addressed in a follow-up PR.

2. Client TLS does not yet handshake with the etcd server

Currently the controller uses the server TLS certificate as the etcd client certificate to talk to the etcd server; it does not use client TLS.

Because the client ca.crt is issued by an independent CA and is not included in the etcd server's --trusted-ca-file, client TLS cannot complete a handshake with the etcd server today.

Future work: bundle the client CA together with the server CA and add the bundle to the etcd server's trust chain (--trusted-ca-file). Once that is in place, client TLS will be able to communicate with the etcd server normally.

/cc @ahrtr @nwnt

### DIFF
--- a/config/samples/sample_autocert_etcdcluster.yaml
+++ b/config/samples/sample_autocert_etcdcluster.yaml
@@ -13,4 +13,4 @@ spec:
     providerCfg:
       autoCfg:
         commonName: "etcd-operator-system"
-        validityDuration: 365d
+        validityDuration: 8760h
--- a/config/samples/sample_certmanager_etcdcluster.yaml
+++ b/config/samples/sample_certmanager_etcdcluster.yaml
@@ -22,6 +22,6 @@ spec:
     providerCfg:
       certManagerCfg:
         commonName: "etcd-operator-system"
-        validityDuration: "365d"
+        validityDuration: "8760h"
         issuerKind: "ClusterIssuer"
         issuerName: "selfsigned"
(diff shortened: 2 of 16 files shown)


ahrtr:
@silentred I haven't reviewed this PR yet. Please consider breaking down such a huge PR into smaller ones as much as you can.  Pls try to fix each relatively independent thing in a separate smaller PR. thx

silentred:
> @silentred I haven't reviewed this PR yet. Please consider breaking down such a huge PR into smaller ones as much as you can. Pls try to fix each relatively independent thing in a separate smaller PR. thx

OK, I will try to break it down after making e2e test passed.

silentred:
seperated to 5 sub-tasks:
- tls-step1-sample-and-cert: https://github.com/etcd-io/etcd-operator/pull/430
- tls-step2-etcdutils-options: add TLS options of etcd client used by operator. https://github.com/etcd-io/etcd-operator/pull/431
- tls-step3-mutual-tls:  https://github.com/etcd-io/etcd-operator/pull/434

PR #429, opened 2026-07-15
Add workflow for v0.3.0

ahrtr:
cc @nwnt @silentred 

I will present this workflow in our next community meeting. 

Some details (e.g. PV/PVC stuff) still need clarification. But this is just a high level workflow, it should be already good enough for now.


nwnt:
Just to put the comments here for the discussion during the meeting:
1. I think the reconciliation loop doesn't need to actively correct the ordinal, and instead should only be correcting when there's a need to replace the pod.  The 'lazy' approach should in theory reduce the potential instability from replacing an existing pod, which might or might not be a leader.
2. The update config should be perhaps be done earlier than later (i.e. fix a failed member or create an additional pod when the expected count is higher).  By doing so, I think it introduces less disruption by not having to go through pod creation and then later kill them because they're mismatched with the current content of `EtcdCluster`.
3. When the leader pod is about to be killed, we should also move the leader to the pod with the least possible ordinal as the higher ones are deemed to be killed first unless they are a leader:
   - In a 5 replica cluster, if the current leader is 3, then 4, 2, 1, 0 are killed and recreated first.  The leadership would be transferred from 3 to 0, as 4 will be killed first and before 3 during a scale-in.
   - If the current leader is 0, it would still be killed last, but the leadership should then be moved to 1 and the operator will leave it there.

docs/design/reconcile_loop_v0.3.0.uxf
```diff
@@ -0,0 +1,1641 @@
...
+  <element>
+    <id>UMLClass</id>
+    <coordinates>
+      <x>650</x>
+      <y>1050</y>
+      <w>960</w>
+      <h>410</h>
+    </coordinates>
+    <panel_attributes>lt=.
+Health check &amp; Fix failed member/POD
```
silentred, line 1290: "failed member/POD" means members that cannot pass health check?

docs/design/reconcile_loop_v0.3.0.uxf
```diff
@@ -0,0 +1,1641 @@
...
+  </element>
+  <element>
+    <id>UMLState</id>
+    <coordinates>
+      <x>930</x>
+      <y>1370</y>
+      <w>230</w>
+      <h>40</h>
+    </coordinates>
+    <panel_attributes>try to fix one member/POD</panel_attributes>
```
silentred, line 1444: Fixing workflow seems to be a heavy work. It depends on cases, quorum hold vs not hold, data corrupted vs not corrupted.

docs/design/reconcile_loop_v0.3.0.uxf
```diff
@@ -0,0 +1,1641 @@
...
+  <element>
+    <id>UMLClass</id>
+    <coordinates>
+      <x>650</x>
+      <y>1490</y>
+      <w>960</w>
+      <h>510</h>
+    </coordinates>
+    <panel_attributes>lt=.
+Exception handling
```
silentred, line 1277: Currently, it is more like "ensuring replicas number". What is the difference with "Scale in & out" section?

Should we wait the pod, created in the previous loop and in the starting process, to be Ready and started in etcd cluster Members API?

nwnt:
Comments from the community meeting:
1. Learner promotion maybe should be done at the fix failed member stage to avoid deadlocks from having multiple learners.
2. During exception handling that we need to remove pods, if the pod we're about to remove is the leader, we need to move it to other member first.
3. If the config of etcd cluster has been updated, the pod with highest ordinal will be replaced first and so on.

silentred:
> 1. Learner promotion maybe should be done at the fix failed member stage to avoid deadlocks from having multiple learners.

A similar case in the meeting: 
For a misconfigured 3-node cluster with one failed member pod, should we choose to fix the member or replace the old pod (user has updated EtcdCluster CR) first? New member may not be able join the cluster, due to the old wrong Spec.

docs/design/reconcile_loop_v0.3.0.uxf
ahrtr: Yes, we need to breakdown this when we work on the [v0.3.0 roadmap tasks](https://github.com/etcd-io/etcd-operator/blob/main/docs/roadmap.md#v030).

docs/design/reconcile_loop_v0.3.0.uxf
ahrtr: Basically yes

ahrtr:
> Comments from the community meeting:
> 
> 1. Learner promotion maybe should be done at the fix failed member stage to avoid deadlocks from having multiple learners.
> 2. During exception handling that we need to remove pods, if the pod we're about to remove is the leader, we need to move it to other member first.
> 3. If the config of etcd cluster has been updated, the pod with highest ordinal will be replaced first and so on.

Updated.



> > 1. Learner promotion maybe should be done at the fix failed member stage to avoid deadlocks from having multiple learners.
> 
> A similar case in the meeting: For a misconfigured 3-node cluster with one failed member pod, should we choose to fix the member or replace the old pod (user has updated EtcdCluster CR) first? New member may not be able join the cluster, due to the old wrong Spec.

Yes

@nwnt @silentred this is just a high level workflow. There are two key points:
- We always try to fix any problems/issues first, before performing any normal operation. 
- We need to ensure the workflow is idempotent

If there is no any objection, let's merge this for now. We can discuss more details separately when we work on the [refactoring work](https://github.com/etcd-io/etcd-operator/issues/362) and [v0.3.0 roadmap tasks](https://github.com/etcd-io/etcd-operator/blob/main/docs/roadmap.md#v030)

ahrtr:
cc @ArkaSaha30 @nwnt @silentred @ivanvc PTAL

nwnt (approved):
Hmm the files haven't been updated according to https://github.com/etcd-io/etcd-operator/pull/429#issuecomment-5037789714, but I agree with your point around the flow to fix the failed/exception pods first before scaling in or out.

ahrtr:
> Hmm the files haven't been updated according to [#429 (comment)](https://github.com/etcd-io/etcd-operator/pull/429#issuecomment-5037789714),

Oops, just checked the file, I did updated the file, somehow the change was missing...

Will send a separate PR to update it, also get all notes included in the TODO comment in https://github.com/etcd-io/etcd-operator/pull/441

PR #430, opened 2026-07-16
chore(tls): cleanup sample yaml + auto cert for client auth

silentred:
sub-task 1 of #428

Questions might be asked:

- **Why change sample yaml?**
`time.ParseDuration` cannot parse strings like `365d`. v0.2.0 produces the same error. If support for parsing the d suffix is desired, it can be addressed in a follow-up PR.

- **Why add `ClientAuth` usage to cert?**
The controller is supposed use the server TLS certificate as the etcd client certificate to talk to the etcd server; it does not use client TLS.
Because the client ca.crt is issued by an independent CA and is not included in the etcd server's --trusted-ca-file, client TLS cannot complete a handshake with the etcd server today.
Future work: bundle the client CA together with the server CA and add the bundle to the etcd server's trust chain (--trusted-ca-file). Once that is in place, client TLS will be able to communicate with the etcd server normally.

/cc @ahrtr @nwnt

### DIFF
--- a/config/samples/sample_autocert_etcdcluster.yaml
+++ b/config/samples/sample_autocert_etcdcluster.yaml
@@ -13,4 +13,4 @@ spec:
     providerCfg:
       autoCfg:
         commonName: "etcd-operator-system"
-        validityDuration: 365d
+        validityDuration: 8760h
--- a/pkg/certificate/auto/provider.go
+++ b/pkg/certificate/auto/provider.go
@@ -294,7 +294,11 @@ func (ac *Provider) createNewSecret(ctx context.Context, secretKey client.Object
 		return fmt.Errorf("validity duration converts to 0 years, must be at least 1 year")
 	}
 
-	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, validityYears)
+	// transport.SelfCert generates a self-signed cert whose ExtKeyUsage
+	// defaults to ServerAuth only. The operator reuses the server Secret
+	// as its own mTLS client identity talking with the etcd cluster.
+	// The cert must also carry ClientAuth.
+	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, validityYears, x509.ExtKeyUsageClientAuth)
 	if selfCertErr != nil {
 		return fmt.Errorf("certificate creation via transport.SelfCert failed: %w", selfCertErr)
 	}
--- a/config/samples/sample_certmanager_etcdcluster.yaml
+++ b/config/samples/sample_certmanager_etcdcluster.yaml
@@ -22,6 +22,6 @@ spec:
     providerCfg:
       certManagerCfg:
         commonName: "etcd-operator-system"
-        validityDuration: "365d"
+        validityDuration: "8760h"
         issuerKind: "ClusterIssuer"
         issuerName: "selfsigned"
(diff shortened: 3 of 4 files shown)


ahrtr:

pkg/certificate/auto/provider.go
```diff
@@ -294,7 +294,11 @@ func (ac *Provider) createNewSecret(ctx context.Context, secretKey client.Object
 		return fmt.Errorf("validity duration converts to 0 years, must be at least 1 year")
 	}
 
-	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, validityYears)
+	// transport.SelfCert generates a self-signed cert whose ExtKeyUsage
+	// defaults to ServerAuth only. The operator reuses the server Secret
+	// as its own mTLS client identity against the etcd client listener.
```
ahrtr:
```suggestion
	// transport.SelfCert generates a self-signed cert whose ExtKeyUsage
	// defaults to ServerAuth only. The operator reuses the server Secret
	// as its own mTLS client identity talking with the etcd cluster.
```

pkg/certificate/auto/provider.go
```diff
@@ -294,7 +294,11 @@ func (ac *Provider) createNewSecret(ctx context.Context, secretKey client.Object
...
+	// The cert must also carry ClientAuth.
+	tlsInfo, selfCertErr := transport.SelfCert(zap.NewNop(), tmpDir, hosts, validityYears, x509.ExtKeyUsageClientAuth)
```
ahrtr, line 301: I am curious why we did not see any issue previously without `x509.ExtKeyUsageClientAuth`? cc @ArkaSaha30

silentred: What I see is that etcdserver pod has never enabled TLS args.

ahrtr:
> * Because the client ca.crt is issued by an independent CA and is not included in the etcd server's --trusted-ca-file, client TLS cannot complete a handshake with the etcd server today.

I think this is a good catch. But why we did not see any issue previously with Statefulset? The etcd-operator needs to perform health check, it should fail right? @ArkaSaha30 did you see any issue previously when you were testing the feature?

pkg/certificate/auto/provider.go
ahrtr: > What I see is that etcdserver pod has never enabled TLS args.

Just double checked the source code on main branch, this is true. We mounted the secret into POD, but actually do not use it at all. Peers talk to each other use http always, and etcd-operator talks to etcd via http always as well.

We missed such an elephant!

config/samples/sample_autocert_etcdcluster.yaml
```diff
@@ -13,4 +13,4 @@ spec:
     providerCfg:
       autoCfg:
         commonName: "etcd-operator-system"
-        validityDuration: 365d
+        validityDuration: 8760h
```
ahrtr, line 16: We need to consider an alternative utility to replace the time.ParseDuration, we should support duration like "365d"

ahrtr (approved):
Good catch! Thanks.

config/samples/sample_autocert_etcdcluster.yaml
silentred: OK, there gona be some research work. Do you mind that I handle it in a follow-up PR?

ahrtr: >Do you mind that I handle it in a follow-up PR?

sure

PR #431, opened 2026-07-16
refactor(etcdutils): encapsulate client settings in ClientConfig struct

silentred:
sub-task 2 of https://github.com/etcd-io/etcd-operator/pull/428

/cc @ahrtr @nwnt

### DIFF
--- a/internal/etcdutils/etcdutils.go
+++ b/internal/etcdutils/etcdutils.go
@@ -2,6 +2,7 @@ package etcdutils
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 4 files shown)


internal/etcdutils/etcdutils.go
```diff
@@ -11,34 +12,62 @@ import (
...
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
-func MemberList(eps []string) (*clientv3.MemberListResponse, error) {
-	cfg := clientv3.Config{
+// Option configures a clientv3.Config built by the helpers in this package.
+// It is the single, backward-compatible extension point that callers use to add
+// TLS (WithTLS) without changing positional signatures.
+type Option func(*clientConfig)
```
ahrtr: It seems that it is and will only be used for TLS thing? Note that etcd-operator doesn't care about any other parameters/options provided by users. 

ref: https://github.com/etcd-io/etcd-operator/pull/93

internal/etcdutils/etcdutils.go
```diff
@@ -230,55 +242,31 @@ func AddMember(eps []string, peerURLs []string, learner bool) (*clientv3.MemberA
...
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	defer func() {
-		err := c.Close()
-		if err != nil {
-			cancel()
-			return
-		}
-		cancel()
-	}()
+	defer func() { closeAndCancel(c, cancel) }()
```
ahrtr, line 247: If you move such kind of minor & clear change into a separate PR, it will be merged right away.

internal/etcdutils/etcdutils.go
silentred: It was kind of over-designed. How about changing signature to 
```
func MemberList(eps []string, *tls.Config) (*clientv3.MemberListResponse, error) {
```
instead of 
```
func MemberList(eps []string, opts ...Option) (*clientv3.MemberListResponse, error) {
```

ahrtr: > func MemberList(eps []string, *tls.Config) (*clientv3.MemberListResponse, error) {

This might not be extensible.  I just realized that my previous comment https://github.com/etcd-io/etcd-operator/pull/431#discussion_r3597223770 confused etcd itself's options with the client options used by etcd-operator to connect to etcd.  This specific change should be the latter, but I think we need to resolve both cases.

ahrtr: > the client options used by etcd-operator to connect to etcd

https://github.com/etcd-io/etcd/blob/6006f405800929b5e7e839e7a821d608a311579f/client/v3/config.go#L118-L128

silentred: `ConfigSpec` might not be the best choice for an input parameter. Since `SecureConfig` contains TLS file paths, but the operator doesn't actually need these files (it can generate `tls.Config` directly), using `ConfigSpec` seems unnecessary. If the goal is extensibility, I believe we should just use `clientv3.Config` directly. (https://github.com/etcd-io/etcd/blob/6006f405800929b5e7e839e7a821d608a311579f/client/v3/config.go#L28)

However, using clientv3.Config directly exposes too many options to the caller, even if the caller is the operator itself. Alternatively, we could encapsulate ep []string and *tls.Config into a new struct, such as etcdClientConfig. This would allow us to add more client configuration fields in the future if needed.

ahrtr: > Since SecureConfig contains TLS file paths, but the operator doesn't actually need these files (it can generate tls.Config directly), using ConfigSpec seems unnecessary.

Right. The point is that we need to support all the options in `ConfigSpec` to be configurable. 

> we could encapsulate ep []string and *tls.Config into a new struct, such as etcdClientConfig

OK, we can proceed with this for now.

ahrtr (approved):

PR #432, opened 2026-07-18
Fix post e2e job from mismatched hash calculation

nwnt:
Fixing the e2e tests from the mismatched hash calculation.  Instead of calculating the expected hash from the static `EtcdCluster` resource, it's not calculating from the retrieve `EtcdCluster` resource.

cc @ahrtr

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -81,7 +81,8 @@ func TestInvalidClusterSize(t *testing.T) {
 			return ctx
 		})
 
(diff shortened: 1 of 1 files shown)


silentred:
Is `TestPodRecovery/multi-node-pod-recovery/verify_multi-node_cluster_recovery_after_pod_deletion` expected to fail?

nwnt:
No.  This seems to be a flake.

ahrtr (approved):
Thanks

PR #434, opened 2026-07-20
feat(tls): support TLS-aware pods and operator mTLS client identity

silentred:
Changes:
1. enable server-side and client-side mTLS communication (~400 lines)
2. fix verifyDataOperations in e2e tests. The old one talks to etcdserver without TLS.
3. add dumpPodLogs helper function in e2e tests.
4. publish not-ready addresses on etcd headless Service.

1,2,4 have to stay togethor, otherwise the e2e tests will fail. 
3 is a super minor change, so I put it in this PR.

### Questions might be asked:

**Why set PublishNotReadyAddresses=true in headless Service?**

In etcd v3.6, the peer listener performs an additional custom `checkSAN` validation after the Go TLS handshake (`client/pkg/transport/listener.go`). This validation conducts forward and reverse DNS lookups on the client certificate's SAN entries and matches the results against the connecting pod's actual IP. It is independent of standard certificate chain validation and cannot be disabled via flags on the peer port.

Headless Services have a default `publishNotReadyAddresses: false`. During bootstrap, a new member is in a NotReady state. Consequently, CoreDNS excludes its IP from the Headless Service's A records.

As a result, the new member's own IP never appears in the forward DNS lookup results. `checkSAN` rejects every peer handshake with the error tls: "<ip>" does not match any of DNSNames. etcd fails to start and returns EOF, causing the pod to enter a `CrashLoopBackOff` state. Since subsequent startup attempts trigger the same check before the pod becomes `Ready`, it can never recover. This creates a chicken-and-egg deadlock: the pod must be Ready to pass the TLS handshake, but it needs to pass the TLS handshake to become Ready.

e2e test `TestClusterAutoCertCreation/cluster-auto-cert-creation/Wait_for_etcd_pods_readiness` fails when `publishNotReadyAddresses: false`.

pod-0 logs contain errors like 
```
{"level":"warn","ts":"2026-07-21T06:45:30.310841Z","caller":"embed/config_logging.go:185","msg":"rejected connection on peer endpoint","remote-addr":"10.244.0.8:43166","server-name":"etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local","ip-addresses":[],"dns-names":["etcd-operator-system","*.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local","etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local"],"error":"tls: \"10.244.0.8\" does not match any of DNSNames [\"etcd-operator-system\" \"*.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local\" \"etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local\"] (lookup etcd-operator-system on 10.96.0.10:53: no such host)"}
```

pod-1 logs:
```
{"level":"warn","ts":"2026-07-21T06:45:30.310937Z","caller":"etcdserver/cluster_util.go:86","msg":"failed to get cluster response","address":"https://etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local:2380/members","error":"Get \"https://etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local:2380/members\": EOF"}
{"level":"error","ts":"2026-07-21T06:45:30.311297Z","caller":"etcdserver/server.go:309","msg":"bootstrap failed","error":"cannot fetch cluster info from peer urls: could not retrieve cluster information from the given URLs","stacktrace":"go.etcd.io/etcd/server/v3/etcdserver.NewServer\n\tgo.etcd.io/etcd/server/v3/etcdserver/server.go:309\ngo.etcd.io/etcd/server/v3/embed.StartEtcd\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:262\ngo.etcd.io/etcd/server/v3/etcdmain.startEtcd\n\tgo.etcd.io/etcd/server/v3/etcdmain/etcd.go:207\ngo.etcd.io/etcd/server/v3/etcdmain.startEtcdOrProxyV2\n\tgo.etcd.io/etcd/server/v3/etcdmain/etcd.go:129\ngo.etcd.io/etcd/server/v3/etcdmain.Main\n\tgo.etcd.io/etcd/server/v3/etcdmain/main.go:40\nmain.main\n\tgo.etcd.io/etcd/server/v3/main.go:31\nruntime.main\n\truntime/proc.go:285"}
```

sub-task 3 of https://github.com/etcd-io/etcd-operator/pull/428

/cc @ahrtr @nwnt

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"strings"
 	"time"
@@ -57,6 +58,7 @@ type reconcileState struct {
 	pods           []*corev1.Pod                // member pods owned by this cluster, sorted by ordinal
 	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
 	memberHealth   []etcdutils.EpHealth         // health information for each etcd member
+	tlsConfig      *tls.Config                  // etcd client TLS config used by every etcdutils call in this loop (nil for non-TLS clusters)
 }
 
 // +kubebuilder:rbac:groups=operator.etcd.io,resources=etcdclusters,verbs=get;list;watch;create;update;patch;delete
@@ -131,6 +133,12 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		))
 	}
 
+	// Build the operator's etcd-client TLS config (from the server cert Secret)
+	clientTLS, tlsErr := r.buildReconcileClientTLS(ctx, ec)
+	if tlsErr != nil {
+		logger.Error(tlsErr, "Failed to build client TLS config; will retry next reconcile")
+	}
+
 	logger.Info("Reconciling EtcdCluster", "spec", ec.Spec)
 
 	pods, err := listOwnedPods(ctx, r.Client, ec)
@@ -138,6 +146,9 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		logger.Error(err, "Failed to list pods. Requesting requeue")
 		return nil, ctrl.Result{RequeueAfter: requeueDuration}, nil
 	}
+
+	// When TLS is unused tlsConfig stays nil.
+	rs := &reconcileState{cluster: ec, pods: pods, tlsConfig: clientTLS}
 
 	// Validate the upgrade path using the image tag of the first pod.
 	if len(pods) > 0 {
@@ -149,7 +160,7 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 			if idx == -1 {
 				logger.Info("could not extract image version from pod image",
 					"image", c.Image)
-				return &reconcileState{cluster: ec, pods: pods}, ctrl.Result{}, nil
+				return rs, ctrl.Result{}, nil
 			}
 			currentVersion := c.Image[idx+1:]
 			targetVersion := ec.Spec.Version
@@ -163,7 +174,7 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 						"target", targetVersion,
 						"error", err,
 					)
-					return &reconcileState{cluster: ec, pods: pods}, ctrl.Result{}, nil
+					return rs, ctrl.Result{}, nil
 				}
 				if err != nil {
 					logger.Error(err, "unsupported upgrade path between current and target versions",
@@ -180,7 +191,17 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		}
 	}
 
-	return &reconcileState{cluster: ec, pods: pods}, ctrl.Result{}, nil
+	return rs, ctrl.Result{}, nil
+}
+
+// buildReconcileClientTLS builds the operator's etcd-client TLS Config for a
+// reconcile loop from the cluster's server certificate Secret. Returns a nil
+// config (and nil error) when the cluster has no TLS configured.
+func (r *EtcdClusterReconciler) buildReconcileClientTLS(ctx context.Context, ec *ecv1alpha1.EtcdCluster) (*tls.Config, error) {
+	if ec.Spec.TLS == nil {
+		return nil, nil
+	}
+	return buildClientTLSConfig(ctx, ec, r.Client)
 }
 
 // bootstrapCluster ensures the headless Service exists and, when no pods are
@@ -211,7 +232,7 @@ func (r *EtcdClusterReconciler) performHealthChecks(ctx context.Context, s *reco
 	logger := log.FromContext(ctx)
 	logger.Info("Now checking health of the cluster members")
 	var err error
-	s.memberListResp, s.memberHealth, err = healthCheck(s.cluster.Name, s.cluster.Namespace, s.pods, logger)
+	s.memberListResp, s.memberHealth, err = healthCheck(s.cluster.Name, s.cluster.Namespace, s.pods, clusterTLSEnabled(s.cluster), s.tlsConfig, logger)
 	if err != nil {
 		return fmt.Errorf("health check failed: %w", err)
 	}
@@ -275,10 +296,10 @@ func (r *EtcdClusterReconciler) reconcileClusterState(ctx context.Context, s *re
 			logger.Info("Learner found", "learnerID", learner, "learnerStatus", learnerStatus)
 			if etcdutils.IsLearnerReady(leaderStatus, learnerStatus) {
 				logger.Info("Learner is ready to be promoted to voting member", "learnerID", learner)
-				eps := clientEndpointsFromPods(s.cluster.Name, s.cluster.Namespace, s.pods)
+				eps := clientEndpointsFromPods(s.cluster.Name, s.cluster.Namespace, s.pods, clusterTLSEnabled(s.cluster))
 				// Exclude the learner (last ordinal) from the endpoint list used for promotion.
 				eps = eps[:(len(eps) - 1)]
-				if err := etcdutils.PromoteLearner(etcdutils.ClientConfig{Endpoints: eps}, learner); err != nil {
+				if err := etcdutils.PromoteLearner(etcdutils.ClientConfig{Endpoints: eps, TLS: s.tlsConfig}, learner); err != nil {
 					return ctrl.Result{}, err
 				}
 			} else {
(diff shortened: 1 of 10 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -131,6 +133,12 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		))
 	}
 
+	// Build the operator's etcd-client TLS config (from the server cert Secret)
+	clientTLS, tlsErr := r.buildReconcileClientTLS(ctx, ec)
```
ahrtr, line 137: Line 126 above (see below as well) already created client certificate for etcd-operator to use to talk to etcd server, we should use that certificate. The problem is that we need to ensure the certificate is issued by the same CA as the certificate used by etcd itself.

https://github.com/etcd-io/etcd-operator/blob/88fb17a854b2137bb4a9a9d6223a3e3c0a3f9f7e/internal/controller/etcdcluster_controller.go#L138

internal/controller/etcdcluster_controller.go
```diff
@@ -139,6 +147,12 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		return nil, ctrl.Result{RequeueAfter: requeueDuration}, nil
 	}
 
+	// Attach the per-reconcile TLS config to every returned state. When TLS is
+	// unused tlsConfig stays nil and ClientConfig callers get the no-TLS shape.
+	stateWithTLS := func(pods []*corev1.Pod) *reconcileState {
+		return &reconcileState{cluster: ec, pods: pods, tlsConfig: clientTLS}
+	}
```
ahrtr: We might not need a named function here, we can either revert this change or define a `rs := reconcileState{cluster: ec, pods: pods, tlsConfig: clientTLS}`, and then return it directly later when needed.

internal/controller/utils.go
```diff
@@ -158,6 +160,19 @@ func createHeadlessServiceIfNotExist(ctx context.Context, logger logr.Logger, c
...
+					// still bootstrapping). etcd v3.6's peer-port TLS handler runs
+					// checkCertSAN, which forward-DNS-resolves the cert's DNSName
+					// and checks the connecting Pod IP against the returned A
+					// records. Without this flag CoreDNS returns only Ready Pod
+					// IPs, so a new member's own IP is absent from the lookup
+					// result, every peer handshake it initiates is rejected
+					// ("tls: \"<ip>\" does not match any of DNSNames"), its etcd
+					// bootstrap dies, and the Pod stays NotReady forever — a
+					// chicken-and-egg deadlock during cluster scale-out.
+					PublishNotReadyAddresses: true,
```
ahrtr, line 175: question: did you see this checken-egg issue in your test/debug before you added `PublishNotReadyAddresses: true`?

internal/controller/utils.go
```diff
@@ -379,6 +396,61 @@ func patchCertificateSecret(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c c
...
+			"for cert-manager use a SelfSigned or CA Issuer (ACME does not emit ca.crt)",
+			secret.Name, provider)
+	}
+	return nil
+}
+
+// buildClientTLSConfig assembles the operator's TLS config from the cluster's
+// server certificate Secret. The operator reuses the server identity (tls.crt/tls.key)
+// and trusts the server CA (ca.crt), so the control plane can reach a TLS-enabled etcd client listener.
+func buildClientTLSConfig(ctx context.Context, ec *ecv1alpha1.EtcdCluster, c client.Client) (*tls.Config, error) {
```
ahrtr, line 418: We might want to add this functionality into pkg/certificate. We can 
- either add a new interface method something `CreateTLSConfigFromSecret`
- [preferred] or update `ValidateCertificateSecret`'s signature to return a secret, and then add a utility function to create a `tls.Config` from the secret.

ahrtr:
cc @ArkaSaha30

internal/controller/utils.go
silentred: I did not see this on my local kind cluster, on which I tested with v3.5.21 and v3.6.12 etcd version. I checked the code, both of them having the `checkSAN` enabled by default.
However on e2e workflow, it is stable to reproduce the issue. The 2nd etcd pod cannot handshake with the 1st pod's peer listener. I have not figured out why. 

job history: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-etcd-operator-test-e2e

etcd-cluster-auto-cert-0 logs:
```
{"level":"warn","ts":"2026-07-21T06:45:30.310841Z","caller":"embed/config_logging.go:185","msg":"rejected connection on peer endpoint","remote-addr":"10.244.0.8:43166","server-name":"etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local","ip-addresses":[],"dns-names":["etcd-operator-system","*.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local","etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local"],"error":"tls: \"10.244.0.8\" does not match any of DNSNames [\"etcd-operator-system\" \"*.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local\" \"etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local\"] (lookup etcd-operator-system on 10.96.0.10:53: no such host)"}
```

etcd-cluster-auto-cert-1 quit here:
```
{"level":"warn","ts":"2026-07-21T06:45:30.310937Z","caller":"etcdserver/cluster_util.go:86","msg":"failed to get cluster response","address":"https://etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local:2380/members","error":"Get \"https://etcd-cluster-auto-cert-0.etcd-cluster-auto-cert.etcd-operator-system.svc.cluster.local:2380/members\": EOF"}
```

internal/controller/etcdcluster_controller.go
silentred: > to ensure the certificate is issued by the same CA as the certificate used by etcd itself

Agree. This would be a large-sized change, and it would be best to bundle it with your suggestion below. `[preferred] or update ValidateCertificateSecret's signature to return a secret, and then add a utility function to create a tls.Config from the secret.` 

I suggest to handle it in a follow-up PR.

ahrtr: > I suggest to handle it in a follow-up PR.

OK.

ahrtr:
@ArkaSaha30 @nwnt do you have any comments on this PR?

The users/applications also need client certificate to access the etcd clusters created/managed by etcd-operator.  We need to extend our initial certificate [design doc](https://docs.google.com/document/d/1k8os77Hqxe36nUaZ31YqBGBuXcgdozVPppJYOfwUDFA/edit?tab=t.0) to cover this use case as well.  @silentred are you able to take care of this?

High level ideas:
- The categories of certificates:
    - peer-to-peer
    - etcd-operator -> etcd
    - application -> etcd
    - backup controller -> etcd (not exist yet, but highly likely we have have such a backup controller, we need to ensure our design is extensible)
-  We might need to have a dedicated secret for CA, so that all certificates are issues by the same CA

internal/controller/utils.go
ahrtr: Thanks

silentred:
> are you able to take care of this?

Yes. I already have a working draft code. I would let you guys know after revising the desgin doc later, if needed.

ahrtr:
> > are you able to take care of this?
> 
> Yes. I already have a working draft code. I would let you guys know after revising the desgin doc later, if needed.

Thanks. Having a clean & clear design doc is more important. Do you have a gmail account so that I can share the edit permission of the doc to you?

silentred:
> a gmail account

Thanks. silentred3@gmail.com

ahrtr:
> > a gmail account
> 
> Thanks. [silentred3@gmail.com](mailto:silentred3@gmail.com)

Done. I added a new tab "v0.3.0". Please feel free to add anything in that tab. We can discuss any details in the bi-weekly community meeting.

ahrtr:
I might do some refactoring on the controller based on the workflow in https://github.com/etcd-io/etcd-operator/pull/429. But we need merge this PR first to avoid conflict. @ArkaSaha30 @nwnt @ivanvc Please take a look at this PR, thx

internal/controller/etcdcluster_controller.go
```diff
@@ -131,6 +133,12 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		))
 	}
 
+	// Build the operator's etcd-client TLS config (from the server cert Secret)
+	clientTLS, tlsErr := r.buildReconcileClientTLS(ctx, ec)
+	if tlsErr != nil && ec.Spec.TLS != nil {
```
ahrtr: If `ec.Spec.TLS == nil`, the method `buildReconcileClientTLS`  always return nil
```suggestion
	if tlsErr != nil {
```

ahrtr (approved):

ahrtr:
Followup:
- https://github.com/etcd-io/etcd-operator/pull/434#discussion_r3621675380
- more importantly, we need to summarize all the use cases of certificates as mentioned in https://github.com/etcd-io/etcd-operator/pull/434#issuecomment-5045573308. Especially for the use case that how applications talk to the etcd clusters.

if anyone has any comment, we can address them separately in followup PRs. Merging this PR for now...

ahrtr:
@silentred @nwnt FYI. I will do a quick refactor on the reconcile workflow based on v0.3.0 workflow today.

internal/controller/etcdcluster_controller.go
```diff
@@ -131,6 +133,12 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		))
 	}
 
+	// Build the operator's etcd-client TLS config (from the server cert Secret)
+	clientTLS, tlsErr := r.buildReconcileClientTLS(ctx, ec)
+	if tlsErr != nil {
+		logger.Error(tlsErr, "Failed to build client TLS config; will retry next reconcile")
+	}
```
ahrtr, line 140: We missed this. We should return error, and ensure server and peer certificiates exist before calling `buildReconcileClientTLS`

ahrtr: resolved in the second commit in https://github.com/etcd-io/etcd-operator/pull/441

PR #445, opened 2026-07-31
feat: add `--service-dns-domain` flag for configurable member FQDN

### LINKED ISSUE #444: Proposal: configurable service DNS domain
## Problem

The operator hardcodes `svc.cluster.local` when building member FQDNs and the default cert SANs. The string appears in `internal/controller/pods.go`, `internal/controller/utils.go`, and `pkg/certificate/interfaces/interface.go` (`DefaultDomainName`).

Any k8s cluster whose apiserver was started with a non-default `--service-dns-domain` (common in self-hosted / air-gapped / k3s / kubeadm
deployments) is broken at first reconcile:
- Pods advertise `--advertise-*-urls` containing `cluster.local`, which
CoreDNS does not resolve.
- Server cert SANs are `*.<svc>.<ns>.svc.cluster.local`, so peer mTLS
handshakes fail `checkCertSAN` and the cluster never forms.

## Proposed fix

The service DNS domain is a property of the *cluster*, not of any individual `EtcdCluster`, so the primary knob should be an operator-level flag:
- Add `--cluster-domain` to `cmd/main.go`, default `cluster.local`.
- Wire it onto `EtcdClusterReconciler` (same pattern as `ImageRegistry`).
- Resolve the suffix everywhere a member URL or cert SAN is built: `clientEndpointForOrdinal`, `defaultArgs`, `peerEndpointForOrdinalIndex`, `createCMCertificateConfig`, `createAutoCertificateConfig`.
- Stop referencing `certInterface.DefaultDomainName` from the controller (keep the constant for backward compatibility, but mark deprecated).

Optional follow-up (not required for the fix): an `EtcdClusterSpec.dnsDomain` override for the rare case of one operator spanning multiple suffix domains or for staged migration.

## Why not just a CRD field

A per-`EtcdCluster` field works, but every cluster then has to repeat the value the operator already knows. A flag matches the one-operator-per-cluster deployment shape and keeps the existing manifest examples unchanged (default remains `cluster.local`).

## Acceptance criteria

- Operator runs unchanged on a `cluster.local` cluster.
- Operator runs on a cluster with `--service-dns-domain=corp.local` when started with `--cluster-domain=corp.local`: Pods advertise and serve on `*.svc.corp.local`, cert SANs match, TLS peer/client handshakes succeed.
- Unit tests assert the domain is honored; e2e test passes on a non-default suffix.
- Existing CRD examples and docs need updates.

silentred:
Resolve #444 

/cc @ahrtr @nwnt

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,6 +64,7 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var enableHTTP2 bool
+	var serviceDNSDomain string
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&imageRegistry, "image-registry", controller.DefaultImageRegistry,
 		"The container registry to pull etcd images from. Defaults to "+controller.DefaultImageRegistry+".")
@@ -77,6 +78,11 @@ func main() {
 		"If set, the metrics endpoint is served securely via HTTPS. Use --metrics-secure=false to use HTTP instead.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	flag.StringVar(&serviceDNSDomain, "service-dns-domain", "",
+		"Kubernetes service DNS suffix used to build member FQDNs. "+
+			"Empty (the default) uses the short form <pod>.<svc>.<ns>.svc, "+
+			"resolvable via the pod's DNS search path. "+
+			"Set this to the kubelet's --cluster-domain value to use the full FQDN form <pod>.<svc>.<ns>.svc.<domain>.")
 	opts := zap.Options{
 		Development: true,
 	}
@@ -151,9 +157,10 @@ func main() {
 	}
 
 	if err = (&controller.EtcdClusterReconciler{
-		Client:        mgr.GetClient(),
-		Scheme:        mgr.GetScheme(),
-		ImageRegistry: imageRegistry,
+		Client:           mgr.GetClient(),
+		Scheme:           mgr.GetScheme(),
+		ImageRegistry:    imageRegistry,
+		ServiceDNSDomain: serviceDNSDomain,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "EtcdCluster")
 		os.Exit(1)
--- a/internal/controller/etcdcluster.go
+++ b/internal/controller/etcdcluster.go
@@ -13,13 +13,13 @@ const (
 	hashLength = 12
 )
 
-func EtcdClusterHash(ec *v1alpha1.EtcdCluster) string {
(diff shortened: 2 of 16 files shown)


cmd/main.go
```diff
@@ -64,8 +64,9 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var enableHTTP2 bool
+	var serviceDNSDomain string
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&imageRegistry, "image-registry", "gcr.io/etcd-development/etcd",
+	flag.StringVar(&imageRegistry, "image-registry", controller.DefaultImageRegistry,
```
ahrtr, line 69: This PR should focus on supporting the new option `--service-dns-domain`, any other changes should be moved into separate PRs. This is a generic comment for better review communications for any PR, especially for such big PR (>500+ line of code change).

ahrtr:
Please ping me back after https://github.com/etcd-io/etcd-operator/pull/445#discussion_r3689360279 is resolved. thx

silentred:
Need to seperate into 2 sub tasks:
1. fix e2e test TestClusterHash: https://github.com/etcd-io/etcd-operator/pull/448
2. add `--service-dns-domain` flag

silentred:
2 changes:
- add `ServiceDNSDomain` field for EtcdClusterReconciler
- add a wrapper struct `etcdPodHelper` to simplify the parameters

cc @ahrtr

ahrtr:
I feel like that this PR is a little over complicated, we only need the following implementations,
- add a new flag `--service-dns-domain`
- function `memberDNSSuffix`
- pass `serviceDNSDomain` as a separate parameter if needed
- update one of existing e2e test cases to set `--service-dns-domain` to either empty or `cluster.local`

I expect the PR should be less than 100 if I am not wrong?

silentred:
> I feel like that this PR is a little over complicated, we only need the following implementations,
> 
> * add a new flag `--service-dns-domain`
> * function `memberDNSSuffix`
> * pass `serviceDNSDomain` as a separate parameter if needed
> * update one of existing e2e test cases to set `--service-dns-domain` to either empty or `cluster.local`
> 
> I expect the PR should be less than 100 if I am not wrong?

Done. Many functions depend on `memberDNSSuffix`, so we need to add `dnsDomain` to their parameters. It is more than 100 lines. 
If it looks OK to you, pls ping me to squash commits.

ahrtr:
I am drafting the new comprehensive design/workflows for v0.3.0, including etcd member lifecycle management and self-healing. It's the top priority. I will try to deliver a PR sometime this week, so that more people can get involved.

I'd like to on-hold other PRs (including this one). I was think it should be a simple feature and accordingly a minor change. I am not sure why it's such a huge PR. We can revisit this PR later. Thanks

silentred:
OK. Let's focusing on v0.3.0 for now.

PR #448, opened 2026-08-01
fix(controller): align etcdcluster hash between controller and tests

silentred:
## Problem

  - **Controller reconcile.** `fetchAndValidateState` filled `ec.Spec.ImageRegistry = r.ImageRegistry` in place, and `buildMemberPod` later called `EtcdClusterHash(ec)` to compute the  pod-template hash annotation.
  - **Unit / e2e tests.** `TestEtcdClusterHash` and e2e `TestClusterHash` just `Get`'d the cluster and called `EtcdClusterHash(etcdCluster)` directly, with no default fill, so `ec.Spec.ImageRegistry` was still `""`.

  A cluster whose user left `ImageRegistry` empty therefore produced **different** hashes on the controller path than on the test paths. That broke the contract that the pod hash
  annotation should match what the tests expect the operator to produce, and it meant the controller and the test could disagree on whether a spec change required a pod roll.

  ## Fix

  Centralize the operator-side defaulting into a new `FillEtcdClusterDefaultValues(ec, defaultImageRegistry)` helper in `internal/controller/utils.go`, and call it from both paths so `EtcdClusterHash` always sees the same input.

  The helper mutates the spec in memory only; the user-owned `Spec` is never persisted back, so the CR is not polluted with operator policy. Currently it fills only `ImageRegistry`, but the helper is the natural home for any future operator-side defaults (TLS material, storage class, etc.).

  Another fix is to ignore `ImageRegistry` when hashing, by adding `clusterSpec.ImageRegistry = ""`. This will ignore changes of `ImageRegistry` unless introducing a dedicated logic for it. So I did not choose this fix.

Link to https://github.com/etcd-io/etcd-operator/pull/445

/cc @ahrtr

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -165,9 +165,10 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		return nil, ctrl.Result{}, err
 	}
 
-	if ec.Spec.ImageRegistry == "" {
-		ec.Spec.ImageRegistry = r.ImageRegistry
-	}
+	// Apply operator-side defaults so downstream phases (hash, pod builder,
+	// cert wiring) all see consistent values regardless of which fields the
+	// user left empty.
+	r.PopulateDefaultValues(ec)
 
 	pods, err := listOwnedPods(ctx, r.Client, ec)
 	if err != nil {
@@ -176,6 +177,17 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 	}
 
 	return &reconcileState{cluster: ec, pods: pods}, ctrl.Result{}, nil
+}
+
+// PopulateDefaultValues mutates `ec` in place to apply operator-side defaults
+// for any EtcdClusterSpec field that the user left empty. Defaults are sourced
+// from the reconciler instance (e.g. its ImageRegistry), keeping the reconcile
+// loop the single owner of how a partially-specified Spec becomes a fully
+// resolved one.
+func (r *EtcdClusterReconciler) PopulateDefaultValues(ec *ecv1alpha1.EtcdCluster) {
+	if ec.Spec.ImageRegistry == "" {
+		ec.Spec.ImageRegistry = r.ImageRegistry
+	}
 }
 
 // buildReconcileClientTLS builds the operator's etcd-client TLS Config for a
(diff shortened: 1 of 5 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -165,9 +165,11 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
...
 	}
 
-	if ec.Spec.ImageRegistry == "" {
-		ec.Spec.ImageRegistry = r.ImageRegistry
-	}
+	// Apply operator-side defaults so downstream phases (hash, pod builder,
+	// cert wiring) all see consistent values regardless of which fields the
+	// user left empty. Defaults are in-memory only — the user-owned Spec is
+	// not persisted back, to avoid polluting the CR with operator policy.
+	FillEtcdClusterDefaultValues(ec, r.ImageRegistry)
```
ahrtr: The function name is a generic name, but it accepts a specific value `r.ImageRegistry`.   Can we rename it to something below? We can populate the `--service-dns-domain` inside it as well in your next PR.

```
(ec *ecv1alpha1.EtcdCluster) PopulateDefaultValues(r *EtcdClusterReconciler)
```

silentred: The signature has to be `(r *EtcdClusterReconciler) PopulateDefaultValues(ec *ecv1alpha1.EtcdCluster)`, otherwise there will be a circular pkg import.

internal/controller/etcdcluster_controller.go
```diff
@@ -165,9 +165,11 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		return nil, ctrl.Result{}, err
 	}
 
-	if ec.Spec.ImageRegistry == "" {
-		ec.Spec.ImageRegistry = r.ImageRegistry
-	}
+	// Apply operator-side defaults so downstream phases (hash, pod builder,
+	// cert wiring) all see consistent values regardless of which fields the
+	// user left empty. Defaults are in-memory only — the user-owned Spec is
+	// not persisted back, to avoid polluting the CR with operator policy.
```
ahrtr: The last sentence is a little confusing to me
```suggestion
	// Apply operator-side defaults so downstream phases (hash, pod builder,
	// cert wiring) all see consistent values regardless of which fields the
	// user left empty. 
```

ahrtr (approved):
LGTM with a minor comment

Issue #107, opened 2025-03-22
StorageSpec accessMode is not required but errors if not provided


SDBrett:
If an ectdCluster is configured with a storageSpec and the accessMode field is not configured the statefulSet cannot create pods and provide the below error.

`create Claim etcd-data-example-0 for Pod example-0 in StatefulSet example failed error: PersistentVolumeClaim "etcd-data-example-0" is invalid: spec.accessModes: Unsupported value: "": supported values: "ReadOnlyMany", "ReadWriteMany", "ReadWriteOnce", "ReadWriteOncePod"`

Example etcdCluster config to get the above error

```yaml
apiVersion: operator.etcd.io/v1alpha1
kind: EtcdCluster
metadata:
  name: example
  namespace: etcd-test
spec:
  size: 1
  version: v3.5.17
  etcdOptions:
    - "--experimental-peer-skip-client-san-verification"
  storageSpec:
    volumeSizeRequest: 100M
```

Snippet of statuefulSet created with above config

```yaml
kind: StatefulSet
apiVersion: apps/v1
metadata:
  name: example
  namespace: etcd-test
spec:
  serviceName: example
      spec:
        accessModes:
          - ''
        resources:
          limits:
            storage: 100M
          requests:
            storage: 100M
        volumeMode: Filesystem
      status:
        phase: Pending
```

shubhindia:
@SDBrett  I just tried with the same spec as yours and it seems to be working fine.
```
❯ k get etcdcluster -n etcd-test example -o yaml
apiVersion: operator.etcd.io/v1alpha1
kind: EtcdCluster
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operator.etcd.io/v1alpha1","kind":"EtcdCluster","metadata":{"annotations":{},"name":"example","namespace":"etcd-test"},"spec":{"etcdOptions":["--experimental-peer-skip-client-san-verification"],"size":1,"storageSpec":{"volumeSizeRequest":"100M"},"version":"v3.5.17"}}
  creationTimestamp: "2025-08-12T09:24:17Z"
  generation: 1
  name: example
  namespace: etcd-test
  resourceVersion: "522"
  uid: 9d07b710-0710-43b9-9e2a-27811c62d059
spec:
  etcdOptions:
  - --experimental-peer-skip-client-san-verification
  size: 1
  storageSpec:
    volumeSizeRequest: 100M
  version: v3.5.17
```

```
[...]

PR #452, opened 2026-08-05
test: cover storage access mode defaulting

mehrdadbn9:
`storageSpec.accessModes` is optional, and leaving it unset used to produce a volume claim template with `accessModes: [""]`, which the API server rejects, so every pod stayed pending. That is what #107 reported.

It was already fixed in 3905389 ("Defaults to ReadWriteOnce when it's empty or not provided"), which is why the recent attempts to reproduce it came back clean. Nothing covers that defaulting though, so it can be dropped again without a test going red - the e2e cases always set `accessModes: ReadWriteOnce` explicitly, so they never touch the empty path.

This adds a table-driven test over `createOrPatchStatefulSet` for the access mode branches:

- unset -> claim template gets `ReadWriteOnce` (the #107 case)
- explicit `ReadWriteOnce` -> kept
- `ReadWriteMany` + `pvcName` -> no claim template, pod volume references the named PVC
- `ReadWriteMany` without `pvcName` -> rejected
- unsupported mode (`ReadOnlyMany`) -> rejected

## Verification

Reverting 3905389 locally (putting `ec.Spec.StorageSpec.AccessModes` back in the claim template) turns exactly the first case red and leaves the other four green:

```
--- FAIL: TestCreateOrPatchStatefulSetStorageAccessModes/defaults_to_ReadWriteOnce_when_accessModes_is_not_provided
    Error: Not equal:
      expected: []v1.PersistentVolumeAccessMode{"ReadWriteOnce"}
      actual  : []v1.PersistentVolumeAccessMode{""}
```

which is the same `accessModes: ['']` from the original report. With the fix in place the whole package is green, and `make lint` reports 0 issues.

Test only, no production code touched.

Refs #107

### DIFF
--- a/internal/controller/utils_test.go
+++ b/internal/controller/utils_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


mehrdadbn9:
Friendly ping for review — this PR adds tests covering storage access-mode defaulting and has been green since it was opened. Happy to address any feedback. Thanks!

PR #453, opened 2026-08-06
Add MoveLeader function

nwnt:
As part of the custom controller to handle config update, we need to move the leader to a different existing member.  This PR is to add such functionality, with unit tests.  New test setup is also required as all of the existing unit tests only test a single member.  However, to move a leader, you need a cluster with more than one member.

@silentred @ahrtr

### DIFF
--- a/internal/etcdutils/etcdutils.go
+++ b/internal/etcdutils/etcdutils.go
@@ -114,8 +114,8 @@ func FindLeaderStatus(healthInfos []EpHealth, logger logr.Logger) (uint64, *clie
 	if leaderStatus != nil {
 		logger.Info("Leader found", "leaderID", leader)
 	}
-	return leader, leaderStatus
 
+	return leader, leaderStatus
 }
 
 func FindLearnerStatus(healthInfos []EpHealth, logger logr.Logger) (uint64, *clientv3.StatusResponse) {
@@ -247,5 +247,18 @@ func RemoveMember(cfg ClientConfig, memberID uint64) error {
 	defer func() { closeAndCancel(c, cancel) }()
 
 	_, err = c.MemberRemove(ctx, memberID)
+	return err
+}
+
+func MoveLeader(cfg ClientConfig, memberId uint64) error {
+	c, err := clientv3.New(cfg.buildConfig())
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer func() { closeAndCancel(c, cancel) }()
+
+	_, err = c.MoveLeader(ctx, memberId)
 	return err
 }


internal/etcdutils/etcdutils_test.go
```diff
@@ -362,3 +395,162 @@ func TestClientConfigBuild(t *testing.T) {
...
+	for _, s := range servers {
+		if uint64(s.Server.MemberID()) == leaderId {
+			count++
+		}
+	}
+	if count != 1 {
+		return fmt.Errorf("expected exactly one leader, found %d", count)
+	}
+	return nil
+}
```
ahrtr: I believe the implementation of this function can be reduced/simplified unless you want to return the leaderID.

I think it makes more sense to add a util function something like below, it returns error if zero or multiple leader found. It can be reused in multiple places in this PR.
```
func getLeader(servers []*embed.Etcd) (*embed.Etcd, error)
```


```suggestion
func ensureOneLeader(servers []*embed.Etcd) error {
	leaderCnt := 0
	for _, s := range servers {
		leader := s.Server.Leader()
		if leader == 0 {
			return fmt.Errorf("etcd %s has no leader", s.Config().Name)
		}
		leaderCnt++
	}
	if leaderCnt != 1 {
		return fmt.Errorf("multiple leaders found: %v", leaders)
	}
	return nil
}
```

nwnt: I had to slightly modify this as the suggested code keeps incrementing when a leader is found even if their leader is the same.

nwnt: Also added `getLeader` per your suggestion.

internal/etcdutils/etcdutils_test.go
```diff
@@ -362,3 +396,160 @@ func TestClientConfigBuild(t *testing.T) {
...
+			leaderCnt++
+		}
+	}
+	if leaderCnt != 1 {
+		return fmt.Errorf("multiple leaders found with leader count %d", leaderCnt)
+	}
+	return nil
+}
+
+func TestMoveLeader(t *testing.T) {
```
ahrtr: I feel like the test case you added in this PR is testing etcd itself instead of etcd-operator. The test test has nothing to do with etcd-operator.

Testing etcd itself isn't the scope of etcd-operator. We should focus on testing etcd-operator.

ahrtr:
Per https://github.com/etcd-io/etcd-operator/pull/453#discussion_r3731178355, I suggest that we only keep the function `MoveLeader` in etcdutil.go for now. We can add test case for etcd-operator when the function is used later in separate PR.

internal/etcdutils/etcdutils_test.go
nwnt: Yep, that's true.  Removed for now.

silentred (approved):
LGTM. Thanks.

ahrtr (approved):

PR #454, opened 2026-08-06
fix: reconcile image version changes without requiring cluster resize

### LINKED ISSUE #327: Updating etcdcluster manifest does not upgrade the cluster unless the number of members is also changed
To reproduce:

1. Install etcd-operator 0.2
2. Create a 3-node etcdcluster with etcd image 3.6.8
3. Update etcdcluster.yaml to use image 3.6.10
4. `kubectl apply` the new manifest from `sample_autocert_etcdcluster.yaml`
5. etcdcluster definition is updated
6. The Statefulset does NOT get updated.  No members actually get upgraded.
7. Try modifying etcdcluster-sample using `kubectl edit`
8. Statefulset still not updated
9. Edit the YAML file one more time, changing the number of members also
10. The statefulset will be updated, including the image version

It appears that changes made to the manifest for an existing cluster are not applied to the statefulset unless the user also changes the number of members.  This is a missing logic loop somewhere.

lavishpal:
This PR fixes #327 by adding `currentStsVersion()` helper to extract version from the running StatefulSet image and`stsImageMatchesCluster()` helper to compare running version against desired spec version. Also added image version check inside the size check block, if size matches but image has drifted, update the StatefulSet

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -267,6 +267,22 @@ func (r *EtcdClusterReconciler) performHealthChecks(ctx context.Context, s *reco
 	return nil
 }
 
+func currentStsVersion(sts *appsv1.StatefulSet) string {
+	if len(sts.Spec.Template.Spec.Containers) == 0 {
+		return ""
+	}
+	image := sts.Spec.Template.Spec.Containers[0].Image
+	idx := strings.Index(image, ":")
+	if idx == -1 {
+		return ""
+	}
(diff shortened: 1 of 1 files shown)


lavishpal:
cc: @ahrtr , @jberkus

ahrtr:
FYI. we are replacing the statefulsets with a customized controller. https://github.com/etcd-io/etcd-operator/issues/362

ahrtr:
Thanks for your effort anyway!

Issue #438, opened 2026-07-21
consider a RepairStrategy API for opinionated repairs


neolit123:
while the operator can have a canonical strategy how it would repair the etcd members, according to the etcd maintainers, for later versions users might have opinionated ideas what to do during member failures in the reconciliation loop.

to allow that it might be a good idea to introduce a RepairStratagy or a similar API per etcd Cluster early to allow later for the extension point.

i don't have examples of how the API could look like but one idea would be to define a set of ordered actions and priorities and allow a custom RepairStrategy to adjust the order and priority.

/kind feature
/priority backlog

ahrtr:
We need to a member replacement strategy to handle the following two cases. 
- When a member has issue/problem
- When rolling upgrade/update

I expect we follow the same pattern (e.g. delete old member & POD first, and add new member & POD later) to resolve both cases above for simplicity. We validate the existing data if present before creating the new member & POD.
- If the data has already corrupted, then cleanup the data & PV & PVC, and pre-prepare the PV/PVC for the new member/POD
- Otherwise, reuse the PV & data

All of the above is based on the assumption that the quorum is still satisfied. If quorum isn't satisfied, then we need a complete different approach to resolve it, e.g. forcibly creating a single node cluster first and adding other members later. 

We should generate a formal workflow / design-doc to clarify this later.

PR #456, opened 2026-08-13
Deliver a comprehensive design and the initial refactoring based on the design

ahrtr:
Introduced a new namespaced CRD, `EtcdMember`, representing exactly one etcd member/Pod slot (ordinal) belonging to an `EtcdCluster`. `EtcdClusterReconciler` now also owns `EtcdMember`. Each member's retry/lifecycle state is durably recorded on `EtcdMember.Status`, so a failed member can be retried across reconcile loops (and controller restarts) and, if retries are exhausted, replaced.

Cluster-wide `CORRUPT` and `NOSPACE` alarms, and total quorum loss, each get their own resumable remediation rather than being treated as ordinary Pod problems. Exactly one action is ever taken at a time per cluster; every other orchestration action (scale, upgrade, config update) waits until every existing member is `Ready`.

- The first commit includes the new workflows and designs
  - The three workflow diagrams (`reconcile_loop_v0.3.0`, `reconcile_cluster_v0.3.0` and `reconcile_member_v0.3.0`) are completely my manual work.
  - The design doc `etcd-member-lifecycle-and-self-healing-v0.3.0.md` was initially generated based on the three workflows above by AI.  Afterwards, it took me huge effort (~4 days) to review and modify the design doc to ensure it's completely align with the workflows above and meet my expectation.
- The second commit includes API changes, manifests/RBAC, and sample YAML updates.
  - Generated by AI. Afterwards, I manually updated some comments and re-generated the manifests.
- The third commit is refactoring based on the design.
  - Initially generated by AI. Afterwards, I reviewed and made some modification. This is basically just a framework/scaffold. There are lots of  inline TODO comments/placeholders, so that more contributors can get involved after this PR gets merged.

Notes to reivewers / contributors:
- Please check the workflows outlined in the **first comment** (and refer to the design doc if anything is unclear). Let me know if you have any questions or comment.
- The new workflows & design also resovles https://github.com/etcd-io/etcd-operator/issues/438. @neolit123 

cc @GunaKKIBM @hakman @ivanvc  @neolit123  @nwnt @silentred


api/v1alpha1/etcdmember_types.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright 2024.
```
neolit123:
```suggestion
Copyright 2026.
```

but more importantly there must be a Copyright owner on the same line.
e.g. 
`Copyright 2026 The etcd-operator Authors`

or remove the year like k8s did, the CNCF now allows it.

`Copyright The etcd-operator Authors`
https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.go.txt#L2

but without a copyright holder this stands no legal ground.

neolit123: i see this is a problem in all files in the repo.

neolit123: - https://github.com/etcd-io/etcd-operator/issues/457

api/v1alpha1/etcdmember_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// EtcdMemberReplacing means recreate retries are exhausted (or a CORRUPT alarm fired); the member is leaving, about to rejoin fresh.
+	EtcdMemberReplacing EtcdMemberPhase = "Replacing"
+	// EtcdMemberTerminating means DeletionTimestamp is set; the member is leaving the cluster for good.
+	EtcdMemberTerminating EtcdMemberPhase = "Terminating"
+)
+
+// EtcdMemberStatus defines the observed state of a single etcd member.
+type EtcdMemberStatus struct {
+	// +optional
+	Phase EtcdMemberPhase `json:"phase,omitempty"`
```
neolit123, line 61: shouldn't the phase be a condition?

api/v1alpha1/etcdmember_types.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// MemberID is the hex etcd member ID, recorded for
+	// observability/reporting only. EtcdClusterReconciler never uses this
+	// persisted value to decide anything — wherever a member's live etcd
+	// identity is needed (e.g. MemberRemove's target in the leave sequence),
+	// it's found by a live MemberList lookup keyed on this ordinal's
+	// deterministic peer URL, never by trusting this field is still
+	// current.
+	// +optional
+	MemberID string `json:"memberID,omitempty"`
```
neolit123, line 82: is it a good idea to also show the member address in the status?
otherwise one needs to lookup from ID, i guess.

api/v1alpha1/etcdmember_types.go
ahrtr: Updated.

Yes, we can update the copyright header in a separate PR on the main branch.

api/v1alpha1/etcdmember_types.go
ahrtr: The peer & client URLs are deterministic derived from the ordinal. We might not need to add them into the status.

api/v1alpha1/etcdmember_types.go
ahrtr: We depend on the Phase to drive the state machine, not just observational info. So a dedicated field might be better.

api/v1alpha1/etcdmember_types.go
neolit123: and it's not a breaking change to add them later, so that's fine.

api/v1alpha1/etcdmember_types.go
neolit123: makes sense.

api/v1alpha1/etcdmember_types.go
ahrtr: yes, it's just a minor change to add them later.

internal/controller/etcdmember.go
```diff
@@ -0,0 +1,156 @@
...
+	if err := controllerutil.SetControllerReference(ec, member, scheme); err != nil {
+		return nil, err
+	}
+	if err := c.Create(ctx, member); err != nil {
+		return nil, err
+	}
+
+	// Phase lives on the status subresource, so it has to be written in a
+	// separate call after the object exists.
+	member.Status.Phase = ecv1alpha1.EtcdMemberPending
```
silentred, line 151: If the operator crashes before persisting `member.Status.Phase`, could it fail to transition the member to the next state?

internal/controller/etcdcluster_controller.go
```diff
@@ -283,232 +305,275 @@ func (r *EtcdClusterReconciler) bootstrapCluster(ctx context.Context, s *reconci
...
+// the dispatcher's per-member-repair step (§4.9 item 5), not this phase.
+func (r *EtcdClusterReconciler) refreshClusterState(ctx context.Context, s *reconcileState) error {
 	logger := log.FromContext(ctx)
 	logger.Info("Now checking health of the cluster members")
 
 	var err error
 	s.memberListResp, s.memberHealth, err = healthCheck(s.cluster.Name, s.cluster.Namespace, s.pods, clusterTLSEnabled(s.cluster), s.tlsConfig, logger)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("health check failed: %w", err)
+		return fmt.Errorf("health check failed: %w", err)
```
silentred, line 327: If any member is unhealthy, the aggregated error will be non-nil. This prevents the workflow from reaching `r.dispatch()`, meaning unhealthy pods won't be repaired.

internal/controller/etcdmember.go
ahrtr: If the controller crashes right before updating the Phase, then when the controller gets started again, it will perform health check, decide what to do based on the live status.

internal/controller/etcdcluster_controller.go
ahrtr: Yes, we need to get each single member's health status. Note this PR is just a framework/scaffold, and there are lots of inline TODO comments, which we will need to resolve in following PRs.

When we resolve the Recreating case, we will resolve this specific problem.

ahrtr: @nwnt @silentred Please pay more attention to the workflow diagrams included in the first commit.

ahrtr:
Link to https://github.com/etcd-io/etcd-operator/issues/362

api/v1alpha1/etcdmember_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// EtcdMemberTerminating means DeletionTimestamp is set; the member is leaving the cluster for good.
+	EtcdMemberTerminating EtcdMemberPhase = "Terminating"
+)
+
+// EtcdMemberStatus defines the observed state of a single etcd member.
+type EtcdMemberStatus struct {
+	// +optional
+	Phase EtcdMemberPhase `json:"phase,omitempty"`
+
+	// PodName is this member's Pod, always "{clusterName}-{ordinal}" —
```
nwnt: Would it be better to name this as `MemberName` to keep it consistent with `MemberID` (like `etcdctl member list` would return member names and member ID)?  I'm thinking that if this is for observability, we should use the terms related to etcd.

api/v1alpha1/etcdmember_types.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// IsHealthy indicates whether the member's etcd process is healthy, as
+	// observed by a live check every reconcile.
+	IsHealthy bool `json:"isHealthy"`
+
+	// IsLearner indicates whether the member is currently a learner in the etcd cluster.
+	// +optional
+	IsLearner bool `json:"isLearner,omitempty"`
+
+	// IsLeader indicates whether this member is currently the cluster leader.
```
nwnt, line 96: What do you think about adding a field to tell what cluster ID it belongs to?

api/v1alpha1/etcdmember_types.go
ahrtr: Makes sense. Updated.

api/v1alpha1/etcdmember_types.go
ahrtr: The MembeName (previous PodName) already includes the clusterName. It should be already good enough.

ahrtr: Let's keep it as it's for now, and revisit this if we indeed want it. it's just a minor change.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
```diff
@@ -0,0 +1,1255 @@
...
+Introduce a new namespaced CRD, `EtcdMember`, representing exactly one etcd
+member/Pod slot (ordinal) belonging to an `EtcdCluster`. `EtcdClusterReconciler`
+— unchanged as the *only* reconciler in this design — now also owns
+`EtcdMember`. Each member's retry/lifecycle state is durably recorded on
+`EtcdMember.Status`, so a failed member can be retried across reconcile
+loops (and controller restarts) and, if retries are exhausted, replaced.
+Cluster-wide `CORRUPT` and `NOSPACE` alarms, and total quorum loss, each get
+their own resumable remediation rather than being treated as ordinary Pod
+problems. Exactly one action is ever taken at a time per cluster; every
+other orchestration action (scale, upgrade, config update) waits until
```
GunaKKIBM, line 30: Are multiple EtcdMembers reconciled at the same time? If we try increasing the maxReconciles? As of now in the code, it doesn't seem like that. 
But, I think we can choose to reconcile for multiple EtcdMembers, hence delivering low latency for the cluster to be up.

GunaKKIBM: This might lead to a problem w.r.t to assigning the ordinal, but that section of the code can be locked.

GunaKKIBM: Sorry, the comment might seem confusing, My question is, is there a scenario, where multiple members can be reconciled at the same time?

ahrtr: all good questions, cc @nwnt and @silentred 

> Are multiple EtcdMembers reconciled at the same time?

We will never reconcile multiple EtcdMember at the same time. We don't need to do that, as usually an etcd cluster has only 3 or 5 members, we don't need to worry about performance issue. Also we will resolve the performance issue using controller-runtime's `MaxConcurrentReconciles` option.  Please read the option B in the section `## 5. Alternatives considered`

> This might lead to a problem w.r.t to assigning the ordinal

For ordinal management, please read
- item 1 & 3 in the section `## 3. Requirements`
- section `### 4.4 Ordinal allocation & closing gaps`

ahrtr: > Also we will resolve the performance issue using controller-runtime's `MaxConcurrentReconciles` option.

To be clearer, we will support concurrently reconcile multiple EtcdClusters.

ahrtr: > We will never reconcile multiple EtcdMember at the same time.

We will never reconcile multiple EtcdMember **of the same EtcdCluster** at the same time. 

But we will support reconciling multiple Etcdcluster at the same time.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
```diff
@@ -0,0 +1,1255 @@
...
+3. **Scale-out allocates the lowest missing ordinal, or `max(existing ordinals) + 1` if there is no gap to reuse.**
+4. Support fixing failed members: the default remedy is delete-and-recreate the
+   Pod with the **latest config** from `EtcdCluster.Spec` — except `Version`,
+   which stays at the member's current running version rather than jumping
+   to the upgrade target. Version changes are rolled out exclusively by the
+   separate upgrade phase, one member at a time.
+5. If recreating doesn't help, fall back to a blunter remedy: remove the etcd member,
+   clean up its data and TLS certificate, and add it back as a new member
+   with a fresh Pod. Simple heuristic for "recreating didn't help": three
+   failed recreate attempts.
```
silentred, line 116: Maybe we could just observe restart count of the recreated member pod, instead of recreating one same pod three times.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
```diff
@@ -0,0 +1,1255 @@
...
+## 3. Requirements
+
+1. **The controller itself never creates an ordinal gap**. A gap can
+   still appear from out-of-band interference (e.g. a human manually
+   deleting a non-highest-ordinal `EtcdMember`); when that happens, it
+   isn't hunted down and fixed on its own, but scale-out is guaranteed to
+   eventually fill it rather than allocate a new, higher ordinal.
+2. **Scale-in removes the highest ordinal first.**
+3. **Scale-out allocates the lowest missing ordinal, or `max(existing ordinals) + 1` if there is no gap to reuse.**
+4. Support fixing failed members: the default remedy is delete-and-recreate the
```
silentred, line 101: > the default remedy

To be precise, only delete-and-recreate the pod, no touch of PVC and PV, if I understand correctly.
For users choosing local PVs or no PV at all, the default remedy may not help for recovery.
In the future, can we add an option to bypass this step, directly to "blunter remedy".

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
```diff
@@ -0,0 +1,1255 @@
...
+   then the PVC if not already present — unconditionally, for every
+   member, before anything etcd-side happens.
+2. If this is ordinal 0 and no other member has ever joined (the bootstrap
+   case), skip straight to step 4 — the first member starts as a full
+   voting member, never a learner.
+3. Otherwise, check (a live `MemberList` call) whether a member with this
+   ordinal's deterministic peer URL already exists; if not, call
+   `MemberAdd`(learner) first — same as today's `scaleCluster`.
+4. Start the Pod if not already present (`--initial-cluster-state=new`
+   only for the bootstrap member, `existing` otherwise), bumping
```
silentred, line 581: How do we tell if the cluster in boostraping?

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
ahrtr: We can't depend on the restart count to make decision.
- It doesn't reset as long as the POD isn't restarted (it only counts the container's restart count)
- It's reset to 0 once the POD is restarted.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
ahrtr: > To be precise, only delete-and-recreate the pod, no touch of PVC and PV, if I understand correctly.

Yes. Please see the reconcile_member workflow diagram.

 

> For users choosing local PVs or no PV at all, the default remedy may not help for recovery.
> In the future, can we add an option to bypass this step, directly to "blunter remedy".

I do not quite follow this comment. But we are open to any discussion in the future.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
ahrtr: It's simple, when there is zero member, then it's bootstrap.

docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md
ahrtr: > > For users choosing local PVs or no PV at all, the default remedy may not help for recovery.
> > In the future, can we add an option to bypass this step, directly to "blunter remedy".
> 
> I do not quite follow this comment. But we are open to any discussion in the future.

Note once we have deterministic way to know the recreating doesn't help, then we can go to Replacing directly. Currently we only one such case: data corruption.  We can add more such case if any.

ahrtr:
Comments:
- support backoff for recreating or replacing cases.
- support upgrade first in case we run into an etcd bug

jberkus:
Comments/feedback:

- we will want to eventually have backoff for Replacement; if it's actually impossible to Replace the failed member, we don't want to be trying every 3 seconds forever
- We will eventually want to support two manual recovery cases where we take an action despite all members not being Ready:

1. Patch application: if the patch update is related to fixing the customer, the administrator should be able to somehow override the standard reconciliation loop in order to force a version update even though all members are not Ready.
2. Shrinking clusters: if the reason for some members not being Ready is storage/resource problems, the administrator should have some way to force shrinking the cluster even though all members are not ready.  In this case, the operator should start by terminating members that are not Ready.

In both of these cases, I am suggesting that the user should have to take some action to initiate one of these special recovery modes.  This might suggest that having a single auto-recovery flag isn't enough; we may need to have specific flags for types of cluster failures the user wants to recover from.

ahrtr:
TODO:
for scale in, 
- probably we should remove the unhealthy member instead of the one with the highest ordinal
- should we handle scale in first ( early)?

silentred:
> * probably we should remove the unhealthy member instead of the one with the highest ordinal

Based on [reconcile loop](https://github.com/etcd-io/etcd-operator/blob/66b1926874f8e5809db61cebb5157b4435ce5b03/docs/design/reconcile_loop_v0.3.0.png), we do "Health check & Failure Recovery" before scale-in. 
But this does not mean there is no unhealthy member when the reconcile loop has just run into the scale-in part.
My suggestion is that in the early stage of scale-in, we can do health check again. If there is unhealthy member, end the current loop, in order to reuse the "Health check & Failure Recovery" in the next loop.

docs/design/reconcile_member_v0.3.0.uxf
```diff
@@ -0,0 +1,1028 @@
...
+    <id>UMLSpecialState</id>
+    <coordinates>
+      <x>820</x>
+      <y>740</y>
+      <w>120</w>
+      <h>60</h>
+    </coordinates>
+    <panel_attributes>type=decision
+Is the
+POD timedout?</panel_attributes>
```
silentred, line 781: The proper time to start a etcd pod usually differs by the db size of the etcd cluster. Maybe we should leave the timeout duration customizable to the users?

ahrtr: > Maybe we should leave the timeout duration customizable to the users?

It's open to discussion in next step.

It's already covered in the item 1 in section `## 7. Open questions`.

ahrtr:
Updates:
- Added a new phase (`Is the EtcdCluster paused`) into the workflow reconcile_loop_v0.3.0
- Added a new flag `AllowVersionUpgradeOnRepair` into `EtcdClusterSpec`. Once enabled (disabled by default), the controller will use both latest config and latest version when recreating a POD. See item 4 in `## 3. Requirements` and section 4.6
- For the scale-in discussion, let's keep it as it's for now. We still fix any failed member first, and ensure all existing members are ready/healthy before performing any normal orchestration actions (including scale in), reasons:
  - if the cluster isn't healthy (quorum lost), there is no way to perform any action, including scale-in, the controller will have to fix the unhealthy EtcdCluster first.
  - If quorum isn't lost, the etcd cluster remains healthy and can still serve requests, even if a minority of members are down. This case is debatable, but it should be fine for now. We expect failure-recovery—or the operator/admin for underlying environment issues—to fix unhealthy members first. We'll keep it as is to avoid premature optimization and revisit this in the future.

cc @ArkaSaha30  @GunaKKIBM @ivanvc  @jberkus @nwnt @silentred  

If there is no any objection, let's merge this PR into the branch customized-controller first. Afterwards, I will merge the customized-controller into main as we discussed in the community meeting.

jberkus:
No objection.  I'll create the scale-down item as its own issue to track it for future releases.

nwnt (approved):
No objection to this.  Thanks Benjamin.

silentred (approved):
LGTM

ivanvc (approved):
No objections. LGTM after the walkthrough in the last two community meetings. Thanks, @ahrtr.

ahrtr:
thx all for the approvals.

Issue #463, opened 2026-08-18
Consider supporting scale-down for a cluster with Not Ready members


jberkus:
There is one specific circumstance where we might want to support scaling down a cluster, even though it has members that are Not Ready:

- It is a 5+ Member cluster
- There are 3 Ready nodes
- Two nodes are Not Ready because the user has resource problems in the Kubernetes cluster
- The user would like to scale down to a 3-node cluster

This is a fairly specific failure case, so we are not building it into release 0.3.  This issue is being created to track the idea/possible requirement for the future.  Do not start work on it at this time.

Ref:  https://github.com/etcd-io/etcd-operator/pull/456#issuecomment-5332566251

PR #466, opened 2026-08-20
Update the design doc and workflow

ahrtr:
Changes:
1. Cleanup any `EtcdMember` with `DeletionTimestamp` set before any repair, as it doesn't make sense to fix a member if it's being deleted.
2. Move the pause check below the validation so that status refreshing still runs even when paused.
3. Remove the unnecessary "Bootstrap etcd cluster" phase from the workflow, as it's already covered by the normal scale out case

cc @ArkaSaha30 @GunaKKIBM @ivanvc  @nwnt @silentred


ahrtr:
> 1. Cleanup any `EtcdMember` with `DeletionTimestamp` set before any repair, as it doesn't make sense to fix a member if it's being deleted.

I believe it should be able to resolve https://github.com/etcd-io/etcd-operator/issues/463 as well. If the etcd-opeartor can't handle the normal scale in due to unhealthy member, then users can manually remove the EtcdMembers that they want to shrink.

internal/controller/etcdcluster_controller.go
silentred: > (§4.9 item 5)

The per-member repair step was renumbered from item 5 → item 6

docs/design/reconcile_loop_v0.3.0.uxf
```diff
@@ -1492,44 +1366,47 @@ bg=blue</panel_attributes>
...
+      <x>930</x>
+      <y>1680</y>
+      <w>290</w>
+      <h>80</h>
     </coordinates>
-    <panel_attributes>unfinished scale in
+    <panel_attributes>check the DeletionTimestamp.
+two cases: 
+- unfinished scale in
+- users manualy remove an EtcdMember
```
silentred: a typo "manually"

silentred (commented):
Looks good overall. Just left 2 comments.

docs/design/reconcile_loop_v0.3.0.uxf
ahrtr: Good catch! Updated & thx

internal/controller/etcdcluster_controller.go
ahrtr: updated & thx

silentred (approved):

internal/controller/etcdcluster_controller.go
```diff
@@ -364,47 +372,53 @@ func (r *EtcdClusterReconciler) dispatch(ctx context.Context, s *reconcileState)
...
-	// 3. CORRUPT alarm on some member (§4.6/§4.7).
-	// TODO: §4.9 item 3 — force the tagged member to Phase: Replacing
+	// 3. Clean up any EtcdMember already Terminating (§4.6/§4.9 item 3,
+	// reconcile_loop_v0.3.0.png). Ranked here — above CORRUPT/NOSPACE/
+	// per-member repair — because a member that's already leaving (from
+	// scale-in, or a user manually removing it) is a settled matter: it's
+	// never a repair candidate, only ever touches its own Pod/PVC/cert/
+	// membership state, and shouldn't linger behind other remediation.
+	for i := range s.members {
+		m := &s.members[i]
```
GunaKKIBM, line 382: Just a minor enhancement, I don't think, the pointer needs to be stored here, it is required for the function call - https://github.com/etcd-io/etcd-operator/pull/466/changes#diff-7e023adf1c245d768f95e3f174cab882db08efd47023a9ce81e29a3a7767c12aR388 - clearMemberFinalizer. 

it can be
`
for i. member := range s.members {
`

Remove 
`m := &s.members[i]`

And
` r.clearMemberFinalizer(ctx, &m)`

ahrtr: > for i. member := range s.members {

There is a minor difference. The member is copy by value, so any updating on the object won't change anything  in s.members. We still expect the local s.members is in sync with the latest object info, including the finalizer, although it isn't used any where for now.

GunaKKIBM: Makes sense. Thank you!

PR #478, opened 2026-08-24
feat: support normal member scale-out via the shared member lifecycle

### LINKED ISSUE #469: Handle the scale out and normal provisioning case
When we create a new EtcdCluster, we create an EtcdMember for each member, and provision each EtcdMember one by one. **We can resolve the PV/PVC stuff separately.**

The related code is [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L422-L430) and [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L555)

Test cases:
- create a one member cluster, and scale out to 3 members
- create a three members cluster, and scale out to 5 members

silentred:
Resolve #469 , Link #362

Design notes:
- Why **empty string** is a valid provisioning phase of EtcdMember: object create and status write in `createEtcdMember` are two calls, and a crash in between leaves a phase-less object that must resume, not restart. See the comments in `reconcileEtcdMember` and `isProvisioningPhase`.
- `Owns(&corev1.Pod{})` covers legacy pods still controlled directly by the EtcdCluster. I am not sure if we should maintain the backward compatability. 
- Why `Replacing` counts as a provisioning phase: the desgin doc  defines it as "leaving, about to rejoin fresh" — the rejoin half reuses the join ladder.

/cc @ahrtr @nwnt @ArkaSaha30 @GunaKKIBM

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -206,7 +206,7 @@ func (r *EtcdClusterReconciler) fetchAndValidateState(ctx context.Context, req c
 		return nil, ctrl.Result{RequeueAfter: requeueDuration}, nil
 	}
 
-	pods, err := listOwnedPods(ctx, r.Client, ec)
+	pods, err := listOwnedPods(ctx, r.Client, ec, members)
 	if err != nil {
 		logger.Error(err, "Failed to list pods. Requesting requeue")
 		return nil, ctrl.Result{RequeueAfter: requeueDuration}, nil
@@ -301,8 +301,8 @@ func (r *EtcdClusterReconciler) ensureClusterPrereqs(ctx context.Context, s *rec
 			logger.Error(err, "Failed to create Client Certificate.")
 		}
 		// Server/peer certs must exist before buildReconcileClientTLS below reads
-		// the server cert Secret. createMemberPod also calls this (idempotently)
-		// once a member's pod is created.
+		// the server cert Secret. reconcileProvisioning also calls this
+		// (idempotently) before creating a member's Pod.
 		// TODO: buildReconcileClientTLS reuses this member server
 		// certificate as the operator's own client identity (see its comment
 		// in utils.go) — that's the wrong cert for the operator to depend on;
@@ -370,11 +370,14 @@ func (r *EtcdClusterReconciler) refreshClusterState(ctx context.Context, s *reco
 // to the next; only an item that actually takes a mutating action ends the
 // loop.
 //
-// Steps 4-8 are TODO no-ops in this milestone (M2) — the mechanics they'd
+// Steps 4-7 are TODO no-ops in this milestone (M2) — the mechanics they'd
 // trigger (join/promote/repair/leave, CORRUPT/NOSPACE remediation,
 // lost-quorum recovery) land in M3-M5 — but the detection that decides
 // *whether* a step claims the loop is real, so the priority order itself is
 // reviewable now, ahead of any of that behavior landing.
+// Item 8 is partially implemented: the provisioning case reconciles one
+// EtcdMember at a time until it is Ready; the replacing case (§4.9 item 8's
+// second bullet) lands in a follow-up PR.
 func (r *EtcdClusterReconciler) dispatch(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
@@ -439,13 +442,10 @@ func (r *EtcdClusterReconciler) dispatch(ctx context.Context, s *reconcileState)
 	// and the cluster is still unhealthy (M5).
 
 	// 8. Advance whatever's left not-Ready (Pending/Provisioning/Replacing).
-	// An existing learner always wins this slot (requirement 11).
-	if notReady := pickNotReadyMember(s.members); notReady != nil {
-		// TODO: §4.9 item 8 — promote if a caught-up learner,
-		// otherwise continue the member's join/replace progress (§4.6, M3).
-		logger.Info("Member not Ready; join/promotion not implemented yet",
-			"member", notReady.Name, "phase", notReady.Status.Phase)
-		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	// An existing learner always wins this slot (requirement 11); with more
+	// than one not-ready member but no learner, the lowest ordinal wins.
+	if notReady := pickNotReadyMember(s); notReady != nil {
+		return r.reconcileEtcdMember(ctx, s, notReady)
 	}
 
 	// 9. Everything existing is Ready (§4.2's gate), so update-config/scale/upgrade
@@ -537,11 +537,8 @@ func (r *EtcdClusterReconciler) updateConfig(ctx context.Context, s *reconcileSt
 // the lowest gap, else max+1) and always removing the highest ordinal first
 // on scale-in (requirement 2).
 //
-// Creating the EtcdMember object — and, for ordinal 0, its cert/PVC/Pod
-// directly (the bootstrap special case, §4.6 step 2) — is the only mechanic
-// this milestone (M2) implements. For every other ordinal, the general join
-// mechanics (MemberAdd + cert/PVC/Pod) are §4.6's TODO (M3), so a
-// newly-created EtcdMember simply sits at Phase: Pending until then. Scale-in
+// Scale-out only creates one Pending EtcdMember. The shared member lifecycle
+// owns bootstrap, member addition, Pod creation, and learner promotion. Scale-in
 // deletes the EtcdMember directly; the finalizer (§4.3) blocks its actual
 // removal until the Terminating leave sequence (§4.6, M3) runs.
 func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
@@ -558,20 +555,8 @@ func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileSt
 	if currentCount < desiredSize {
 		ordinal := nextOrdinal(memberOrdinals(s.members))
 		logger.Info("[Scale out] creating a new EtcdMember", "ordinal", ordinal)
-		member, err := createEtcdMember(ctx, r.Client, s.cluster, ordinal, r.Scheme)
-		if err != nil {
+		if _, err := createEtcdMember(ctx, r.Client, s.cluster, ordinal, r.Scheme); err != nil {
 			return ctrl.Result{}, err
-		}
-
-		if ordinal == 0 {
-			// Bootstrap special case: the first member starts as a full
-			// voting member with its Pod created directly — there's no
-			// cluster yet to MemberAdd(learner) into.
-			if err := createMemberPod(ctx, logger, r.Client, s.cluster, ordinal, r.Scheme); err != nil {
-				return ctrl.Result{}, err
-			}
-		} else {
-			logger.Info("Join mechanics not implemented yet; member stays Pending", "member", member.Name)
 		}
 		return ctrl.Result{RequeueAfter: requeueDuration}, nil
 	}
@@ -785,8 +770,6 @@ func (r *EtcdClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&ecv1alpha1.EtcdCluster{}).
 		Owns(&ecv1alpha1.EtcdMember{}).
-		Owns(&corev1.Pod{}).
-		Owns(&corev1.PersistentVolumeClaim{}).
 		Owns(&corev1.Service{})
 
 	if isCertManagerCRDPresent(mgr) {
--- a/internal/controller/etcdcluster_controller_test.go
+++ b/internal/controller/etcdcluster_controller_test.go
@@ -34,16 +34,35 @@ import (
 )
 
 // TestFetchAndValidateState verifies the fetchAndValidateState helper across
-// a range of conditions (missing cluster, no pods, pods owned by this cluster,
-// and pods owned by a different cluster).
+// a range of conditions (missing cluster, no pods, pods owned via their
+// EtcdMember by this cluster, and pods owned via a member of a different
+// cluster).
 func TestFetchAndValidateState(t *testing.T) {
 	scheme := runtime.NewScheme()
 	_ = corev1.AddToScheme(scheme)
 	_ = ecv1alpha1.AddToScheme(scheme)
 
-	// helper to build a minimal owned pod with a specific etcd image tag.
-	ownedPod := func(clusterName, namespace, uid, imageTag string) *corev1.Pod {
-		return &corev1.Pod{
+	// helper to build a minimal EtcdMember together with its Pod (controlled
+	// by that member), running a specific etcd image tag. The member's
+	// controller reference points at an EtcdCluster with the given UID.
+	memberAndPod := func(clusterName, namespace, clusterUID, imageTag string) (*ecv1alpha1.EtcdMember, *corev1.Pod) {
+		member := &ecv1alpha1.EtcdMember{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      clusterName + "-0",
+				Namespace: namespace,
+				Labels:    clusterNameLabels(clusterName),
+				UID:       types.UID("member-" + clusterName + "-0"),
+				OwnerReferences: []metav1.OwnerReference{{
+					APIVersion: ecv1alpha1.GroupVersion.String(),
+					Kind:       "EtcdCluster",
+					Name:       clusterName,
+					UID:        types.UID(clusterUID),
+					Controller: new(true),
+				}},
+			},
+			Spec: ecv1alpha1.EtcdMemberSpec{ClusterName: clusterName, Ordinal: 0},
+		}
+		pod := &corev1.Pod{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      clusterName + "-0",
 				Namespace: namespace,
@@ -53,10 +72,10 @@ func TestFetchAndValidateState(t *testing.T) {
 				},
 				OwnerReferences: []metav1.OwnerReference{{
 					APIVersion: ecv1alpha1.GroupVersion.String(),
-					Kind:       "EtcdCluster",
-					Name:       clusterName,
-					UID:        types.UID(uid),
-					Controller: pointerToBool(true),
+					Kind:       "EtcdMember",
+					Name:       member.Name,
+					UID:        member.UID,
+					Controller: new(true),
 				}},
 			},
 			Spec: corev1.PodSpec{
@@ -65,14 +84,18 @@ func TestFetchAndValidateState(t *testing.T) {
 				},
 			},
 		}
+		return member, pod
 	}
+	ownedMember, ownedPod := memberAndPod("etcd", "default", "2", "3.5.17")
+	foreignMember, foreignPod := memberAndPod("etcd", "default", "other-uid", "3.5.17")
 
 	cases := []struct {
-		name   string
-		req    ctrl.Request
-		ec     *ecv1alpha1.EtcdCluster
-		pods   []*corev1.Pod
-		assert func(t *testing.T, state *reconcileState, res ctrl.Result, err error)
+		name    string
+		req     ctrl.Request
+		ec      *ecv1alpha1.EtcdCluster
+		members []*ecv1alpha1.EtcdMember
+		pods    []*corev1.Pod
+		assert  func(t *testing.T, state *reconcileState, res ctrl.Result, err error)
 	}{
 		{
 			name: "EtcdCluster Not Found",
@@ -104,8 +127,9 @@ func TestFetchAndValidateState(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: "etcd", Namespace: "default", UID: "2"},
 				Spec:       ecv1alpha1.EtcdClusterSpec{Size: 1, Version: "3.5.17"},
 			},
-			pods: []*corev1.Pod{ownedPod("etcd", "default", "2", "3.5.17")},
-			req:  ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
+			members: []*ecv1alpha1.EtcdMember{ownedMember},
+			pods:    []*corev1.Pod{ownedPod},
+			req:     ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
 			assert: func(t *testing.T, state *reconcileState, res ctrl.Result, err error) {
 				require.NotNil(t, state)
 				assert.Equal(t, "etcd", state.cluster.Name)
@@ -120,9 +144,11 @@ func TestFetchAndValidateState(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: "etcd", Namespace: "default", UID: "3"},
 				Spec:       ecv1alpha1.EtcdClusterSpec{Size: 1, Version: "3.5.17"},
 			},
-			// Pod has different UID owner → filtered out by listOwnedPods.
-			pods: []*corev1.Pod{ownedPod("etcd", "default", "other-uid", "3.5.17")},
-			req:  ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
+			// The Pod's controlling member belongs to a different cluster
+			// (UID mismatch) → filtered out by listOwnedPods.
+			members: []*ecv1alpha1.EtcdMember{foreignMember},
+			pods:    []*corev1.Pod{foreignPod},
+			req:     ctrl.Request{NamespacedName: types.NamespacedName{Name: "etcd", Namespace: "default"}},
 			assert: func(t *testing.T, state *reconcileState, res ctrl.Result, err error) {
 				require.NotNil(t, state)
 				assert.Empty(t, state.pods)
@@ -139,6 +165,9 @@ func TestFetchAndValidateState(t *testing.T) {
 			objs := []client.Object{}
 			if tc.ec != nil {
 				objs = append(objs, tc.ec)
+			}
+			for _, member := range tc.members {
+				objs = append(objs, member)
 			}
 			for _, pod := range tc.pods {
 				objs = append(objs, pod)
@@ -303,11 +332,10 @@ func TestEnsureClusterPrereqs(t *testing.T) {
 	})
 }
 
-// TestScaleClusterBootstrap verifies the scale-out dispatcher step's
-// bootstrap special case (§4.6 step 2): with zero existing members, it
-// creates EtcdMember ordinal 0 and its Pod directly. For any other ordinal,
-// only the EtcdMember shell is created — join mechanics are M3's TODO.
-func TestScaleClusterBootstrap(t *testing.T) {
+// TestScaleCluster verifies that scaling only changes the desired set of
+// EtcdMember objects. Provisioning is owned by reconcileEtcdMember and must
+// never run from this policy-level helper.
+func TestScaleCluster(t *testing.T) {
 	scheme := runtime.NewScheme()
 	_ = corev1.AddToScheme(scheme)
 	_ = ecv1alpha1.AddToScheme(scheme)
@@ -324,7 +352,7 @@ func TestScaleClusterBootstrap(t *testing.T) {
 		},
 	}
 
-	t.Run("No members — creates EtcdMember ordinal 0 and its Pod", func(t *testing.T) {
+	t.Run("`scaleCluster` creates exactly one Pending member without a Pod", func(t *testing.T) {
 		ctx := t.Context()
 		fakeClient := fake.NewClientBuilder().
 			WithScheme(scheme).
@@ -345,11 +373,15 @@ func TestScaleClusterBootstrap(t *testing.T) {
 		assert.Contains(t, member.Finalizers, memberCleanupFinalizer)
 		require.Len(t, member.OwnerReferences, 1)
 		assert.Equal(t, ec.Name, member.OwnerReferences[0].Name)
+		assert.Equal(t, ec.Name, member.Labels[clusterNameLabel])
+
+		members := &ecv1alpha1.EtcdMemberList{}
+		require.NoError(t, fakeClient.List(ctx, members, client.InNamespace(ec.Namespace)))
+		assert.Len(t, members.Items, 1)
 
 		pod := &corev1.Pod{}
-		require.NoError(t, fakeClient.Get(ctx, client.ObjectKey{Name: "etcd-0", Namespace: ec.Namespace}, pod))
-		require.Len(t, pod.OwnerReferences, 1)
-		assert.Equal(t, ec.Name, pod.OwnerReferences[0].Name)
+		err = fakeClient.Get(ctx, client.ObjectKey{Name: "etcd-0", Namespace: ec.Namespace}, pod)
+		assert.True(t, apierrors.IsNotFound(err), "scaleCluster must not provision the bootstrap Pod")
 	})
 
 	t.Run("One ready member — creates EtcdMember shell only, no Pod", func(t *testing.T) {
@@ -364,7 +396,7 @@ func TestScaleClusterBootstrap(t *testing.T) {
 					Kind:       "EtcdCluster",
 					Name:       ec.Name,
 					UID:        ec.UID,
-					Controller: pointerToBool(true),
+					Controller: new(true),
 				}},
 			},
 			Spec:   ecv1alpha1.EtcdMemberSpec{ClusterName: ec.Name, Ordinal: 0, Version: ec.Spec.Version},
@@ -501,7 +533,11 @@ func TestDispatch(t *testing.T) {
 			Status:     ecv1alpha1.EtcdMemberStatus{Phase: ecv1alpha1.EtcdMemberPending},
 		}
 
-		fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(ec, &pending).Build()
+		fakeClient := fake.NewClientBuilder().
+			WithScheme(scheme).
+			WithStatusSubresource(&ecv1alpha1.EtcdMember{}).
+			WithObjects(ec, &pending).
+			Build()
 		r := &EtcdClusterReconciler{Client: fakeClient, Scheme: scheme}
 		state := &reconcileState{cluster: ec, members: []ecv1alpha1.EtcdMember{pending}}
 
@@ -511,7 +547,8 @@ func TestDispatch(t *testing.T) {
 
 		list := &ecv1alpha1.EtcdMemberList{}
 		require.NoError(t, fakeClient.List(ctx, list, client.InNamespace(ec.Namespace)))
-		assert.Len(t, list.Items, 1)
+		require.Len(t, list.Items, 1)
+		assert.Equal(t, ecv1alpha1.EtcdMemberProvisioning, list.Items[0].Status.Phase)
 	})
 
 	t.Run("Falls through to scale-out when everything is Ready", func(t *testing.T) {
@@ -526,7 +563,7 @@ func TestDispatch(t *testing.T) {
 					Kind:       "EtcdCluster",
 					Name:       ec.Name,
 					UID:        ec.UID,
-					Controller: pointerToBool(true),
+					Controller: new(true),
 				}},
 			},
 			Spec:   ecv1alpha1.EtcdMemberSpec{ClusterName: ec.Name, Ordinal: 0, Version: ec.Spec.Version},
@@ -550,6 +587,65 @@ func TestDispatch(t *testing.T) {
 		require.NoError(t, fakeClient.Get(ctx, client.ObjectKey{Name: "etcd-1", Namespace: ec.Namespace}, member))
 		assert.Equal(t, 1, member.Spec.Ordinal)
 	})
+}
+
+func TestGapAwareScaleOutCreatesLowestMissingOrdinal(t *testing.T) {
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+	require.NoError(t, ecv1alpha1.AddToScheme(scheme))
+
+	clusterUID := types.UID("cluster-uid")
+	cluster := &ecv1alpha1.EtcdCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: "etcd", Namespace: "default", UID: clusterUID},
+		Spec:       ecv1alpha1.EtcdClusterSpec{Size: 3, Version: "3.5.17"},
+	}
+	readyMember := func(ordinal int) ecv1alpha1.EtcdMember {
+		return ecv1alpha1.EtcdMember{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      etcdMemberName(cluster.Name, ordinal),
+				Namespace: cluster.Namespace,
+			},
+			Spec: ecv1alpha1.EtcdMemberSpec{
+				ClusterName: cluster.Name,
+				Ordinal:     ordinal,
+				Version:     cluster.Spec.Version,
+			},
+			Status: ecv1alpha1.EtcdMemberStatus{Phase: ecv1alpha1.EtcdMemberReady},
+		}
+	}
+	member0 := readyMember(0)
+	member2 := readyMember(2)
+	pod0 := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: member0.Name, Namespace: cluster.Namespace}}
+	pod2 := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: member2.Name, Namespace: cluster.Namespace}}
+	fakeClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithStatusSubresource(&ecv1alpha1.EtcdMember{}).
+		WithObjects(cluster, &member0, &member2, pod0, pod2).
+		Build()
+	reconciler := &EtcdClusterReconciler{Client: fakeClient, Scheme: scheme}
+	state := &reconcileState{
+		cluster: cluster,
+		members: []ecv1alpha1.EtcdMember{member0, member2},
+		pods:    []*corev1.Pod{pod0, pod2},
+	}
+
+	result, err := reconciler.scaleCluster(t.Context(), state)
+	require.NoError(t, err)
+	assert.Equal(t, ctrl.Result{RequeueAfter: requeueDuration}, result)
+	target := &ecv1alpha1.EtcdMember{}
+	require.NoError(t, fakeClient.Get(t.Context(), client.ObjectKey{
+		Name: "etcd-1", Namespace: cluster.Namespace,
+	}, target))
+	require.Len(t, target.OwnerReferences, 1)
+	require.Equal(t, clusterUID, target.OwnerReferences[0].UID)
+	assert.Equal(t, 1, target.Spec.Ordinal)
+
+	// First lifecycle pass durably enters Provisioning.
+	state.members = []ecv1alpha1.EtcdMember{member0, *target, member2}
+	result, err = reconciler.reconcileEtcdMember(t.Context(), state, target)
+	require.NoError(t, err)
+	assert.Equal(t, ctrl.Result{RequeueAfter: requeueDuration}, result)
+	assert.Equal(t, ecv1alpha1.EtcdMemberProvisioning, target.Status.Phase)
 }
 
 // TestReconcilePaused verifies requirement 15's pause gate: Spec.Paused
--- a/internal/controller/etcdmember.go
+++ b/internal/controller/etcdmember.go
@@ -18,13 +18,21 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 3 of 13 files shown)


ahrtr:
@silentred please rebase this PR, I will take a look later.

silentred:
> @silentred please rebase this PR, I will take a look later.

OK. Do you need me to split this into smaller PRs?

ahrtr:
Please rebase first.  I will take a look sometime later. Afterwards, let's discuss how to proceed.

silentred:
> Please rebase first. I will take a look sometime later. Afterwards, let's discuss how to proceed.

Done

internal/controller/etcdcluster_controller.go
```diff
@@ -370,11 +373,12 @@ func (r *EtcdClusterReconciler) refreshClusterState(ctx context.Context, s *reco
 // to the next; only an item that actually takes a mutating action ends the
 // loop.
 //
-// Steps 4-8 are TODO no-ops in this milestone (M2) — the mechanics they'd
+// Steps 4-7 are TODO no-ops in this milestone (M2) — the mechanics they'd
 // trigger (join/promote/repair/leave, CORRUPT/NOSPACE remediation,
 // lost-quorum recovery) land in M3-M5 — but the detection that decides
 // *whether* a step claims the loop is real, so the priority order itself is
 // reviewable now, ahead of any of that behavior landing.
+// Item 8 is implemented: it reconciles one EtcdMember at a time, until it is ready.
```
ahrtr: This PR only partially implements item 8.

In item 8, eventually we will need to support the following cases:
- provisioning etcd member
- replacing etcd member

internal/controller/etcdcluster_controller.go
```diff
@@ -440,12 +444,12 @@ func (r *EtcdClusterReconciler) dispatch(ctx context.Context, s *reconcileState)
 
 	// 8. Advance whatever's left not-Ready (Pending/Provisioning/Replacing).
 	// An existing learner always wins this slot (requirement 11).
-	if notReady := pickNotReadyMember(s.members); notReady != nil {
-		// TODO: §4.9 item 8 — promote if a caught-up learner,
-		// otherwise continue the member's join/replace progress (§4.6, M3).
-		logger.Info("Member not Ready; join/promotion not implemented yet",
-			"member", notReady.Name, "phase", notReady.Status.Phase)
-		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	notReady, err := pickNotReadyMember(s)
```
ahrtr: If there are multiple members not ready, we should pick up the learner if exist; otherwise pick up the member with the lowest ordinal.

internal/controller/etcdcluster_controller.go
```diff
@@ -537,15 +541,21 @@ func (r *EtcdClusterReconciler) updateConfig(ctx context.Context, s *reconcileSt
...
 	logger := log.FromContext(ctx)
+	// §4.2 gate: membership may only change while every existing member is
+	// Ready. pickNotReadyMember in dispatch only covers phases the
+	// provisioning ladder owns; repair and leave phases (Recreating,
+	// Terminating) are deliberately outside that ladder, so this check
+	// is the contract that keeps scale-out/in from mutating membership
+	// mid-repair once those phases land.
+	if !allReady(s.members) {
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
```
ahrtr: If there are any not-ready member, them the reconciliation loop won't reach here at all. Please see the r econcile_loop_v0.3.0 diagram
```suggestion
```

internal/controller/etcdcluster_controller.go
```diff
@@ -777,15 +775,52 @@ func isCertManagerCRDPresent(mgr ctrl.Manager) bool {
...
+	// Both Pod watches are required: Owns() covers legacy Pods that older
+	// operator versions owned directly by the EtcdCluster, while the Watches
+	// map covers new Pods whose controller owner is the EtcdMember. Owns() is
+	// not transitive, so member-owned Pods need the explicit map.
+	// TODO: remove Owns(&corev1.Pod{}) in the future.
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&ecv1alpha1.EtcdCluster{}).
 		Owns(&ecv1alpha1.EtcdMember{}).
 		Owns(&corev1.Pod{}).
+		Watches(&corev1.Pod{}, handler.EnqueueRequestsFromMapFunc(mapMemberPodToCluster)).
```
ahrtr: Please move this to separate PR.

Let's focus this PR to resolve the scale out case (and provisioning) only.

internal/controller/utils.go
```diff
@@ -203,6 +205,52 @@ func peerEndpointForOrdinalIndex(ec *ecv1alpha1.EtcdCluster, index int) (string,
...
+}
+
+// parsePeerEndpoint is the inverse of peerEndpointForOrdinalIndex: it parses a
+// peer URL of the shape
+// {scheme}://{cluster}-{ordinal}.{cluster}.{namespace}.svc.{domain}:2380
+// back into its identity triple. The cluster label of the
+// hostname is cross-checked against the parsed pod label so a foreign peer URL
+// can never be mistaken for one of this cluster's deterministic endpoints. It
+// returns ok=false for anything that is not a deterministic peer endpoint.
+func parsePeerEndpoint(peerURL string) (clusterName string, ordinal int, namespace string, ok bool) {
```
ahrtr: IT seems that this function is only used by test? If yes, I'd suggest to remove it

internal/controller/etcdmember_test.go
```diff
@@ -15,13 +15,1053 @@
...
+		*ecv1alpha1.EtcdMember,
+		func(*ecv1alpha1.EtcdMemberStatus),
+	) error
+}
+
+type countingStatusWriter struct {
+	client.SubResourceWriter
+	updateCalls       int
+	remainingConflict int
+}
```
ahrtr: This test file is over complicated. 

For a unit test file, I expect that we only test specific functions or methods in etcdmember.go.

For functionalities, we should verify them using e2e test cases.

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,90 @@ func TestClusterHealthy(t *testing.T) {
...
 }
 
+func TestNormalMemberProvisioning(t *testing.T) {
+	testCases := []struct {
+		name        string
+		initialSize int
+		desiredSize int
+	}{
+		{name: "CreateSize3", initialSize: 3, desiredSize: 3},
+		{name: "ScaleOutFrom1To3", initialSize: 1, desiredSize: 3},
```
ahrtr, line 138: Pls add one more case below,

```
{name: "ScaleOutFrom3To5", initialSize: 3, desiredSize: 5},
```

test/e2e/helpers_test.go
```diff
@@ -137,6 +137,60 @@ func waitForPodReadiness(t *testing.T, c *envconf.Config, name string, expectedR
 	}, wait.WithTimeout(5*time.Minute), wait.WithInterval(10*time.Second))
 }
 
+// waitForOrderedEtcdMemberReadiness waits for every expected EtcdMember to
+// become Ready and fails immediately if a higher ordinal appears before every
+// lower ordinal is durably Ready.
+func waitForOrderedEtcdMemberReadiness(
```
ahrtr: Please consider consolidate this helper function with the existing one `waitForPodReadiness`. We can have an unified helper function below,

```
func waitForAllEtcdMemberReady(t *testing.T, c *envconf.Config, ec *ecv1alpha1.EtcdCluster)
```

In this function, we should check both Pod status and all etcd member's status using etcd client SDK

internal/controller/etcdmember_test.go
ahrtr: > For a unit test file, I expect that we only test specific functions or methods

If there isn't such method or function to test, we can just remove all the changes in etcdmember_test.go to simplify this PR.

internal/controller/etcdcluster_controller.go
silentred: Updated the comment.

internal/controller/etcdcluster_controller.go
silentred: Updated.

internal/controller/etcdcluster_controller.go
silentred: Removed

internal/controller/etcdcluster_controller.go
silentred: Removed

internal/controller/utils.go
silentred: No. it is used by `findLiveMember` and `deterministicMemberNameForPeers`.

internal/controller/etcdmember_test.go
silentred: Removed all changes in etcdmember_test.go

test/e2e/e2e_test.go
silentred: Updated

test/e2e/helpers_test.go
silentred: Updated.

internal/controller/etcdcluster_controller.go
silentred: https://github.com/etcd-io/etcd-operator/blob/7b8968b33ed7356becd331a7d56e725156c68a69/internal/controller/pods.go#L75-L77

FYI, I did not change this logic in `listOwnedPods` either.

silentred:
I will squash the commits when it is ready to merge.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -324,7 +324,7 @@ func TestScaleClusterBootstrap(t *testing.T) {
 		},
 	}
 
-	t.Run("No members — creates EtcdMember ordinal 0 and its Pod", func(t *testing.T) {
+	t.Run("No members creates exactly one Pending member without a Pod", func(t *testing.T) {
```
ahrtr:
```suggestion
	t.Run("`scaleCluster` creates exactly one Pending member without a Pod", func(t *testing.T) {
```

internal/controller/etcdcluster_controller_test.go
```diff
@@ -552,6 +560,63 @@ func TestDispatch(t *testing.T) {
...
 }
 
+func TestGapAwareScaleOutCreatesLowestMissingOrdinal(t *testing.T) {
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+	require.NoError(t, ecv1alpha1.AddToScheme(scheme))
+
+	cluster := &ecv1alpha1.EtcdCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: "etcd", Namespace: "default", UID: "cluster-uid"},
+		Spec:       ecv1alpha1.EtcdClusterSpec{Size: 3, Version: "3.5.17"},
```
ahrtr, line 600: let's use the latest etcd version 3.7.1 as the example

curious why set a UID here?

internal/controller/etcdcluster_controller_test.go
```diff
@@ -552,6 +560,63 @@ func TestDispatch(t *testing.T) {
...
+	}
+	readyMember := func(ordinal int) ecv1alpha1.EtcdMember {
+		return ecv1alpha1.EtcdMember{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      etcdMemberName(cluster.Name, ordinal),
+				Namespace: cluster.Namespace,
+				UID:       types.UID(fmt.Sprintf("member-%d-uid", ordinal)),
```
ahrtr: why set a UID?

internal/controller/pods.go
```diff
@@ -71,10 +72,19 @@ func listOwnedPods(ctx context.Context, c client.Client, ec *ecv1alpha1.EtcdClus
 		return nil, fmt.Errorf("failed to list pods for cluster %s: %w", ec.Name, err)
 	}
 
-	var owned []*corev1.Pod
+	owned := []*corev1.Pod{}
 	for i := range podList.Items {
 		if metav1.IsControlledBy(&podList.Items[i], ec) {
 			owned = append(owned, &podList.Items[i])
```
ahrtr: We need to remove this, the etcd PODs should only be owned by EtcdMember.

internal/controller/pods.go
```diff
@@ -84,6 +94,32 @@ func listOwnedPods(ctx context.Context, c client.Client, ec *ecv1alpha1.EtcdClus
...
+	if err := c.Get(ctx, key, member); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("getting controlling EtcdMember %q for Pod %q: %w", key.String(), pod.Name, err)
+	}
+
+	if owner.UID != member.UID || member.Spec.ClusterName != ec.Name {
+		return false, nil
+	}
```
ahrtr: If this happens, then it's a bug.

ahrtr (commented):
still in progress of review...

internal/controller/pods.go
silentred: For example, manually created Pod and EtcdMember with incorrect field values may hit this condition. Should we add a Error log here or any other recommendations?

internal/controller/etcdcluster_controller_test.go
silentred: > let's use the latest etcd version 3.7.1 as the example

There are a lot of "3.5.17" literal in tests. Maybe we could update them in a follow-up PR.

> why set a UID 

```
require.Len(t, target.OwnerReferences, 1)
require.Equal(t, clusterUID, target.OwnerReferences[0].UID)
```
I added an assert to verify the newly created EtcdMember's owner UID euqals EtcdCluster's UID.

internal/controller/etcdcluster_controller_test.go
silentred: It is not used for now. Should we remove it?

internal/controller/pods.go
silentred: OK. Removal affects some tests, I will resolve them.

silentred: Should we also remove `Owns(&corev1.Pod{})` in `SetupWithManager`?

internal/controller/etcdcluster_controller_test.go
ahrtr: yes, please.  Let's make it as simple as possible, to avoid any potential confusion.

internal/controller/pods.go
ahrtr: I see that you already removed it, it's OK.

We already monitor all PVC and service, it's weird. We should only monitor the PVC and service created by etcd-operator. We can discuss & revisit this separately

```
Owns(&corev1.PersistentVolumeClaim{}).
Owns(&corev1.Service{})
```

internal/controller/pods.go
ahrtr: Users are only supposed to operate EtcdCluster. Let's remove it for now. We can consider to add verification in test, we can revisit this later separately

Example,
https://github.com/etcd-io/etcd/blob/main/client/pkg/verify/verify.go

internal/controller/etcdcluster_controller_test.go
ahrtr: > Maybe we could update them in a follow-up PR.

Sure.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -34,16 +35,34 @@ import (
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      clusterName + "-0",
+				Namespace: namespace,
+				UID:       types.UID("member-" + clusterName + "-0"),
+				OwnerReferences: []metav1.OwnerReference{{
+					APIVersion: ecv1alpha1.GroupVersion.String(),
+					Kind:       "EtcdCluster",
+					Name:       clusterName,
+					UID:        types.UID(clusterUID),
+					Controller: pointerToBool(true),
```
ahrtr: Note starting from go 1.26, the build-in `new` allows its operand to be an expression. https://go.dev/doc/go1.26#language

This is a generic comment, please update all such usage in this PR.
```suggestion
					Controller: new(true),
```

internal/controller/pods.go
silentred: Removed.

internal/controller/etcdcluster_controller_test.go
silentred: Removed.

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,504 @@ import (
...
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
+
+	// 2. Bootstrap: ordinal 0 with no live membership is a brand-new cluster.
+	// Create its first Pod with a self-contained one-member
+	// ETCD_INITIAL_CLUSTER; this is the only path that starts etcd with
+	// cluster-state=new — every later member joins an existing cluster.
+	if member.Spec.Ordinal == 0 && (state.memberListResp == nil || len(state.memberListResp.Members) == 0) {
+		name, peerURL := peerEndpointForOrdinalIndex(state.cluster, member.Spec.Ordinal)
+		err := r.ensureProvisioningPod(
```
ahrtr: We need to ensure the provisioned Secret(certificate) and PVC are owned by the `EtcdMember` instead of the `EtcdCluster`. The existing implementation is owned by the EtcdCluster. Please make it correct in this PR.

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,504 @@ import (
...
+	default:
+		return ctrl.Result{}, nil
+	}
+}
+
+// reconcileProvisioning establishes the durable write-before-mutate phase
+// boundary and implements normal provisioning only. Ordinal 0 with no live
+// membership is the bootstrap voter branch; every other unregistered peer is
+// added as a learner.
+func (r *EtcdClusterReconciler) reconcileProvisioning(
```
ahrtr, line 71: Please refactor this method per seudo code below,

```

	/*
		if phase == "" || phase == Pending {
			change phase to provisioning and requeue
		}

		// The implementation should be idempotent
		provisionCertificates(ctx, state, member)

		// The implementation should be idempotent
		provisionPVC(ctx, state, member)

		if bootstrap case {
			initialState = "new"
		} else {
			initialState = "existing"
			m := findMemberByPeerURL(...)
			if m == nil {
				add a new learner into the etcd cluster
			}
		}

		if the POD not exist {
			start the POD with `--initial-cluster-state ${initialState}`
			requeue to give the POD enough time to get started
		}

		if the member not healthy {
			// We should have a common function/method to take care of this. We can resolve it in a separate PR after https://github.com/etcd-io/etcd-operator/issues/472 is supported.
			replace if RecreateCount has already reached 3, otherwise recreate the POD and increment the RecreateCount 
			requeue
		}
	
		if the member is a learner {
			promote the learner
		}
	
		change phase to ready
	*/
```

internal/controller/etcdmember.go
silentred: Agree with PVC be owned by `EtcdMember`.
In contrast, currently each `EtcdCluster` has 3 `Secrets`(cluster-client-tls, cluster-server-tls, cluster-peer-tls), which is quite intuitive. It does not make sense to me to generate secrets for each `EtcdMember`.

ahrtr: > In contrast, currently each `EtcdCluster` has 3 `Secrets`(cluster-client-tls, cluster-server-tls, cluster-peer-tls), which is quite intuitive. It does not make sense to me to generate secrets for each `EtcdMember`.

It's the current implementation, the secrets/certificates are cluster-wide. Ideally, we should have per-member secrets/certificates. But it adds additional complexity. So let's keep it as it's for now. 

For PVC, it's owned by EtcdMember. For Secrets/certificates, they are still owned by EtcdCluster for now.

It means that we need to update our member terminating and cluster finalizing logic.

silentred: > Ideally, we should have per-member secrets/certificates.

See your point. Currently we have certs with `*.cluster.ns.svc` SAN for all members. This may not be the best practice. We could discuss it in future.

nwnt:
.. umm would it be possible to request for breaking this PR down into smaller ones?  I'm having a bit of a hard time trying to follow what's going on, seems like too many things are happening at once in this PR.

ahrtr:
> .. umm would it be possible to request for breaking this PR down into smaller ones?

I would suggest keep updating this PR until we feel comfortable & confident on the overall change/logic. Sometimes breaking down without a clear big picture is even harder. Just adding more commits when addressing review comments as you've been doing, and only squash before merging.

> I'm having a bit of a hard time trying to follow what's going on, seems like too many things are happening at once in this PR.

The etcdmember.go, especially the `reconcileProvisioning ` function, is the key implementation. Once we get it sorted out, I believe it will be clearer. Please focusing on improve this function first.

internal/controller/etcdcluster_controller.go
```diff
@@ -537,11 +537,8 @@ func (r *EtcdClusterReconciler) updateConfig(ctx context.Context, s *reconcileSt
...
-// Creating the EtcdMember object — and, for ordinal 0, its cert/PVC/Pod
-// directly (the bootstrap special case, §4.6 step 2) — is the only mechanic
-// this milestone (M2) implements. For every other ordinal, the general join
-// mechanics (MemberAdd + cert/PVC/Pod) are §4.6's TODO (M3), so a
-// newly-created EtcdMember simply sits at Phase: Pending until then. Scale-in
+// Scale-out only creates one Pending EtcdMember. The shared member lifecycle
+// owns bootstrap, member addition, Pod creation, and learner promotion. Scale-in
 // deletes the EtcdMember directly; the finalizer (§4.3) blocks its actual
 // removal until the Terminating leave sequence (§4.6, M3) runs.
 func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
```
nwnt, line 544: Should this now be gated as proposed by[ point 4.2](https://github.com/etcd-io/etcd-operator/blob/main/docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md#42-readiness-gate-only-orchestrate-when-everything-is-ready) in the proposed designed?   My understanding is we need to perform it at the top of this function too.

nwnt: Meant to highlight the function `scaleCluster`...

silentred: https://github.com/etcd-io/etcd-operator/pull/478#discussion_r3852969440
Benjamin suggested removing the gate.

internal/controller/etcdmember.go
silentred: Updated.

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+	} else if _, registered := findLiveMember(state.cluster, member, state.memberListResp); !registered {
+		if hasLiveLearner(state.memberListResp) {
+			return ctrl.Result{RequeueAfter: requeueDuration}, nil
+		}
+		if err := r.addLearner(state, member); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Requeue so the Pod is created in a later pass that renders
+		// ETCD_INITIAL_CLUSTER from a fresh membership snapshot.
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
```
ahrtr: We don't need to requeue here. 
```suggestion
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+	// idempotent here because registration is checked against live state.
+	bootstrap := member.Spec.Ordinal == 0 &&
+		(state.memberListResp == nil || len(state.memberListResp.Members) == 0)
+	initialState := etcdClusterStateExisting
+	if bootstrap {
+		initialState = etcdClusterStateNew
+	} else if _, registered := findLiveMember(state.cluster, member, state.memberListResp); !registered {
+		if hasLiveLearner(state.memberListResp) {
+			return ctrl.Result{RequeueAfter: requeueDuration}, nil
+		}
```
ahrtr: If there is a learner, then it must be the etcd member that we are provisioning right now.
```suggestion
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+		if err := r.addLearner(state, member); err != nil {
+			return ctrl.Result{}, err
+		}
```
ahrtr: We need to check whether or not the member (learner) has already been added first (that's the reason why we need a `findMemberByPeerURL` or `findEtcdNodeForEtcdMember`). If not, add it as a learner; otherwise, do nothing here. 

- For the implementation of `findMemberByPeerURL`: we generate the peerURL from the EtcdMember.Spec.Ordinal, and then use the generated peerURL as a key to lookup the `state.memberListResp`

- [**preferred**] For the implementation of `findEtcdNodeForEtcdMember` (we use EtcdNode for the real etcd member from etcd perspective to avoid naming conflict, while `EtcdMember` for the CR): the implementation is identical to `findMemberByPeerURL`, but we get the details included in the function.

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+	}
+
+	// 3. Idempotent prerequisite: the member's own PVC for ReadWriteOnce
+	// storage.
+	if err := r.provisionPVC(ctx, state, member); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// 4. Initial state and membership registration. Bootstrap — ordinal 0
+	// with no live membership — starts a brand-new cluster; this is the only
```
```suggestion
	// with no live membership — starts the very first voter for the brand-new cluster; this is the only
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+			ctx,
+			state,
+			member,
+			initialState,
+			initialCluster,
+		); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
```
```suggestion
	if pod := findPodForEtcdMember(state, member); pod == nil {
		if err := r.createPodForEtcdMember(ctx, state, member); err != nil {
			return ctrl.Result{}, err
		}
		return ctrl.Result{RequeueAfter: requeueDuration}, nil
	}
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+			}
+		}
+	}
+
+	return nil, nil, fmt.Errorf(
+		"live health for EtcdMember %q with member ID %x is not mapped",
+		member.Name,
+		liveMember.ID,
+	)
+}
```
```suggestion
func findHealthStatusForEtcdMember(state *reconcileState, member *ecv1alpha1.EtcdMember) (*etcdutils.EpHealth, error) {
	memberName := memberUniqueName(member)
	if state.health != nil {
		if memberHealthStatus, ok := state.health.Members[memberName]; ok {
			return &memberHealthStatus, nil
		}
		return nil, fmt.Errorf("health status is not found for member %s", memberName)
	}

	return nil, fmt.Errorf("cannot find the health status for member %s as the health report is empty", memberName)
}
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +44,536 @@ import (
...
+		// TODO: Keep this direct etcdutils call scoped to normal scale-out. If
+		// remove/replacement expands the membership mutation surface, introduce
+		// a controller-owned seam in a separate PR so the boundary stays clear
+		// and its fakes remain easy to maintain.
+		if err := etcdutils.PromoteLearner(
+			etcdutils.ClientConfig{Endpoints: endpoints, TLS: state.tlsConfig},
+			liveMember.ID,
+		); err != nil {
+			return ctrl.Result{}, fmt.Errorf("promoting learner for EtcdMember %q: %w", member.Name, err)
+		}
```
ahrtr: Let's get all the details included in one method/function. 

We don't need to check the whether the learner is ready or not. Just try to promote it and requeue if failed.

internal/controller/etcdmember.go
ahrtr: It's better now. Please ping me back when you resolved all the new review comments.

internal/controller/etcdmember.go
silentred:
```
if etcdNode := findEtcdNodeForEtcdMember(state, member); !bootstrap && etcdNode == nil {
	if err := r.addLearner(state, member); err != nil {
		return ctrl.Result{}, err
	}
	return ctrl.Result{RequeueAfter: requeueDuration}, nil
}
```
I am not sure if `return ctrl.Result{RequeueAfter: requeueDuration}, nil` should be removed. The `addLearner` updated the membership of etcd cluster, so the `reconcileState` is out-of-date and should be refreshed by requeuing the request.

All other issues are resolved. Please take a look. Thanks.

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,491 @@ import (
...
+		(state.memberListResp == nil || len(state.memberListResp.Members) == 0)
+
+	if etcdNode := findEtcdNodeForEtcdMember(state, member); !bootstrap && etcdNode == nil {
+		if err := r.addLearner(state, member); err != nil {
+			return ctrl.Result{}, err
+		}
+		// addLearner updated the membership of etcd cluster, so
+		// we need to refresh the reconcileState by requeuing the request.
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
```
```suggestion
	if !bootstrap {
		if etcdNode := findEtcdNodeForEtcdMember(state, member); etcdNode == nil {
			if err := r.addLearner(state, member); err != nil {
				return ctrl.Result{}, err
			}
			// addLearner updated the membership of etcd cluster, so
			// we need to refresh the reconcileState by requeuing the request.
			return ctrl.Result{RequeueAfter: requeueDuration}, nil
		}
	}	
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,491 @@ import (
...
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
+
+	// 6. Convergence: wait until this member's endpoint reports healthy.
+	etcdNode := findEtcdNodeForEtcdMember(state, member)
+	if etcdNode == nil {
+		return ctrl.Result{}, fmt.Errorf("live etcd node for EtcdMember %q is not mapped", member.Name)
+	}
```
```suggestion
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,491 @@ import (
...
+	if !health.Health {
+		// TODO(#472): once a common recovery ladder exists, replace the
+		// member once RecreateCount has reached its limit; otherwise
+		// recreate the Pod and increment RecreateCount.
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
+
+	// 7. Promotion: if the member is still a learner, wait until the leader
+	// considers it caught up, then promote it to a voting member.
+	if etcdNode.IsLearner {
```
ahrtr, line 153:
```suggestion
	etcdNode := findEtcdNodeForEtcdMember(state, member)
	if etcdNode == nil {
		return ctrl.Result{}, fmt.Errorf("live etcd node for EtcdMember %q is not found", member.Name)
	}
	if etcdNode.IsLearner {
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,491 @@ import (
...
+		}
+	}
+	// 8. Completion: healthy voting member — mark it Ready.
+	if err := r.updateEtcdMemberStatus(ctx, member, func(status *ecv1alpha1.EtcdMemberStatus) {
+		status.Phase = ecv1alpha1.EtcdMemberReady
+		status.RecreateCount = 0
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{RequeueAfter: requeueDuration}, nil
```
ahrtr, line 167:
```suggestion

	// 8. Completion: healthy voting member — mark it Ready.
	if err := r.updateEtcdMemberStatus(ctx, member, func(status *ecv1alpha1.EtcdMemberStatus) {
		status.Phase = ecv1alpha1.EtcdMemberReady
		status.RecreateCount = 0
	}); err != nil {
		return ctrl.Result{}, err
	}
	
	return ctrl.Result{RequeueAfter: requeueDuration}, nil
```

internal/controller/etcdmember.go
ahrtr: It much better now. Please resolve the new review comments. 



> The `addLearner` updated the membership of etcd cluster, so the `reconcileState` is out-of-date and should be refreshed by requeuing the request.

It's a good point. Let's keep it for simplicity, otherwise we will have to update the state dynamically in the progress.

internal/controller/etcdmember.go
ahrtr: Note: we just move this code right above promoting the learner.

internal/controller/etcdmember.go
ahrtr: this comment just adds two blank lines.

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+	}
+}
+
+// reconcileProvisioning establishes the durable write-before-mutate phase
+// boundary and drives normal provisioning as a linear sequence of idempotent
+// steps: prerequisites (certificates, PVC), membership registration, Pod
+// creation, and health convergence. Ordinal 0 with no live membership is the
+// bootstrap voter branch; every other unregistered peer is added as a
+// learner.
+func (r *EtcdClusterReconciler) reconcileProvisioning(
```
ahrtr, line 71: The implementation of this function is in good shape now! The functions it calls may need some improvements, I will raise comments separately.

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+	// of restarting it.
+	enteringProvisioning := member.Status.Phase == "" || member.Status.Phase == ecv1alpha1.EtcdMemberPending
+	if err := r.updateEtcdMemberStatus(ctx, member, func(status *ecv1alpha1.EtcdMemberStatus) {
+		status.Phase = ecv1alpha1.EtcdMemberProvisioning
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
+	if enteringProvisioning {
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
```
```suggestion
	if enteringProvisioning {
		if err := r.updateEtcdMemberStatus(ctx, member, func(status *ecv1alpha1.EtcdMemberStatus) {
			status.Phase = ecv1alpha1.EtcdMemberProvisioning
		}); err != nil {
			return ctrl.Result{}, err
		}
		return ctrl.Result{RequeueAfter: requeueDuration}, nil
	}	
```

silentred:
@ahrtr All resolved.

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+	return nil
+}
+
+// provisionPVC ensures the member's own PVC exists. A per-member PVC is only
+// needed for ReadWriteOnce storage; clusters without a storage spec or with
+// ReadWriteMany storage skip it. Idempotent.
+func (r *EtcdClusterReconciler) provisionPVC(ctx context.Context, state *reconcileState, member *ecv1alpha1.EtcdMember) error {
+	cluster := state.cluster
+	if cluster.Spec.StorageSpec == nil || cluster.Spec.StorageSpec.AccessModes == corev1.ReadWriteMany {
```
ahrtr, line 187:
```suggestion
	// ReadWriteMany is assumed to be statically provisioned (e.g. a shared
	// NFS-backed PVC) and managed outside the operator, so no per-member PVC
	// is created for it.
	if cluster.Spec.StorageSpec == nil || cluster.Spec.StorageSpec.AccessModes == corev1.ReadWriteMany {
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+			}
+			if clusterName == state.cluster.Name && namespace == state.cluster.Namespace && ordinal == member.Spec.Ordinal {
+				etcdNode = liveMember
+				return
+			}
+		}
+	}
+
+	return
+}
```
```suggestion
func findEtcdNodeForEtcdMember(state *reconcileState, member *ecv1alpha1.EtcdMember) *etcdserverpb.Member {
	if state.memberListResp == nil {
		return nil
	}

	for _, node := range state.memberListResp.Members {
		for _, peerURL := range node.PeerURLs {
			memberName := node.Name
			if memberName == "" {
				memberName = getMemberNameFromPeerURL(peerURL)
			}
			if memberName == member.Name {
				return node
			}
		}
	}

	return nil
}
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+		true,
+	)
+	if err != nil {
+		return fmt.Errorf("adding learner for EtcdMember %q: %w", member.Name, err)
+	}
+	if response == nil {
+		return fmt.Errorf("adding learner for EtcdMember %q returned an empty response", member.Name)
+	}
+	return nil
+}
```
ahrtr, line 212:
```suggestion
func (r *EtcdClusterReconciler) addLearner(state *reconcileState, member *ecv1alpha1.EtcdMember) error {
	endpoints := clientEndpointsFromPods(state.cluster.Name, state.cluster.Namespace, state.pods, clusterTLSEnabled(state.cluster))
	if len(endpoints) == 0 {
		return fmt.Errorf("cannot add learner for EtcdMember %q: no live client endpoints", member.Name)
	}

	_, peerURL := peerEndpointForOrdinalIndex(state.cluster, member.Spec.Ordinal)
	if _, err := etcdutils.AddMember(etcdutils.ClientConfig{Endpoints: endpoints, TLS: state.tlsConfig}, []string{peerURL}, true); err != nil {
		return fmt.Errorf("failed to add learner for EtcdMember %q: %w", member.Name, err)
	}

	return nil
}
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+// absent at snapshot time.
+func findPodForEtcdMember(state *reconcileState, member *ecv1alpha1.EtcdMember) (pod *corev1.Pod) {
+	podName := memberPodName(state.cluster.Name, member.Spec.Ordinal)
+	for _, p := range state.pods {
+		if p.Name == podName {
+			pod = p
+			return
+		}
+	}
+	return
```
```suggestion
func findPodForEtcdMember(state *reconcileState, member *ecv1alpha1.EtcdMember) *corev1.Pod {
	podName := memberPodName(state.cluster.Name, member.Spec.Ordinal)
	for _, p := range state.pods {
		if p.Name == podName {
			return p
		}
	}
	return nil
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +42,495 @@ import (
...
+	return nil, fmt.Errorf("cannot find the health status for member %s as the health report is empty", memberName)
+}
+
+// renderInitialCluster renders every peer in an authoritative MemberAdd or
+// MemberList response. It never infers topology from an ordinal range.
+func renderInitialCluster(
+	cluster *ecv1alpha1.EtcdCluster,
+	owned []ecv1alpha1.EtcdMember,
+	authoritative []*etcdserverpb.Member,
+) (string, error) {
```
ahrtr: We expect that the EtcdMember should be always consistent with the etcd nodes, otherwise it's a critical bugs. We can add a verification in a followup PR.
```suggestion
func renderInitialCluster(cluster *ecv1alpha1.EtcdCluster, members []ecv1alpha1.EtcdMember) string {
```

ahrtr:
Please also try to cleanup unused functions/methods/code (that are added in this PR) if any.

internal/controller/etcdmember.go
```diff
@@ -33,17 +41,430 @@ import (
...
+		member := &state.members[i]
+		if !isProvisioningPhase(member) {
+			continue
+		}
+		if selected == nil || member.Spec.Ordinal < selected.Spec.Ordinal {
+			selected = member
+		}
+	}
+	return selected
+}
```
ahrtr, line 398: This should be a generic function, it returns an EtcdMember which needs provisioning or replace.
Pls remove `isProvisioningPhase` as well.
```suggestion
// pickNotReadyMember selects from the current live snapshot: a live learner
// mapped to its EtcdMember first, otherwise the lowest-ordinal member that
// isn't Ready yet. state.members is assumed sorted by ordinal, so the first
// match in the second pass is the lowest-ordinal one.
func pickNotReadyMember(
	state *reconcileState,
) *ecv1alpha1.EtcdMember {
	for i := range state.members {
		member := &state.members[i]
		if member.Status.Phase == ecv1alpha1.EtcdMemberReady {
			continue
		}
		if etcdNode := findEtcdNodeForEtcdMember(state, member); etcdNode != nil && etcdNode.IsLearner {
			return member
		}
	}

	for i := range state.members {
		member := &state.members[i]
		if member.Status.Phase != ecv1alpha1.EtcdMemberReady {
			return member
		}
	}
	return nil
}
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +41,430 @@ import (
...
+	switch member.Status.Phase {
+	// Creating the EtcdMember object and writing its status Phase are two
+	// separate calls in createEtcdMember. If the operator crashes in between,
+	// the object survives with an empty Phase, which must resume provisioning.
+	case "", ecv1alpha1.EtcdMemberPending, ecv1alpha1.EtcdMemberProvisioning, ecv1alpha1.EtcdMemberReplacing:
+		return true
+	default:
+		return false
+	}
+}
```
```suggestion
```

internal/controller/etcdmember.go
```diff
@@ -33,17 +41,430 @@ import (
...
+	return nil
+}
+
+// findEtcdNodeForEtcdMember correlates an EtcdMember resource with the
+// current etcd membership by name: a live node matches when its etcd name —
+// or, while a newly added node has not started and is still unnamed, the name
+// encoded in its peer URL — equals the EtcdMember's name.
+func findEtcdNodeForEtcdMember(state *reconcileState, member *ecv1alpha1.EtcdMember) *etcdserverpb.Member {
+	if state.memberListResp == nil {
+		return nil
```
silentred, line 357: Found another problem here:

https://github.com/etcd-io/etcd-operator/blob/7b8968b33ed7356becd331a7d56e725156c68a69/internal/controller/etcdcluster_controller.go#L343-L346

`state.memberListResp` is last updated in the function `refreshClusterState`. if only MemberList returns error, then `state.memberListResp` is nil, even the cluster is not in bootstraping. This may cause `reconcileProvisioning` trying to add duplicated member to etcd cluster.

Maybe we could resolve it in a follow-up PR. cc @ahrtr

internal/controller/pods.go
```diff
@@ -84,6 +89,29 @@ func listOwnedPods(ctx context.Context, c client.Client, ec *ecv1alpha1.EtcdClus
 	return owned, nil
 }
 
+func podIsControlledByClusterMember(
```
ahrtr: I re-implemented both `listOwnedPods` and `podIsControlledByClusterMember`, pls update this PR.

```
// listOwnedPods returns all Pods controlled by one of the given EtcdMembers,
// sorted by name. EtcdCluster never owns Pods directly; a Pod belongs to the
// cluster when its controlling EtcdMember is in members. Checking against
// this already-fetched slice avoids re-reading each Pod's controlling
// EtcdMember from the API server.
func listOwnedPods(ctx context.Context, c client.Client, ec *ecv1alpha1.EtcdCluster, members []ecv1alpha1.EtcdMember) ([]*corev1.Pod, error) {
	podList := &corev1.PodList{}
	if err := c.List(ctx, podList,
		client.InNamespace(ec.Namespace),
		client.MatchingLabels(etcdClusterLabels(ec)),
	); err != nil {
		return nil, fmt.Errorf("failed to list pods for cluster %s: %w", ec.Name, err)
	}

	owned := []*corev1.Pod{}
	for i := range podList.Items {
		pod := &podList.Items[i]
		if podIsControlledByClusterMember(pod, members) {
			owned = append(owned, pod)
		}
	}

	sort.Slice(owned, func(i, j int) bool {
		return owned[i].Name < owned[j].Name
	})
	return owned, nil
}

func podIsControlledByClusterMember(pod *corev1.Pod, members []ecv1alpha1.EtcdMember) bool {
	owner := metav1.GetControllerOf(pod)
	if owner == nil || owner.APIVersion != ecv1alpha1.GroupVersion.String() || owner.Kind != "EtcdMember" {
		return false
	}

	for i := range members {
		if members[i].Name == owner.Name {
			return true
		}
	}
	return false
}
```

internal/controller/utils.go
```diff
@@ -203,6 +219,64 @@ func peerEndpointForOrdinalIndex(ec *ecv1alpha1.EtcdCluster, index int) (string,
...
+// member name encoded in a peer URL of the shape produced by
+// peerEndpointForOrdinalIndex, or "" for anything that is not a
+// deterministic peer endpoint.
+func getMemberNameFromPeerURL(peerURL string) string {
+	clusterName, ordinal, _, ok := parsePeerEndpoint(peerURL)
+	if !ok {
+		return ""
+	}
+	return etcdMemberName(clusterName, ordinal)
+}
```
ahrtr: Please also remove `parsePeerEndpoint` and `parseMemberName`
```suggestion
// getMemberNameFromPeerURL returns the {cluster}-{ordinal} member name
// encoded in a peer URL of the shape produced by peerEndpointForOrdinalIndex
// — the hostname label before its first ".". Returns "" if the URL has no
// host.
func getMemberNameFromPeerURL(peerURL string) string {
	u, err := url.Parse(peerURL)
	if err != nil {
		return ""
	}
	host := u.Hostname()
	if idx := strings.Index(host, "."); idx != -1 {
		return host[:idx]
	}
	return host
}
```

ahrtr (commented):
It's close to the merge state now...

silentred:
ping @ahrtr

internal/controller/etcdmember.go
ahrtr: Let me take care of this separately, if the MemberList fails, the controller shouldn't run into this phase.

ahrtr:
One last question, have you manually tested this PR? 
- create a 1 or 3 or 5 member cluster
- scale out from 1 to 3
- scale out from 3 to 5

silentred:
> have you manually tested this PR?

No. e2e tests include:
- create 1-node, 3-node cluster
- scale from 1 to 3
- scale from 3 to 5

ahrtr (approved):
Thanks for the great work and collaboration.

ahrtr:
Merging this PR...to unblock others' work (avoid conflict)

@ArkaSaha30 @GunaKKIBM  @nwnt please feel free to raise any question even after this PR gets merged.

ahrtr:
@silentred please squash the commits

silentred:
> Thanks for the great work and collaboration.

Thanks for the help. Let me squash the commits.

internal/controller/etcdmember.go
ahrtr: see https://github.com/etcd-io/etcd-operator/pull/487

PR #479, opened 2026-08-24
Support both both serializable and linearizable health checks

### LINKED ISSUE #468: Healthcheck should explicitly perform both serializable and linearizable checks
For each member, we only perform the serializable check. We can send a serializable read request. The current related code is [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L166).

We also need to check the alarms (CORRUPTION or NOSPACE), the alarm request itself is a linearizable request. Also note the MemberList is also a linearizable request by default. So we can depend on both of them to perform the linearizable check.

The health check result should explicitly contain the following info,
- Each etcd member’s health status (serializability)
- Any alarm if any
- Cluster health status (linearizability)

ahrtr:
Fix https://github.com/etcd-io/etcd-operator/issues/468

@ArkaSaha30 @GunaKKIBM @ivanvc   @nwnt @silentred

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -69,7 +69,7 @@ type reconcileState struct {
 	members        []ecv1alpha1.EtcdMember      // EtcdMembers owned by this cluster, sorted by ordinal
 	pods           []*corev1.Pod                // member pods owned by this cluster, sorted by ordinal
 	memberListResp *clientv3.MemberListResponse // member list fetched from the etcd cluster
-	memberHealth   []etcdutils.EpHealth         // health information for each etcd member
+	health         *etcdutils.ClusterHealth     // cluster/member health and active alarms from the latest health check
 	tlsConfig      *tls.Config                  // etcd client TLS config used by every etcdutils call in this loop (nil for non-TLS clusters)
 }
 
@@ -168,9 +168,7 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	// 5. Always-on refresh: live member list/health (§4.2) — never gated.
-	if err = r.refreshClusterState(ctx, state); err != nil {
-		return ctrl.Result{}, err
-	}
+	r.refreshClusterState(ctx, state)
 
 	// 6. Pause (requirement 15): do nothing else at all this loop, ahead of
 	// dispatch (including an already-started lost-quorum recovery).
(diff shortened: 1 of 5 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -730,9 +754,9 @@ func (r *EtcdClusterReconciler) updateConditions(s *reconcileState) {
 		Message:            "All etcd members are healthy",
 	}
 
-	if s.memberListResp != nil && len(s.memberHealth) > 0 {
+	if s.memberListResp != nil && len(memberHealth) > 0 {
 		var unhealthyMembers []string
-		for _, health := range s.memberHealth {
+		for _, health := range memberHealth {
 			if !health.Health && health.Status != nil {
```
silentred: https://github.com/etcd-io/etcd-operator/blob/709fee39df73424a3fee631d988e8b1e960570c7/internal/etcdutils/etcdutils.go#L224-L240

If `eh.Health` is false, then `eh.Status` is always nil. So the unhealthyMembers cannot be successfully collected.
Maybe we should drop `health.Status != nil` condition.

internal/controller/pods.go
```diff
@@ -162,53 +162,67 @@ func clusterTLSEnabled(ec *ecv1alpha1.EtcdCluster) bool {
...
 	if err != nil {
-		return memberlistResp, nil, err
+		return memberlistResp, health, err
 	}
 
 	var memberErrors []error
-	for _, healthInfo := range healthInfos {
+	for _, healthInfo := range memberHealth {
 		if !healthInfo.Health {
 			memberErrors = append(memberErrors, errors.New(healthInfo.String()))
```
silentred, line 217: If any member is unhealthy, the aggregated error will be non-nil. This prevents the reconcile workflow from reaching `r.dispatch()`, meaning unhealthy pods won't be repaired.

internal/etcdutils/etcdutils.go
```diff
@@ -181,16 +229,13 @@ func ClusterHealth(cfg ClientConfig) ([]EpHealth, error) {
...
-					eh.Health = false
-					eh.Error = fmt.Sprintf("Unable to fetch the status :%s", err.Error())
-				} else {
+				// Best-effort metadata fetch: its outcome does not affect
+				// Health, which is determined solely by the range request
+				// above.
+				if epStatus, err := cli.Status(ctx, ep); err == nil {
 					eh.Status = epStatus
-					if len(epStatus.Errors) > 0 {
-						eh.Health = false
```
silentred, line 191: This is probably neccesary. When there is an active alarm in the cluster, `epStatus.Errors` will be non-empty. Without this check, `eh.Health` would incorrectly evaluate to true even if all members have active alarms.

internal/controller/etcdcluster_controller.go
ahrtr: good catch, updated

internal/controller/pods.go
ahrtr: This PR only collects the health check result, and I was planning to handle such case in followup PR.

Updated anyway.

internal/etcdutils/etcdutils.go
ahrtr: alarms are handled separately. For each member's health, we only depend on the serializable read request.

ahrtr: Pls also read the https://github.com/etcd-io/etcd-operator/blob/main/docs/design/reconcile_loop_v0.3.0.png, the alarms are handled first before fixing any unhealthy members

silentred: Got it. Thanks for the explaination.

internal/etcdutils/etcdutils.go
```diff
@@ -118,47 +166,63 @@ func FindLeaderStatus(healthInfos []EpHealth, logger logr.Logger) (uint64, *clie
...
+		cfg  *clientv3.Config
+		name string
+	}
+
+	var cfgs = make([]epConfig, 0, len(cfg.Endpoints))
+	for i, ep := range cfg.Endpoints {
 		epCfg := ClientConfig{Endpoints: []string{ep}, TLS: cfg.TLS}
 		built := epCfg.buildConfig()
-		cfgs = append(cfgs, &built)
+		cfgs = append(cfgs, epConfig{cfg: &built, name: cfg.Names[i]})
```
silentred, line 213: Do we need to validate that `cfg.Names` is non-nil (and has the same length as `cfg.Endpoints`) in MemberHealth?

https://github.com/etcd-io/etcd-operator/blob/981731bdc7fb4177a1a8ffd19812007d6c9d0058/internal/etcdutils/etcdutils.go#L31-L33
The comment on `ClientConfig.Names` says "optional, callers that don't need EpHealth.Name populated may leave it nil", but the this line indexes `cfg.Names[i]` unconditionally, so a caller following the documented contract would hit an index-out-of-range panic.

https://github.com/etcd-io/etcd-operator/blob/981731bdc7fb4177a1a8ffd19812007d6c9d0058/internal/controller/pods.go#L205
I know the current callers are fine — `healthCheck` builds Names from the same pods slice, so the lengths always match. My concern is future callers trusting the "optional" contract.

ahrtr: > Do we need to validate that `cfg.Names` is non-nil (and has the same length as `cfg.Endpoints`) in MemberHealth?

Each name will never be nil, as we get the value from POD name. We can consider how to refactor the implementation to make it more readable if needed in future.



> The comment on `ClientConfig.Names` says "optional,

It's because the `ClientConfig` is used in multiple places. When calling `MemberHealth`, each name in `ClientConfig.Names` will never be nil.

silentred: > When calling `MemberHealth`, each name in `ClientConfig.Names` will never be nil.

So it is caller's responsibility to guarantee `ClientConfig.Names` is not a nil string slice. Could we add this to comment of `MemberHealth`?

ahrtr: Added.

silentred (approved):
LGTM. Thanks.

ahrtr:

PR #485, opened 2026-08-25
Simplify e2e teardown for ephemeral KinD clusters

### LINKED ISSUE #122: Unneeded e2e tear down steps
testEnv.Finish has a number of steps which are not required and flood the terminal with extra information which make it harder to view test results.

Steps such as removing the namespaces, CRDS etc are not required if the kind cluster is being removed. These could be reasonable when testing against a persistent cluster instead of an ephemeral one such as Kind.

I'm happy to work on this issue

karangupta982:
Closes #122

Simplifies e2e teardown for ephemeral KinD clusters.

- Keep the EtcdCluster/EtcdMember cleanup required while the operator is running.
- Remove redundant teardown steps for the operator, CRDs, namespace, Prometheus, and cert-manager.
- Destroy the KinD cluster directly after the required etcd resource cleanup.
- Preserve `ETCD_E2E_SKIP_TEARDOWN` behavior.

Since the KinD cluster is ephemeral, destroying it already removes the resources installed in the cluster.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -176,72 +176,33 @@ func TestMain(m *testing.M) {
 
 	// Use the Environment.Finish method to define clean up steps
 	testEnv.Finish(
-		// cleanup environment
+		// Force-drain EtcdCluster/EtcdMember finalizers while the operator is
+		// still running.  This MUST happen before the cluster is destroyed;
+		// the operator needs to be alive to process the finalizers.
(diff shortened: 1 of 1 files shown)


karangupta982:
Hi @ArkaSaha30 , just following up on this PR. Please let me know if there is anything I should modify or address. 
Thanks!

karangupta982:
Hi @silentred , @nwnt , is there any checks or something inappropriate? Please let me know if i missed something.

PR #488, opened 2026-08-31
feat: add pod config drift detection

nwnt:
Fix #411. Implement `updateConfig` config drift detection part.  After detected, the controller updates the corresponding `EtcdMember`'s phase to Recreating.  The reconciliation of Recreating logic will be implemented in phase 3.

cc @ahrtr @silentred

PR #495, opened 2026-09-02
fix: hash only explicit pod-config fields in EtcdClusterHash

silentred:
`EtcdClusterHash` only hash fields in `clusterHashInput`. 
/cc @ahrtr @nwnt 
Link to https://github.com/etcd-io/etcd-operator/pull/488

### DIFF
--- a/internal/controller/etcdcluster.go
+++ b/internal/controller/etcdcluster.go
@@ -13,19 +13,56 @@ const (
 	hashLength = 12
 )
 
+// clusterHashInput is the explicit set of EtcdClusterSpec fields that
+// determine the etcd Pod's configuration. Only these fields participate in
+// EtcdClusterHash; a new spec field must be added here deliberately to
+// become part of the hash. Fields intentionally left out:
+//   - Size: scaling is handled separately from configuration.
+//   - Version: upgrades roll one member at a time elsewhere.
+//   - PodTemplate.Metadata: applied in place without a Pod roll.
+//   - Paused, AllowVersionUpgradeOnRepair: operator knobs, not Pod config.
+//
+// Because the input is this private struct rather than EtcdClusterSpec
+// itself, cosmetic API changes (renaming unrelated fields, adding new
+// operator knobs) cannot change the hash; only changes here or to a
+// whitelisted nested type do.
+type clusterHashInput struct {
+	ImageRegistry string
+	StorageSpec   *v1alpha1.StorageSpec
+	TLS           *v1alpha1.TLSCertificate
+	EtcdOptions   []string
+	PodSpec       *v1alpha1.PodSpec
+}
+
+// EtcdClusterHash returns a stable hash over the fields of the cluster spec
+// that determine the etcd Pod's configuration (see clusterHashInput).
+// Changing any other spec field must not change the hash, so toggling
+// operator knobs or scaling the cluster never marks Pods as drifted; in
+// particular a PodTemplate without a Spec — metadata only — hashes like no
+// PodTemplate at all.
+//
+// NOTE: changing this function or any whitelisted nested type (StorageSpec,
+// TLSCertificate, PodSpec) changes every hash output. After an operator
+// upgrade, every existing Pod's stored hash annotation would then look
+// drifted, rolling the whole cluster once. TestEtcdClusterHashPinned fails
+// on such a change; when it does, revisit this function before updating its
+// pinned expected value.
 func EtcdClusterHash(ec *v1alpha1.EtcdCluster) string {
-	clusterSpec := ec.DeepCopy().Spec
-	// size is not used for calculating the hash
-	clusterSpec.Size = 0
-	// version is handled separately
-	clusterSpec.Version = ""
-	clusterSpec.EtcdOptions = createArgs(ec.Name, ec.Spec.EtcdOptions, clusterTLSEnabled(ec))
-	// we don't want to roll etcd pods when metadata is changed.
-	// instead, we'd do in-place update for them.
-	if clusterSpec.PodTemplate != nil {
-		clusterSpec.PodTemplate.Metadata = nil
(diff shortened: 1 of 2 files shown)


internal/controller/etcdcluster.go
```diff
@@ -13,19 +13,40 @@ const (
...
-	// size is not used for calculating the hash
-	clusterSpec.Size = 0
-	// version is handled separately
-	clusterSpec.Version = ""
-	clusterSpec.EtcdOptions = createArgs(ec.Name, ec.Spec.EtcdOptions, clusterTLSEnabled(ec))
-	// we don't want to roll etcd pods when metadata is changed.
-	// instead, we'd do in-place update for them.
-	if clusterSpec.PodTemplate != nil {
-		clusterSpec.PodTemplate.Metadata = nil
+	input := clusterHashInput{
```
ahrtr, line 51: can we just reuse the existing `EtcdClusterSpec`?

silentred: I tested `dump.ForHash` with 2 version of empty `EtcdClusterSpec`:
1. current empty `EtcdClusterSpec`: 
- result of `dump.ForHash`: `(v1alpha1.EtcdClusterSpec){Size:(int)0 ImageRegistry:(string) Version:(string) StorageSpec:(*v1alpha1.StorageSpec)<nil> TLS:(*v1alpha1.TLSCertificate)<nil> EtcdOptions:([]string)<nil> PodTemplate:(*v1alpha1.PodTemplate)<nil> Paused:(bool)false AllowVersionUpgradeOnRepair:(bool)false}`
- hash: `8db165db98ac`
2. empty `EtcdClusterSpec` with an extra field `Foo string`:
- result of `dump.ForHash`: `... AllowVersionUpgradeOnRepair:(bool)false Foo:(string)}`
- hash: `0ac2c71d4299`

Maybe we should think about this question: how do we prevent all etcd pods from rolling update, when the `clusterHashInput` or `EtcdClusterSpec` added a new field(CRD version update)?

ahrtr: - Let's just reuse `EtcdClusterSpec` in this function.

- Feel free to add a unit test case to detect any future scheme change on `EtcdClusterSpec`, and a comment to reminder maintainer to revisit `EtcdClusterHash` if the test case fails

silentred: Updated.

internal/controller/etcdcluster_test.go
```diff
@@ -132,3 +220,29 @@ func TestEtcdClusterHash(t *testing.T) {
...
+	// at commit 5483df641bf23347d98a56b8d966bc2c495edcb5.
+	const expectedEmptySpecHash = "19f6087f4f17"
+
+	currentHash := EtcdClusterHash(&v1alpha1.EtcdCluster{Spec: v1alpha1.EtcdClusterSpec{}})
+	if currentHash != expectedEmptySpecHash {
+		t.Logf("EtcdClusterHash of an empty EtcdClusterSpec changed from %q to %q: "+
+			"the EtcdClusterSpec schema has changed. Note that after an operator "+
+			"upgrade the stored hash annotation of every existing Pod will mismatch "+
+			"and trigger a one-time rolling update of the whole cluster; revisit "+
+			"EtcdClusterHash before updating expectedEmptySpecHash.",
```
ahrtr: I missed this. It seems that your previous solution (define a dedicated struct for hash) is better; it can avoid the unnecessary PODs recreation. 

Please add the struct back, thx

internal/controller/etcdcluster.go
ahrtr: pls see https://github.com/etcd-io/etcd-operator/pull/495#discussion_r3913639701. sorry for the back and forth.

silentred: No worries, let me add the struct back.

internal/controller/etcdcluster_test.go
```diff
@@ -132,3 +220,29 @@ func TestEtcdClusterHash(t *testing.T) {
...
+	if currentHash != expectedEmptySpecHash {
+		t.Logf("EtcdClusterHash of an empty EtcdCluster changed from %q to %q: "+
+			"the hash input or its serialization has changed. Note that after an "+
+			"operator upgrade the stored hash annotation of every existing Pod will "+
+			"mismatch and trigger a one-time rolling update of the whole cluster; "+
+			"revisit EtcdClusterHash before updating expectedEmptySpecHash.",
+			expectedEmptySpecHash, currentHash)
+	}
+	assert.Equal(t, expectedEmptySpecHash, currentHash)
+}
```
ahrtr:
```suggestion
// TestEtcdClusterHashPinned pins EtcdClusterHash's output for an empty
// EtcdClusterSpec, so that any future change to the EtcdClusterSpec schema
// fails this test. If it fails, revisit EtcdClusterHash to decide whether
// the schema change should affect the hash before updating
// expectedEmptySpecHash.
func TestEtcdClusterHashPinned(t *testing.T) {
	// Captured for an all-zero-value EtcdClusterSpec at commit
	// 5483df641bf23347d98a56b8d966bc2c495edcb5.
	const expectedEmptySpecHash = "f71305f40ce2"

	currentHash := EtcdClusterHash(&v1alpha1.EtcdCluster{Spec: v1alpha1.EtcdClusterSpec{}})
	assert.Equal(t, expectedEmptySpecHash, currentHash,
		"hash input changed, see comment above before updating this constant")
}
```

internal/controller/etcdcluster_test.go
silentred: Updated

internal/controller/etcdcluster_test.go
silentred: Updated

ahrtr (approved):

Issue #470, opened 2026-08-21
Support upgrade case


ahrtr:
We only support upgrading one minor version, or patch versions of the same minor version.The related code is here. We upgrade one member at a time.

The related code is [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L442).

ahrtr:
@GunaKKIBM any update on this one? thx

GunaKKIBM:
@ahrtr. It is almost done. Will share the PR in next 12 hours.

GunaKKIBM:
@ahrtr I have raised the PR for upgrade. I am quite sure one of the following changes are necessary. The changes in the PR work perfectly only in case pvc existence for the cluster members. 

Approach 1:
1. Always remove and add the members (and delete and recreate pods).  This is a clean approach. (But, based on standard etcd upgrade docs I have seen, just stopping and starting the member is sufficient. Apologies, I haven't deployed an etcd cluster outside kubernetes that often. Hence, the confusion). 

Approach 2:
1. If the pod isn't backed by storage, remove and add the member. 
2. If the pods are backed by storage, delete and recreate the pod (concern, what if SC is WaitForFirstConsumer mode). 

I did the changes following the etcd docs upgrade procedure. Please let me know, if I should remove and add the member. 

Other questions:
1. There are quite some constraints around certain version upgrades. Like
`Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to 3.5.32 or later](https://etcd.io/blog/2026/july-patch-release/)`
Should we take care of all this in the code? or is it upto to the user? 

2. We need to provide the ability of downloading snapshot, should we add that now? or take it in a different PR. 

Thank you!

GunaKKIBM:
@ahrtr . I did refer the design doc, where in the Recreate phase, only the pod is deleted and recreated, but I think this can fail. So, if recreate fails, according to the design we should proceed to replacing. But, replacing, if I am not wrong, there is a concern. Is it really necessary to delete the pvc and certificates? the pvc may or may not be retained.

silentred:
> Is it really necessary to delete the pvc and certificates? the pvc may or may not be retained

For PVC, we could introduce the `RetentionPolicy` in the future to control the behavior. I think, in current design, removing and creating PVC is a clean approach.
[...]

PR #498, opened 2026-09-03
Update member phase for upgrade functionality

GunaKKIBM:
Addresses: https://github.com/etcd-io/etcd-operator/issues/470
AI assistance is used for the following
1. Add comments on the newly defined functions.
2. Adding Unit test cases

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -578,16 +578,42 @@ func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileSt
 
 // upgradeCluster rolls one member to the etcd version requested in
 // EtcdCluster.Spec.Version, highest ordinal first.
-//
-// TODO: not implemented yet. validateSpec already checks, by comparing
-// the first Pod's image tag against EtcdCluster.Spec.Version, whether the
-// upgrade path is supported when the two differ — but it doesn't persist
-// that comparison anywhere on reconcileState. This phase needs to redo the
-// same current-vs-target version comparison, then bump one EtcdMember.Spec.Version
-// at a time (highest ordinal, non-leader first as a preference); the member
-// notices the drift and recreates via §4.6's ladder (M3).
+
+// upgradeCluster selects the next member to upgrade and marks it for
+// recreation by setting Spec.Version and Phase=Recreating.
+// TODO: All pod-level work
+// (delete old pod, create new pod, health convergence) is handled by
+// reconcileRecreating once dispatch routes the Recreating member there.
 func (r *EtcdClusterReconciler) upgradeCluster(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
-	return ctrl.Result{}, nil
+	logger := log.FromContext(ctx)
+	targetVersion := s.cluster.Spec.Version
+	var healthInfo map[string]etcdutils.EpHealth
+	if s.health != nil {
+		healthInfo = s.health.Members
+	}
+	memberToUpgrade := pickMemberToUpgrade(s.members, healthInfo, targetVersion)
+	if memberToUpgrade == nil {
+		logger.V(1).Info("All members are at desired version", "version", targetVersion)
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("[Upgrade] marking member for recreation",
+		"member", memberToUpgrade.Name,
+		"targetVersion", targetVersion, "currentVersion", memberToUpgrade.Spec.Version,
+	)
+
+	if err := r.updateEtcdMemberStatus(ctx, memberToUpgrade, func(status *ecv1alpha1.EtcdMemberStatus) {
+		status.Phase = ecv1alpha1.EtcdMemberRecreating
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	memberToUpgrade.Spec.Version = targetVersion
+	if err := r.Update(ctx, memberToUpgrade); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{RequeueAfter: requeueDuration}, nil
 }
 
 // updateStatus reflects the current observed state onto EtcdCluster.Status.
--- a/internal/controller/etcdcluster_controller_test.go
+++ b/internal/controller/etcdcluster_controller_test.go
@@ -31,6 +31,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	ecv1alpha1 "go.etcd.io/etcd-operator/api/v1alpha1"
+	"go.etcd.io/etcd-operator/internal/etcdutils"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
 // TestFetchAndValidateState verifies the fetchAndValidateState helper across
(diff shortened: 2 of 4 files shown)


internal/controller/etcdmember.go
```diff
@@ -392,6 +392,35 @@ func pickNotReadyMember(
...
+		if m.Spec.Version == targetVersion {
+			continue
+		}
+		if m.Status.IsLeader {
+			if leaderCandidate == nil {
+				leaderCandidate = m
+			}
+			continue
+		}
+		return m
```
ahrtr: The recreating workflow will take care of the transfering leader.
```suggestion
		if m.Spec.Version != targetVersion {
			return m
		}
```

internal/controller/etcdmember.go
```diff
@@ -392,6 +392,35 @@ func pickNotReadyMember(
...
+	}
+
+	return leaderCandidate
```
ahrtr, line 430:
```suggestion
	return nil
```

internal/controller/etcdcluster_controller.go
```diff
@@ -578,16 +578,37 @@ func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileSt
...
+// TODO: All pod-level work
+// (delete old pod, create new pod, health convergence) is handled by
+// reconcileRecreating once dispatch routes the Recreating member there.
 func (r *EtcdClusterReconciler) upgradeCluster(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
-	return ctrl.Result{}, nil
+	logger := log.FromContext(ctx)
+	targetVersion := s.cluster.Spec.Version
+	memberToUpgrade := pickMemberToUpgrade(s.members, targetVersion)
+	if memberToUpgrade == nil {
+		logger.V(1).Info("All members are at desired version", "version", targetVersion)
```
ahrtr, line 596: It may create lots of noise.

or change it to debug?
```suggestion
```

internal/controller/etcdcluster_controller.go
```diff
@@ -578,16 +578,37 @@ func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileSt
...
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("[Upgrade] marking member for recreation",
+		"member", memberToUpgrade.Name,
+		"targetVersion", targetVersion,
+	)
```
ahrtr, line 603: also print the current version in the log

internal/controller/etcdcluster_controller.go
```diff
@@ -578,16 +578,37 @@ func (r *EtcdClusterReconciler) scaleCluster(ctx context.Context, s *reconcileSt
...
+
+	memberToUpgrade.Spec.Version = targetVersion
+	if err := r.Update(ctx, memberToUpgrade); err != nil {
+		return ctrl.Result{}, err
+	}
+	if err := r.updateEtcdMemberStatus(ctx, memberToUpgrade, func(status *ecv1alpha1.EtcdMemberStatus) {
+		status.Phase = ecv1alpha1.EtcdMemberRecreating
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
```
ahrtr: Update phase first, and version later. 

If etcd-operator crashes in between, then it will unnecessarily recreate the POD with the same version. But if you update the version first, then it may end up never upgrading anymore.


Refer to the rule at https://github.com/etcd-io/etcd-operator/blob/main/docs/design/etcd-member-lifecycle-and-self-healing-v0.3.0.md#46-membership-lifecycle-and-the-failure-recovery-state-machine-per-etcdmember

silentred (commented):
It seems Benjamin has pointed out all the issues. I don't have any other comments. Thanks.

internal/controller/etcdmember.go
GunaKKIBM: I think it is better to check the leader here, and choose it last, reason I am saying, 

If we don't check for leader here, it could be one of the chosen members, assuming, member-3, for which we wouldn't want to upgrade version, given leader should be kept last for upgrade.

Either way, if you suggest we can revisit this later, I will update according to your suggestion. Thanks!

ahrtr: > I think it is better to check the leader here, and choose it last

Agreed. I recall that we discussed in one of previous community meetings

internal/controller/etcdcluster_controller.go
GunaKKIBM: There is no debug call available, only the log level can be set, V(1) here, which won't be called, unless we set in 
opts := zap.Options{
		Development: true,
	}

internal/controller/etcdcluster_controller.go
GunaKKIBM: Done

internal/controller/etcdcluster_controller.go
GunaKKIBM: Done

GunaKKIBM:
@ahrtr . I think I have addressed all the comments. Please let me know if there's anything else that needs to be taken care of. Thank you!

internal/controller/etcdmember.go
```diff
@@ -392,6 +392,35 @@ func pickNotReadyMember(
...
+// Returns nil if all members are already at targetVersion.
+func pickMemberToUpgrade(members []ecv1alpha1.EtcdMember, targetVersion string) *ecv1alpha1.EtcdMember {
+	var leaderCandidate *ecv1alpha1.EtcdMember
+
+	// members is sorted in ascending ordinal order. Iterate in reverse (highest ordinal first).
+	for i := len(members) - 1; i >= 0; i-- {
+		m := &members[i]
+		if m.DeletionTimestamp != nil {
+			continue
+		}
```
```suggestion
```

internal/controller/etcdmember.go
```diff
@@ -392,6 +392,35 @@ func pickNotReadyMember(
...
+		if m.Spec.Version == targetVersion {
+			continue
+		}
+		if m.Status.IsLeader {
```
ahrtr: The m.Status.Isleader is observational only. We don't make decision based on it. Let's use the info (endpoint status) returned by the etcd cluster. We can add one more parameter `healthInfo map[string]EpHealth`.

Inside the `EpHealth`, there is a `Status *clientv3.StatusResponse`, we can use the `Leader` if the status isn't nil.

GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: Done

internal/controller/etcdmember.go
```diff
@@ -392,6 +392,45 @@ func pickNotReadyMember(
...
+		if h.Status != nil && h.Status.Header.MemberId == h.Status.Leader {
+			return name
+		}
+	}
+	return ""
+}
+
+// pickMemberToUpgrade selects the next member to upgrade to targetVersion.
+// It iterates from highest to lowest ordinal and prefers non-leader members first,
+// leaving the leader member for last. If all members needing upgrade are leaders
```
ahrtr: Please remove `If all members needing upgrade are leaders`, there will be only one leader. Even it may return multiple leaders due to temporary leader change, let's also remove this to avoid any confusion.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -820,3 +824,184 @@ func TestFinalizeCluster(t *testing.T) {
...
+		assert.Equal(t, ecv1alpha1.EtcdMemberRecreating, got.Status.Phase)
+
+		// Leader (member-2) must be untouched.
+		gotLeader := &ecv1alpha1.EtcdMember{}
+		require.NoError(t, fakeClient.Get(ctx, client.ObjectKey{Name: m2.Name, Namespace: m2.Namespace}, gotLeader))
+		assert.Equal(t, "v3.5.32", gotLeader.Spec.Version)
+		assert.Equal(t, ecv1alpha1.EtcdMemberReady, gotLeader.Status.Phase)
+	})
+
+	t.Run("only the leader remains — leader is picked", func(t *testing.T) {
```
ahrtr: All the sub test cases in this test are very similar, can you use a table driven sub test cases?

internal/controller/etcdcluster_controller.go
ahrtr: We may need to revisit log message, to ensure we have consistent way to print log. This can be discussed separately

internal/controller/etcdcluster_controller_test.go
GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: Done

ahrtr (approved):
LGTM

Thanks

ahrtr:
Please resolve the workflow failure. 

Please also feel free to add yourself into etcd-io if you want. See example https://github.com/kubernetes/org/pull/5715

GunaKKIBM:
@ahrtr. I am doing updates to my git. Will raise a PR to add myself as member soon. Thank you! 
I have resolved the workflow failure

PR #501, opened 2026-09-07
Update design doc to clarify cleanup of secrets and some minor refactoring

ahrtr:
Main changes:
- clarify that per-member terminating/replacing doesn't delete the secret/certificate, as all members share the same certificates for now
- Clarify the workflow of deleting an etcdCluster, it doesn't change any etcd state, instead it just removes Kubernetes resources. See section 4.13
- Consolidate per-member's replacing and terminate workflow, but  with minor differences.

cc @ArkaSaha30 @GunaKKIBM @ivanvc  @silentred


ahrtr:
also added an open question in `## 7. Open questions` as mentioned in https://github.com/etcd-io/etcd-operator/pull/500#discussion_r3949701755

silentred (approved):
LGTM

PR #505, opened 2026-09-08
Validate storagespec and adjust pvc creation

### LINKED ISSUE #410: Support provisioning PVC/PV for each etcd POD
Previously we depend on StatefulSet controller to take care of `volumeClaimTemplates`, now after migrating to customized controller, we need to manually handle it.

ahrtr:
Fix https://github.com/etcd-io/etcd-operator/issues/410

cc @ArkaSaha30 @GunaKKIBM @ivanvc  @nwnt @silentred

### DIFF
--- a/internal/controller/etcdmember.go
+++ b/internal/controller/etcdmember.go
@@ -305,6 +305,7 @@ func (r *EtcdClusterReconciler) provisionCertificates(ctx context.Context, state
 // ReadWriteMany storage skip it. Idempotent.
 func (r *EtcdClusterReconciler) provisionPVC(ctx context.Context, state *reconcileState, member *ecv1alpha1.EtcdMember) error {
 	cluster := state.cluster
+
 	// ReadWriteMany is assumed to be statically provisioned (e.g. a shared
 	// NFS-backed PVC) and managed outside the operator, so no per-member PVC
 	// is created for it.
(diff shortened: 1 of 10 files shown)


internal/controller/etcdmember.go
```diff
@@ -305,10 +305,8 @@ func (r *EtcdClusterReconciler) provisionCertificates(ctx context.Context, state
 // ReadWriteMany storage skip it. Idempotent.
 func (r *EtcdClusterReconciler) provisionPVC(ctx context.Context, state *reconcileState, member *ecv1alpha1.EtcdMember) error {
 	cluster := state.cluster
-	// ReadWriteMany is assumed to be statically provisioned (e.g. a shared
-	// NFS-backed PVC) and managed outside the operator, so no per-member PVC
-	// is created for it.
-	if cluster.Spec.StorageSpec == nil || cluster.Spec.StorageSpec.AccessModes == corev1.ReadWriteMany {
```
silentred, line 311: https://github.com/etcd-io/etcd-operator/blob/dbc8d7859dc50152050f25c50df772af6ffb0cb5/internal/controller/pods.go#L380-L386

For `ReadWriteMany` PVC, pod volumes uses user-defined `ec.Spec.StorageSpec.PVCName` instead of the PVC operator has just provisioned.

ahrtr: Good catch. Updated.

We also need to add e2e test case to cover `ReadWriteMany`, added a TODO comment on the test case `TestDataPersistence`

silentred (approved):
LGTM

PR #506, opened 2026-09-08
fix: clean up member alarms and resources during deletion

silentred:
Changes as per https://github.com/etcd-io/etcd-operator/issues/471#issuecomment-5573767115:
- disarm alarms when cleaning up EtcdMember resources.
- delete Pod and PVC of each member when terminating a EtcdCluster.
- Remove `TestNormalMemberProvisioning` and add cases to `TestScaling`

/cc @ahrtr

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -526,23 +526,22 @@ func (r *EtcdClusterReconciler) ensureClusterFinalizer(ctx context.Context, s *r
 	return r.Update(ctx, s.cluster)
 }
 
-// finalizeCluster handles an EtcdCluster with DeletionTimestamp set: it
-// removes every EtcdMember it still owns — reusing the same interim
-// finalizer-clearing dispatch()'s Terminating-cleanup step uses for a
-// single Terminating member — and, once none remain, releases
-// clusterCleanupFinalizer so Kubernetes can finish deleting the EtcdCluster
-// itself.
-//
-// TODO: like that step, this skips the real six-step leave sequence (§4.6,
-// M3): members are removed without ever calling etcd's MemberRemove, so
-// deleting a cluster with live members can leave stale entries in etcd's
-// own membership list. Acceptable for now because the Pods backing those
-// members are torn down along with everything else, but M3 should replace
-// this loop with the real sequence rather than just deleting faster.
+// finalizeCluster performs Kubernetes-only cleanup for whole-cluster deletion.
+// Each member's Pod and PVC delete requests must succeed before its finalizer
+// is released. Once no members remain, the cluster's shared TLS certificate
+// Secrets are deleted and then the cluster's own finalizer is released,
+// letting the EtcdCluster object disappear. No live etcd calls are needed.
 func (r *EtcdClusterReconciler) finalizeCluster(ctx context.Context, s *reconcileState) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
 	if len(s.members) == 0 {
+		if err := deleteClusterCertificateSecrets(ctx, s.cluster, r.Client, []string{
+			getClientCertName(s.cluster.Name),
+			getServerCertName(s.cluster.Name),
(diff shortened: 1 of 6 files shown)


internal/controller/etcdcluster_controller.go
ahrtr, line 545: Before we remove the cluster finalizer, we need to clean up the certificates/secrets if not yet.

test/e2e/e2e_test.go
```diff
@@ -128,85 +128,6 @@ func TestClusterHealthy(t *testing.T) {
...
-		{name: "ScaleOutFrom1To3", initialSize: 1, desiredSize: 3},
-		{name: "ScaleOutFrom3To5", initialSize: 3, desiredSize: 5},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			feature := features.New(tc.name)
-			clusterName := "normal-" + strings.ToLower(tc.name)
-
-			feature.Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
```
ahrtr, line 147: nice to move it into TestScale.

internal/controller/etcdcluster_controller_test.go
```diff
@@ -1095,168 +1098,203 @@ func createClusterHealthWithLeader(ec *ecv1alpha1.EtcdCluster, size, leaderOrdin
...
+			require.NoError(t, fakeClient.Get(ctx, client.ObjectKeyFromObject(member0), member0))
+			assert.Contains(t, member0.Finalizers, memberCleanupFinalizer)
+			require.NoError(t, fakeClient.Get(ctx, client.ObjectKeyFromObject(member1), member1))
+			assert.Nil(t, member1.DeletionTimestamp, "the next member must not start after an error")
+			require.NoError(t, fakeClient.Get(ctx, client.ObjectKeyFromObject(pvc), pvc))
+			if !tc.failPVC {
+				require.NoError(t, fakeClient.Get(ctx, client.ObjectKeyFromObject(pod), pod))
+			} else {
+				assert.True(t, apierrors.IsNotFound(fakeClient.Get(ctx, client.ObjectKeyFromObject(pod), pod)))
+			}
```
ahrtr: The implementation of this test seems a little complicated and worse readability. 

Please consider simplifying it if possible.

ahrtr:
Please also rebase this PR. I just merged https://github.com/etcd-io/etcd-operator/pull/505

silentred:
> Please also rebase this PR. I just merged #505

All resolved, PTAL.

internal/controller/etcdmember.go
```diff
@@ -105,6 +107,10 @@ func (r *EtcdClusterReconciler) cleanupEtcdMember(ctx context.Context, s *reconc
 		return ctrl.Result{}, err
 	}
 
+	if err := disarmForEtcdMember(s, member); err != nil {
```
ahrtr, line 110: There is one potential bug here. If previous the controller crashes or interrupted for some reason after previous step `removeEtcdNode`, but before this step, then it will never be able to disalarmed,  because you won't be able to fine the etcdNode anymore.

Two solutions:
- The alarms are persisted on every etcd node, so we don't have to find the exact etcd node to read the alarm; instead we can read alarm from any etcd node, to ensure we get the latest alarm, reading it from the leader is better. It also needs to ensure we know the etcd node's ID even it's been removed, so we need to ensure the EtcdMember.Status.MemberID (and other fields) is updated. 
- keep it as it's; and we depends on the separate CORRUPTION and NOSPACE workflows to handle them.

No matter which way, we need to ensure the EtcdMember.Status is updated in one centralized place in a generic way, probably in a deferred function in `reconcileEtcdMember` . Let's get this resovled first in a followup first, and then evaluate how to resolve the above issue.

ahrtr (approved):
LGTM with one comment, let's resolve it in a followup PR.

thx

internal/controller/etcdmember.go
silentred: Good catch!

PR #508, opened 2026-09-11
fix: disarm removed member alarms through the leader

silentred:
Link https://github.com/etcd-io/etcd-operator/issues/471#issuecomment-5586719637
/cc @ahrtr

### DIFF
--- a/internal/controller/etcdmember.go
+++ b/internal/controller/etcdmember.go
@@ -101,8 +101,8 @@ func (r *EtcdClusterReconciler) markMemberTerminating(ctx context.Context, membe
 // and finally release memberCleanupFinalizer so Kubernetes can finish the deletion.
 // Membership and resource cleanup are no-ops once done, so re-entering after
 // an interruption (operator restart, transient etcd error) resumes harmlessly.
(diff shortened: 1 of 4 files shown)


internal/controller/etcdmember.go
```diff
@@ -166,19 +166,40 @@ func removeEtcdNode(s *reconcileState, member *ecv1alpha1.EtcdMember) error {
...
+
+	for i := range s.members {
+		if s.members[i].Name == leaderName {
+			leader = &s.members[i]
+			break
+		}
+	}
+	if leader == nil {
+		return fmt.Errorf("failed to find EtcdMember for leader %q not found", leaderName)
+	}
```
ahrtr, line 192: I just realized that alarm is a linearizable request, it doesn't matter we get alarm from which etcd server.

It seems that second solution mentioned in https://github.com/etcd-io/etcd-operator/pull/506/changes#r3958924801 is better & cleaner. 

Let's take care of this in phase 4 when handling the alarms.

thx for the effort anyway.

silentred: OK, let me close this PR.

PR #512, opened 2026-09-16
feat: support the EtcdMember Replacing workflow

### LINKED ISSUE #473: Support the Replacing case
If recreating doesn’t help, then we need to replace the EtcdMember. It first remove the etcd from the cluster and cleanup all resources owned by the EtcdMember, then follow the same logic to provisioning the EtcdMember.

Refer to the Replacing case in the reconcile_member diagram

silentred:
Fixes #473

- Reset RecreateCount when marking a member Replacing. 
- Reuse member cleanup, retain the finalizer, and wait for Pod/PVC deletion before transitioning directly to Provisioning.
- Add e2e test for Replacing workflow

/cc @ahrtr @nwnt @GunaKKIBM 
FYI: `r.markMemberReplacing()` is reusable for `Recreating` workflow.


internal/controller/etcdmember.go
```diff
@@ -137,8 +182,10 @@ func (r *EtcdClusterReconciler) cleanupEtcdMember(ctx context.Context, s *reconc
 		return ctrl.Result{}, err
 	}
 
-	if err := r.clearMemberFinalizer(ctx, member); err != nil {
-		return ctrl.Result{}, err
+	if member.Status.Phase != ecv1alpha1.EtcdMemberReplacing {
```
```suggestion
	if member.Status.Phase == ecv1alpha1.EtcdMemberTerminating {
```

internal/controller/etcdmember.go
```diff
@@ -96,9 +140,10 @@ func (r *EtcdClusterReconciler) markMemberTerminating(ctx context.Context, membe
 	})
 }
 
-// cleanupEtcdMember is the §4.6 Terminating leave for one member: remove it
-// from etcd's live membership, disarm its alarms, delete its owned Pod and PVC,
-// and finally release memberCleanupFinalizer so Kubernetes can finish the deletion.
+// cleanupEtcdMember is the §4.6 leave for one member: remove it from etcd's
+// live membership, disarm its alarms, and delete its owned Pod and PVC.
+// Replacing retains the finalizer; otherwise, release memberCleanupFinalizer
```
```suggestion
// Only the Terminating case removes the memberCleanupFinalizer
```

ahrtr (approved):

Issue #472, opened 2026-08-21
Support the Recreate case


ahrtr:
If an etcd POD isn’t somehow healthy and not belong to the corruption and NOSPACE cases, we should try to recreate it up to 3 times. If recreating doesn’t work, then try to replace it. The related code is [here](https://github.com/etcd-io/etcd-operator/blob/c8a968fc5575ff84bba294f0ea9d3ec5d1ffd409/internal/controller/etcdcluster_controller.go#L404-L416).

Refer to the Recreating case in the reconcile_member diagram.

ahrtr:
@GunaKKIBM This task is relatively important, and many other workflows depend on this one. So I expect we can resolve this early next week. Please let me know if you have any difficulties.

ahrtr:
Resolved in https://github.com/etcd-io/etcd-operator/pull/515 and https://github.com/etcd-io/etcd-operator/pull/519

PR #515, opened 2026-09-18
Recreate Member workflow

GunaKKIBM:
Addresses https://github.com/etcd-io/etcd-operator/issues/472


GunaKKIBM:
@ahrtr @silentred @ArkaSaha30  I request you to review this PR. 

There's a small change as compared to the design the these changes. 

I think, we should check for the pod time out (on the target version pod) before checking recreate count, with this, the 3rd iteration of recreating will also wait for the timeout duration. Please let me know your thoughts.

I haven't added any tests yet, will add once after addressing the first round of comments related to any functional changes. 

Thanks!

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+				status.RecreateCount = 0
+			}); err != nil {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{RequeueAfter: requeueDuration}, nil
+		}
+		// Pod is healthy but still on old image — fall through to replace it.
+		logger.Info("[Recreating] member is healthy but on old image, will replace pod",
+			"member", member.Name,
+		)
```
ahrtr, line 433: When the member is healthy, we also need to check whether there is config change. If yes, we will also need to continue the recreation workflow.

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+	}
+
+	// 3. Member is unhealthy on target version, check if the pod has timed out
+	// If the pod hasn't timed out, requeue
+	if memberPod != nil && onTargetVersion {
```
ahrtr, line 438: why do we need to check `onTargetVersion` here?

ahrtr:
> I think, we should check for the pod time out (on the target version pod) before checking recreate count, with this, the 3rd iteration of recreating will also wait for the timeout duration. Please let me know your thoughts.

Agreed. thx

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
 	return ctrl.Result{RequeueAfter: requeueDuration}, nil
 }
 
+// podStartTimeout is the maximum time a Pod is allowed to be running but not
+// Ready before the reconciler treats it as timed out and replaces it.
+const podStartTimeout = 2 * time.Minute
```
ahrtr, line 385: 2 minutes seems too long, let's use 30s instead for now. Also add a TODO comment to clarify that we may add a configuration item in future.

internal/controller/etcdmember.go
silentred: I think the conditions to set `Phase=Ready` should be:
1. health.Health is true
2. AND `pod.Annotations[HashMetadataKey]` equals `EtcdClusterHash(ec)`
3. AND `tagOf(pod.Spec.Containers[0].Image)` equals `EtcdMember.Spec.Version`

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+		logger.Info("[Recreating] member is healthy but on old image, will replace pod",
+			"member", member.Name,
+		)
+	}
+
+	// 3. Member is unhealthy on target version, check if the pod has timed out
+	// If the pod hasn't timed out, requeue
+	if memberPod != nil && onTargetVersion {
+		timedOut := memberPod.Status.StartTime != nil &&
+			time.Since(memberPod.Status.StartTime.Time) > podStartTimeout
```
silentred, line 440: It seems `Status.StartTime` is not set anywhere. Alternatively, we can use `pod.CreationTimestamp` as specified in the design doc。

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+			return ctrl.Result{RequeueAfter: requeueDuration}, nil
+		}
+		// Pod has timed out — fall through to delete + recreate below.
+		logger.Info("[Recreating] pod timed out", "member", member.Name, "pod", memberPod.Name)
+	}
+
+	// 4. Escalate to Replacing if RecreateCount is exhausted.
+	if member.Status.RecreateCount >= 3 {
+		logger.Info("[Recreating] RecreateCount exhausted, escalating to Replacing", "member", member.Name, "recreateCount", member.Status.RecreateCount)
+		if err := r.updateEtcdMemberStatus(ctx, member, func(status *ecv1alpha1.EtcdMemberStatus) {
```
silentred, line 453: We could use `r.markMemberReplacing()` here.

internal/controller/etcdmember.go
```diff
@@ -378,6 +380,136 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+// the Recreating case in reconcile_member_v0.3.0.png
+func (r *EtcdClusterReconciler) reconcileRecreating(
+	ctx context.Context,
+	state *reconcileState,
+	member *ecv1alpha1.EtcdMember,
+) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// 1. Member health check.
+	health, err := findHealthStatusForEtcdMember(state, member)
```
silentred, line 398: There is a dead loop here. After the pod was deleted in the previous loop, `findHealthStatusForEtcdMember` will never find pod's health info. 
Because in `healthCheck`, `etcdutils.MemberHealth` accpets `podNames`, in witch the just deleted pod is excluded. 

https://github.com/etcd-io/etcd-operator/blob/ca1c0717dd6f5c79a900dd4d54522f765442ccff/internal/controller/pods.go#L214-L229

I have not figured out a solution for this. @ahrtr

ahrtr: I think it's deterministic to know whether the POD exists or not. We can check it first.

ahrtr: Also if the POD doesn't exist, shouldn't the health check against the member/POD getting a unhealthy result?

silentred: I've drafted some pseudocode. Could you please take a look when you have a moment?

```
func reconcileRecreating(...) (ctrl.Result, error) {
	pod := findPodForEtcdMember(state, member)

	// 1. if pod is in the middle of deletion, requeue the request
	if memberPod != nil && memberPod.DeletionTimestamp != nil {
        return ctrl.Result{RequeueAfter: requeueDuration}, nil
	}

  	// 2. if pod doest not exists
	if pod == nil {
  		// mark Phase=Replacing if RecreateCount >= 3;
		if member.Status.RecreateCount >= maxRecreates {
			persist member.Status {
				Phase = Replacing
				RecreateCount = 0
            }
			return requeue
        }

  		// Create a new pod.
  		createPodForEtcdMember() // increases RecreateCount

		return ctrl.Result{RequeueAfter: requeueDuration}, nil
	}

  	// 3. health check
	health, err := findHealthStatusForEtcdMember(state, member)
	if err != nil {
  		// state snapshotting may have failed, but it does not mean the Node is unhealty.
  		// so we requeue for snapshotting again in the next loop.
		return ctrl.Result{RequeueAfter: requeueDuration}, nil
	}


  	// 4. Node is healthy and conditions(configs, version) are all met, set Phase=Ready
	configMatches := pod.Annotations[HashMetadataKey] ==
					EtcdClusterHash(state.cluster)
	imageMatches := etcdContainer(pod).Image ==
					desiredImage(state.cluster, member)
	if health.Health && configMatches && imageMatches {
		persist member.Status {
			Phase = Ready
			RecreateCount = 0
		}
		return requeue
	}

	// 5. if Node is not healthy
	if !health.Health {
		// if not timed out, requeue to wait
		if !podReadyTimeout(pod.CreationTimestamp) {
			return requeue
		}

		// if RecreateCount >=3, set Phase=Replacing
		if member.Status.RecreateCount >= maxRecreates {
			persist member.Status {
				Phase = Replacing
				RecreateCount = 0
			}
			return requeue
		}
	}

	// 6. Node is healthy.
	// if the current member is leader, move leader to the Node with lowest ordinal and ready Phase
	if isEtcdMemberLeader(state, member) {
		tgt := selectTransferTarget()
		trasferLeader(tgt)
	}

	// 7. delete the drifted pod
	delete(pod)
	return requeue
}
```

ahrtr: Overall looks good. Just one minor comment, in step 6, if `trasferLeader` fails, requeue.

GunaKKIBM: Looks good to me. Thank you! I will add the changes.

GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: Done

internal/controller/etcdmember.go
GunaKKIBM: My original thought was, if the member is unhealthy and not on target version, it can be recreated/replaced, but with new code flow I have pushed, this doesn't exist.

internal/controller/etcdmember.go
GunaKKIBM: Done

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,154 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
+
+	// 4. Node is healthy and conditions(configs, version) are all met, set Phase=Ready
+	configMatches := memberPod.Annotations[HashMetadataKey] ==
+		EtcdClusterHash(state.cluster)
+	onTargetVersion := false
+
+	expectedImage := fmt.Sprintf("%s:%s", state.cluster.Spec.ImageRegistry, member.Spec.Version)
+	for _, c := range memberPod.Spec.Containers {
```
silentred: Can we just regard `memberPod.Spec.Containers[0]` as the etcd container? There should be only one container now.

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,154 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+		}
+	}
+
+	// 5. Transfer leadership away before taking this member offline, so the
+	// cluster doesn't lose its leader unnecessarily. This is a best-effort
+	// operation — if it fails, requeue
+	if member.Status.IsLeader {
+		endpoints := clientEndpointsFromPods(state.cluster.Name, state.cluster.Namespace, state.pods, clusterTLSEnabled(state.cluster))
+		if len(endpoints) > 0 {
+			// Pick the lowest-ordinal member that is not this one as the transfer target.
```
silentred: Target member should be healthy as well. We can resolve this in a follow-up PR.

silentred (commented):
Just 2 minor comments.

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,154 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+						"error", err,
+					)
+					return ctrl.Result{RequeueAfter: requeueDuration}, err
+				}
+			}
+		}
+	}
+
+	// 6. Delete the timed-out Pod or pod with old version of image, if it still exists.
+	logger.Info("[Recreating] deleting timed-out pod", "member", member.Name, "pod", memberPod.Name)
```
```suggestion
	logger.Info("[Recreating] deleting unhealthy pod", "member", member.Name, "pod", memberPod.Name)
```

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,154 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+				if err := etcdutils.MoveLeader(etcdutils.ClientConfig{Endpoints: endpoints, TLS: state.tlsConfig}, transferTargetID); err != nil {
+					logger.Info("[Recreating] leader transfer failed, requeueing",
+						"member", member.Name,
+						"error", err,
+					)
+					return ctrl.Result{RequeueAfter: requeueDuration}, err
+				}
+			}
+		}
+	}
```
ahrtr, line 494: we can get all the details of moving leader in a function.

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,154 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+		return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	}
+
+	// 5. If Node is not healthy
+	if !health.Health {
+		timedOut := time.Since(memberPod.CreationTimestamp.Time) > podStartTimeout
+
+		if !timedOut {
+			// Pod is still starting up — give it more time.
+			logger.Info("[Recreating] pod not yet timed out, requeueing", "member", member.Name, "pod", memberPod.Name)
```
ahrtr: add `elapsed` and `timeout`
`[Recreating] Pod readiness not timed out yet, requeueing. elapsed: 45s, timeout: 60s`

internal/controller/etcdmember.go
GunaKKIBM: it could be unhealthy or a healthy pod in older version

logger.Info("[Recreating] deleting pod (unhealthy or outdated version)",
		"member", member.Name,
		"pod", memberPod.Name,
		"currentImage", memberPod.Spec.Containers[0].Image,
		"targetImage", expectedImage,
	)
	
Is this better?

ahrtr:
```
[Recreating] deleting pod, member: etcd-1, reason: xxx
```

The reason could be one of
- unhealthy
- upgrade (3.5.6 -> 3.5.7)
- config update

ahrtr:
Please mark the comment as resolved after you resolve them. 

https://github.com/etcd-io/etcd-operator/pull/515#discussion_r4056141177

internal/controller/etcdmember.go
GunaKKIBM: May I know if you are expecting conditional logging? 
Like checking config, health and version and logging accordingly or just logging the entire thing in log line, no. matter whether they are different or same

ahrtr: > May I know if you are expecting conditional logging?

yes.

internal/controller/etcdmember.go
```diff
@@ -368,16 +380,186 @@ func (r *EtcdClusterReconciler) reconcileProvisioning(
...
+	// TODO: The target member should be healthy before making it the leader
+	if member.Status.IsLeader {
+		endpoints := clientEndpointsFromPods(state.cluster.Name, state.cluster.Namespace, state.pods, clusterTLSEnabled(state.cluster))
+		if len(endpoints) > 0 {
+			tgt := selectTransferTarget(state, member)
+			if err := r.transferLeader(ctx, state, member, endpoints, tgt); err != nil {
+				return ctrl.Result{RequeueAfter: requeueDuration}, err
+			}
+		}
+	}
```
ahrtr, line 494: several comments:
1. We should use the health check result to select a healthy member
2. My previous comment https://github.com/etcd-io/etcd-operator/pull/515#discussion_r4056789804 might not be optimal. The leader transfer should be just best effort, we don't have to requeue on error. Sorry for the confusion.

Let me do a quick followup once this one get merged.

ahrtr (approved):
LGTM & thx

internal/controller/etcdmember.go
ahrtr: also we shouldn't use member.Status.IsLeader to check whether it's a leader, use the endpoint status instead.

silentred (approved):
LGTM, thanks for the contribution.

PR #516, opened 2026-09-18
Update recreate workflow: check pod timeout before recreateCount

ahrtr:
Per https://github.com/etcd-io/etcd-operator/pull/515#issuecomment-5728303363

Check pod timeout before checking RecreateCount

cc @GunaKKIBM @silentred @ArkaSaha30 @nwnt


ahrtr:
@ivanvc

nwnt:
Nit: can we instead change it to 'Pod Readiness Wait Timeout' instead of just time out.  I think putting just time out is a bit ambiguous.

ahrtr:
> Nit: can we instead change it to 'Pod Readiness Wait Timeout' instead of just time out. I think putting just time out is a bit ambiguous.

Makes sense. Updated

nwnt (approved):

